PR #14, opened 2023-05-19
Include columnOffset for jest Script creation

BobdenOs:
When Jest compiles a file it does not forward the size of the module wrapper to node. This causes issue in the node debugger source map logic. As it cannot resolve the source file as the filesystem contains different code then the compiled Script. By adding the offset the contents match up again and the editor will not open an read only window anymore.

Using the normal Jest babel configuration breaks a lot of debugging information all values on hover and inside the debug console are not the actual value they should be. Which means that it is not possible to copy a line from the source file and run it directly in the debug console. As the moment a property is selected the root value has become the value of the property.

Also tried to make a PR to Jest directly, but not certain about the Corporate relation between SAP and Facebook. https://github.com/jestjs/jest/pull/14148

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,7 +16,10 @@ const enable = process.argv
   })
   .filter(a => a)
 
+require('./test/jest')
+
 module.exports = {
+  transform: {},
   reporters: process.env.CI ? ['github-actions', 'summary'] : ['default'],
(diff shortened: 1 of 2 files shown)


BobdenOs:
fixed in jest instead

PR #42, opened 2023-06-02
Use CAP Plugin Mechanism for Auto Configuration

sebastianesch:
Attempt to use the CAP Plugin mechanism to set the required VCAP labels / tags for service resolution in Cloud Foundry for minimal manual configuration.

Currently the approach does not work, as CAP resolves services from env before the plugin is triggered.


patricebender:
I think this is already works, can we close?

danjoa:
is that still relevant?

PR #89, opened 2023-07-03
cqn4sql: Replace `$self` in aggregations and token streams / propagate join relevance

### LINKED ISSUE #88: dump in cqn4sql for aggregation on `$self.<column>`
```js
    it.only('WHERE with implicit table alias', () => {
      let query = cqn4sql(
        CQL`SELECT from bookshop.Authors {
            name as author,
            years_between(dateOfBirth, dateOfDeath) as age
          }
          group by $self.author
         `, model
        )
      expect(query).to.deep.equal(
        CQL`SELECT from bookshop.Authors as Authors {
          Authors.name as author,
          years_between(Authors.dateOfBirth, Authors.dateOfDeath) as age
        }
        group by $self.name
       `,
      )
    })
```

patricebender:
plug in new function introduced with #70, add some tests and apply basic refactoring on the fly.

The function which handles `$self` paths, needed to be moved up in the scope to be available not only for the columns transformation but also for `order by / group by` and token stream transformation (such as `having`, `where`)

---

also fix the case where the `$self` path addresses join relevant columns, the join relevance flag must be propagated in order for the table alias replacement to work properly.

fixes #88


johannes-vogel:
changelog?

johannes-vogel (approved):

Issue #93, opened 2023-07-05
`UPDATE` with join relevant conditions


patricebender:
```
UPDATE.entity(Books).data( { texts: [ { locale: 'es', descr: 'Òla'} ] } ).where(`genre.name = 201`).forSQL()
```
yields this, invalid, CQN:
```json5
Query {
  UPDATE: {
    entity: { ref: [ 'sap.capire.bookshop.Books' ], as: 'Books' },
    data: { texts: [ { locale: 'es', descr: 'Òla' } ] },
    where: [ { ref: [ 'genre', 'name' ] }, '=', { val: 201 } ],
    from: {
      join: 'left',
      args: [
        { ref: [ 'sap.capire.bookshop.Books' ], as: 'Books' },
        { ref: [ 'sap.capire.bookshop.Genres' ], as: 'genre' }
      ],
      on: [
        { ref: [ 'genre', 'ID' ] },
        '=',
        { ref: [ 'Books', 'genre_ID' ] }
      ]
    }
  }
}
```

this could probably be translated to an `UPDATE` with a `where exists` subquery, correlating both tables.

patricebender:
decision from the db steering meeting on 10th july:
- turn such queries into `UPDATE` with `where exists`

patricebender:
addendum: We do not understand the token stream such as a `where` clause, it is not easily possible to turn `UPDATE.entity('bookshop.Books').where(`author.name LIKE '%Bron%' or ( author.name LIKE '%King' and title = 'The Dark Tower') and stock >= 15`)` into a corresponding `where exists` subquery. For now, I would forbid path expressions in `UPDATE` statements, until we found a proper solution. @stewsk  FYI

patricebender:
resolved in #97

PR #97, opened 2023-07-07
infer: reject path expressions for `UPDATE` statements

patricebender:
as described in #93, a path expression which induces a join in an `UPDATE` statement, errorneously leaded to the transformed `UPDATE` statement to include both an `entity` and a `from` property. 

Unfortunately, we do not really understand a token stream such as a where clause,
hence we cannot easily rewrite a path expression into a `where exists` subquery.

For the moment, we should issue a proper error instead of dumping.

see https://github.com/cap-js/cds-dbs/issues/93#issuecomment-1632460212


stewsk (approved):

patricebender:
for reference: https://stackoverflow.com/questions/1293330/how-can-i-do-an-update-statement-with-join-in-sql-server

PR #99, opened 2023-07-08
Point links at npmjs to GitHub repository

gregorwolf:
The links at https://www.npmjs.com/package/@cap-js/postgres are currently very generic. I think we should point them to the now public GitHub repository.


patricebender (commented):
LGTM, would you mind adding this to the sqlite package.json as well as part of your change?

BobdenOs (approved):

Issue #100, opened 2023-07-09
Multiple execution of cds deploy --profile pg fails with "duplicate key value violates unique constraint"


gregorwolf:
When following [Using Defaults with [pg] Profile](https://cap.cloud.sap/docs/guides/databases-postgres#using-defaults-with-pg-profile) and running:

```
cds deploy --profile pg
```

with provided sample data it works fine at the first execution. But running it a second time fails with:

```
error: in cds.deploy(): duplicate key value violates unique constraint
```

In `cds help deploy` I don't find any option that would allow an upsert.

PR #102, opened 2023-07-11
Postgres: schema_evolution = 'auto' by default

simonoswald:
Currently, we do this in a hacky way in `cds deploy` - we should instead do it more centrally in the plugin.

Should fix https://github.com/cap-js/cds-dbs/issues/100

### DIFF
--- a/postgres/package.json
+++ b/postgres/package.json
@@ -46,7 +46,8 @@
               "user": "postgres",
               "password": "postgres",
               "database": "postgres"
-            }
+            },
+            "schema_evolution": "auto"
           }
         },
         "postgres": {
--- a/postgres/package.json
+++ b/postgres/package.json
@@ -35,7 +35,6 @@
       "kinds": {
         "sql": {
           "[production]": {
-            "schema_evolution": "auto",
             "kind": "postgres"
           },
           "[pg!]": {
@@ -46,8 +45,7 @@
               "user": "postgres",
               "password": "postgres",
               "database": "postgres"
-            },
-            "schema_evolution": "auto"
+            }
           }
         },
         "postgres": {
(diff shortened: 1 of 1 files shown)


simonoswald:
I am not 100% sure, but the way I understand it, the plugin will provide the things configured in the package.json as default cds options, right?

postgres/package.json
```diff
@@ -45,15 +46,17 @@
...
           }
         },
         "postgres": {
           "impl": "@cap-js/postgres",
           "dialect": "postgres",
           "vcap": {
             "label": "postgresql-db"
-          }
+          },
+          "schema_evolution": "auto"
```
johannes-vogel, line 59: I think it's only required to be set here. The profiles refer to this config via `kind` property.

gregorwolf:
Should be add a test to check this automatically?

simonoswald:
> Should be add a test to check this automatically?

Maybe a check that prints the `cds env` and checks that schema_evolution is set to "auto"? 🤔 Do your tests have access to `cds env`?

johannes-vogel:
> > Should be add a test to check this automatically?
> 
> Maybe a check that prints the `cds env` and checks that schema_evolution is set to "auto"? 🤔 Do your tests have access to `cds env`?

We could also add a similar test that uses schema migration as we do in cds package?

simonoswald:
> > > Should be add a test to check this automatically?
> > 
> > 
> > Maybe a check that prints the `cds env` and checks that schema_evolution is set to "auto"? 🤔 Do your tests have access to `cds env`?
> 
> We could also add a similar test that uses schema migration as we do in cds package?

I think that rebuilding the cds-deploy tests here is a bit overkill

PR #103, opened 2023-07-11
First draft: test for calc-elem

stewsk:


stewsk:
Hi @patricebender , here some very first and very rough tests on calc elems. Not thoroghly checked, more a first hint how it could look like ,,,

db-service/test/cqn4sql/model/booksWithExpr.cds
```diff
@@ -0,0 +1,71 @@
...
+  area : Decimal = length * width;
+  volume : Decimal = area * height;
+  storageVolume : Decimal = stock * volume;
+
+  // -- with paths
+  authorLastName = author.lastName;
+  authorName = author.name;
+  authorAdrText = author.addressText;
+
+  ca3 = f.ca; // F:ca is a calculated element with an assoc path
```
patricebender, line 38: i can't find `f` anywhere, is this a typo or how does this work?

stewsk: I started using "simple" names like f, c, ... and then changed to using "nice" names like author, name, ...
In the state I put into the PR there still is a mixture of both. Need to clean that up.

patricebender (approved):
Lets merge this one and then grow as we go!

PR #104, opened 2023-07-12
reuse compliance

johannes-vogel:
I had a sync with a colleague from OSPO team to identify the missing things in our repo. These are the missing things


patricebender (approved):
Could you add just a little commit message from where this is coming from and why we need it?

johannes-vogel:
> Could you add just a little commit message from where this is coming from and why we need it?

I had a sync with a colleague from OSPO team to identify the missing things in our repo. These are the missing things :)

PR #113, opened 2023-07-18
Enable calculated elements in db-service

patricebender:
With this change, the first calculated elements are enabled in runtime queries.

#### changes in `infer`
- detect calculated elements while walking through the `columns` or a token stream
- recursively drill down into the model and attach `$refLinks` if applicable
- a calculated element with a `ref` or an `xpr` in it's `.value`, might end up in several paths. Calculate all of those paths and check each of them for _join relevance_, each _join relevant_ path is then inserted into the join tree:
example: the column `authorAdrText` -> `authorAdrText = author.addressText` -> `addressText = address.text` -> `text = street || ', ' || city` will end up in two paths:
`[author, address, street]` and `[author, address, city]`
--> TODO: Optimize algorithm to not merge the same calculated element multiple times.

#### changes in `cqn4sql`
not much to do here.. main logic is in `infer`
- recursively replace calculated elements in place in `columns` and token streams


db-service/test/cqn4sql/calculated-elements.test.js
```diff
@@ -98,17 +169,116 @@ describe.skip('Unfolding calculated elements in select list', () => {
...
+
+// ? calc elem at several places in one query (select, where, order ...) ?
+
+describe('Unfolding calculated elements ... misc', () => {
+  let model
+  beforeAll(async () => {
+    model = cds.model = await cds.load(__dirname + '/model/booksWithExpr').then(cds.linked)
+  })
+  // Calculated elements on-write are not supported, yet (in entity:“booksCalc.Books”/element:“areaS”)
+  it.skip('calculated element on-write (stored) is not unfolded', () => {
```
stewsk, line 245: please check, compiler supports on-write with 4.0

patricebender: We will wait for #119

db-service/test/cqn4sql/calculated-elements.test.js
```diff
@@ -98,17 +169,116 @@ describe.skip('Unfolding calculated elements in select list', () => {
...
+    const expected = CQL`SELECT from localized.booksCalc.LBooks as LBooks {
+        LBooks.ID,
+        LBooks.title,
+        LBooks.length * LBooks.width as area
+      }`
+    expected.SELECT.localized = true
+    expect(query).to.deep.equal(expected)
+  })
+
+  it.skip('calculated element refers to localized element', () => {
```
patricebender, line 272: @stewsk FYI: had to skip this and comment out the calc element which points to a localized element. Will re-enable once compiler 4.1 is on npm

PR #114, opened 2023-07-18
[PoC] generate .d.ts files from js doc

BobdenOs:
## Concept

With the pluggable DB layer being a public API. It is important for it to be documented clearly. Additionally there are requests for typescript support. So to address both this PR is providing an implementation that generates `.d.ts` files from the `JSDoc` comments. Allowing the `db-service` package to be javascript, but still provide type definitions that reflect reality.

## Issues

Typescript does not follow its own rules when generating `.d.ts` files from `JSDoc`s.

### Corrupted `.d.ts` files

In javascript it is allowed to call a property `class`, but in typescript `class` is a reserved word. This does how ever not stop Typescript from generating a `.d.ts` file that uses the `class` reserved word.

```js
module.exports = () => render()
module.exports.class = CQN2SQLRenderer
```

For `CQN2SQL` to be exposed as a class on the root of the module it is required to reference `cqn2sql.class`. This creates the following `.d.ts` files

```typescript
CQN2SQL_1 = import('cqn2sql')
CQN2SQL = CQN2SQL_1.class // <-- class is a reserved word 🤷
exports = {
  CQN2SQL
}
```

This is done transparently without any compile errors

### no (implicit) any

For Typescript to have a chance to make sense. It is important that all types are defined on every layer. This means that if there is a single library which does not define its types. Everyone consuming it has to define the correct types or the compiler can't provide any guarantees.

Therefor this PR include `strict:true` in the `tsconfig`, but when compiling from `JSDoc` this is ignored and `any` is written explicitly when no type is defined or when a type is defined that could not be resolved. Therefor it would be required to enhance the logic to validate the generated `.d.ts` files and ensure that they don't include a single explicit `any`.

https://typescript-eslint.io/rules/no-explicit-any/

"The `any` type in TypeScript is a dangerous "escape hatch" from the type system. Using `any` disables many type checking rules and is generally best used only as a last resort or when prototyping code."

## References

[discussions#2](https://github.com/orgs/cap-js/discussions/2)


patricebender (approved):
LGTM. 
-> I will follow up on the `unknown` stuff in `infer/`

PR #115, opened 2023-07-18
[Fix] Enhance error message when '*' is not expanded

BobdenOs:
When a query does not reference an entity in the model `'*'` is not expanded and can cause issues in `cqn2sql`. Therefor this PR updates the error thrown when this case is hit. Clarifying that the target could not be inferred or if that is expected that `'*'` can not be used in the query.


patricebender:
Could you add a test which documents this?

patricebender (approved):

PR #119, opened 2023-07-19
actions: Use compiler v4^ in `package-lock.json`

patricebender:
`"@sap/cds": ">=7"` -> leads to `cds@7.0.0` which still depends on `compiler@3.9.2`

### DIFF
--- a/postgres/test/__snapshots__/csn2postgres.test.js.snap
+++ b/postgres/test/__snapshots__/csn2postgres.test.js.snap
@@ -3,7 +3,7 @@
 exports[`CSN to PostgreSQL Create Statements should return PostgreSQL compatible statements for beershop-admin-service 1`] = `
 "
 CREATE TABLE BeershopAdminService_UserScopes (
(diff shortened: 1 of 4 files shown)


postgres/test/__snapshots__/csn2postgres.test.js.snap
```diff
@@ -42,7 +42,7 @@ CREATE TABLE csw_TypeChecks (
   type_Time TIME,
   type_DateTime TIMESTAMP,
   type_Timestamp TIMESTAMP,
-  type_String VARCHAR(5000),
+  type_String VARCHAR(255),
```
patricebender, line 45: do we need a changelog entry for this?

BobdenOs: This comes from the compiler. So it is expected that most projects already have the latest compiler version. Which should have this as part of its changelog.

stewsk (approved):

Issue #117, opened 2023-07-19
@cap-js/postgres: collation "xx" for encoding "UTF-8" does not exist


sjvans:
## issue

CAP reduces locales to two letters.

MORE TO FOLLOW...

## workaround

create the missing collation:
```sql
CREATE COLLATION en FROM "en_US";
```

danjoa:
The 'fix' PR looks frightening → why don't we simply pass the original 5-letter locale down to postgres in set session context?

sjvans:
> The 'fix' PR looks frightening → why don't we simply pass the original 5-letter locale down to postgres in set session context?

afaik, it is not available from `cds.Request` and i don't think we should reimplement getting it from the express request.

`ICU` should always be there, so in 99% we'll just append `-x-icu`.

BobdenOs:
That is also my understanding the locale will always be normalized to the shortest possible form. Which ICU supports natively. So the easiest way to remove any complex part of the 'fix' PR is to drop libc collation support and rely on ICU. If we find that ICU is 100% of the time available we can even skip the availability check and the fallback sorting function. As ICU is a compile flag for postgres it is technically possible for it to not be there, but we have received the full list of collations available in the oldest postgres version of the hyperscaler postgres service which contains ICU. Which indicates that it could be a realistic assumption for it to always be available or it to be user error.

patricebender:
looks like the issue has been fixed by Sebastian, hence closing.

PR #120, opened 2023-07-19
fix: order by icu OR with (best-effort) mapping OR without collation

sjvans:
re https://github.com/cap-js/cds-dbs/issues/117

### DIFF
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -798,7 +798,7 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
     function resolveCalculatedElement(calcElement) {
       if (alreadySeenCalcElements.has(calcElement)) return
       else alreadySeenCalcElements.add(calcElement)
-      const { ref, /* val, */ xpr, func } = calcElement.value
+      const { ref, xpr, func } = calcElement.value
       if (ref || xpr) {
         attachRefLinksToArg(calcElement.value, { definition: calcElement.parent, target: calcElement.parent }, true)
         // column is now fully linked, now we need to find out if we need to merge it into the join tree
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -798,7 +798,7 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
     function resolveCalculatedElement(calcElement) {
       if (alreadySeenCalcElements.has(calcElement)) return
       else alreadySeenCalcElements.add(calcElement)
-      const { ref, val, xpr, func } = calcElement.value
+      const { ref, /* val, */ xpr, func } = calcElement.value
       if (ref || xpr) {
         attachRefLinksToArg(calcElement.value, { definition: calcElement.parent, target: calcElement.parent }, true)
         // column is now fully linked, now we need to find out if we need to merge it into the join tree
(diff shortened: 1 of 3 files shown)


postgres/lib/PostgresService.js
```diff
@@ -178,26 +177,31 @@ class PostgresService extends SQLService {
...
+              (c.element?.[this.class._localized] ? ` COLLATE "${locale}"` : '') +
               (c.sort === 'desc' || c.sort === -1 ? ' DESC' : ' ASC')
           : c => this.expr(c) + (c.sort === 'desc' || c.sort === -1 ? ' DESC' : ' ASC'),
       )
     }
 
+    orderBy(orderBy, localized) {
+      // best-effort locale expansion (example: 'de' -> 'de_DE')
+      const locale =
+        this.context.locale === 'en'
```
BobdenOs, line 194: If the translation from `de` to `de_DE` is something we want to do it would be better to create a map based upon the available collations in the database.

After determining that the database does not include `icu` it would be possible to do a query like:
```sql
SELECT collname FROM pg_collation WHERE collprovider='c'
```

Which returns all the available `c` collations and it would be possible to use the first two characters as `key` and then assign the collation name as value. The main issue being that the collations don't seem to have a priority defined https://www.postgresql.org/docs/current/catalog-pg-collation.html. Which means that it might use `de_AT` when translating `de`.

Additionally it is problematic to provide `c` support without a way of testing it. Locally I only have 3 `c` provided collate options.

Additionally with the current collation check it is still possible that `icu` and `c` are both missing from the system and the exact same error would comeback it would only say `en_US`/`de_DE` instead of `en`/`de`.

sjvans: yeah, i wanted to save another round trip. but i guess one or two additional selects during startup don't hurt too much...

actually, i think the variant shouldn't matter as we use only for ordering, right? the austrians probably sort the same as the germans and the australians the same as the british, no?

in case both are missing, i'd vote for default rather than an error.

BobdenOs:
@sjvans Could you double check that the changes make sense ?

sjvans:
> @sjvans Could you double check that the changes make sense ?

i have to clue what your mapping selection sql does 😂

but the overall orchestration makes sense!

sjvans:
@BobdenOs @johannes-vogel should we merge this? i cannot approve my own pr 😉

db-service/lib/infer/index.js
```diff
@@ -798,7 +798,7 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
     function resolveCalculatedElement(calcElement) {
       if (alreadySeenCalcElements.has(calcElement)) return
       else alreadySeenCalcElements.add(calcElement)
-      const { ref, val, xpr, func } = calcElement.value
+      const { ref, /* val, */ xpr, func } = calcElement.value
```
BobdenOs, line 801: @patricebender is this fine ?

BobdenOs (commented):
This should resolve the issues we received, but I would prefer if @johannes-vogel @vobu @gregorwolf could provide their opinion on this approach.

db-service/lib/infer/index.js
patricebender: 👍
```suggestion
      const { ref, xpr, func } = calcElement.value
```

sjvans: i just wanted to get rid of the warning, but didn't want to delete stuff from coding i'm not familiar with ;)

BobdenOs (approved):

PR #128, opened 2023-07-22
[APIs] Cleanup API for implementers

danjoa:
This PR is to cleanup SPIs, starting with [`DatabaseService`](https://github.com/cap-js/cds-dbs/pull/128/files#diff-d11382d99cc2b608b41a6efd4461bfd7e730bea8af2eb62b4e3fea9c690659ca), and impls so that we can make them public with clear and simple contracts. E.g. `this.release()` instead of monkey-patched instances.


db-service/lib/common/generic-pool.js
```diff
@@ -14,16 +14,17 @@ function _track_connections4(pool) {
     async acquire() {
       const connections = (this._trackedConnections ??= new Set())
       try {
-        this.dbc = await acquire.call(this)
+        let dbc = await acquire.call(this)
+        connections.add((dbc._beginStack = new Error('begin called from:')))
```
BobdenOs, line 18: This was intentionally done later. As the goal of the connection tracking is as follows:

```js
const connections = [
  cds.tx(),
  cds.tx(),
]
const [a,b] = await promise.all(connections)
// a will connect and fill the pool
// b will throw with the information that a is blocking the connection.
```
For clarity the tracked connections should only consist of successfully acquired connections and not all connections trying to acquire a connection. As they are not the root cause of the acquire timeout error.

db-service/lib/common/DatabaseService.js
```diff
@@ -1,143 +1,142 @@
...
-    this._beginStack = new Error('begin called from:')
-    connections.push(this)
-    this._release = async dbc => {
-      await pool.release(dbc)
-      connections.splice(connections.indexOf(this), 1)
-    }
+    if (!ctx) return this.tx().begin() // REVISIT: Is this correct? When does this happen?
+
+    // REVISIT: tenant should be undefined if !this.isMultitenant
+    // REVISIT: This should be a standard property in this.options
```
danjoa, line 46:
```suggestion
```

db-service/lib/common/generic-pool.js
danjoa: Is that relevant for SQLite only?

patricebender:
looks like the tests time out, what could be the reason?

db-service/lib/common/DatabaseService.js
```diff
@@ -1,143 +1,141 @@
...
   async rollback() {
-    const dbc = this.dbc
-    if (!dbc) return
-    try {
-      await this.send('ROLLBACK')
-    } finally {
-      this._release(dbc)
-    }
+    if (!this.dbc) return
+    else
```
BobdenOs, line 80: no need for the `else`

danjoa: Yes, I sometimes do this for linguistic reasons and better readability (for  ;me)
We can remove it if it has the opposite effect on you, of course

danjoa:
Any chance to get this done? 
As said last week when we reviewed: the longer we wait the harder it gets → 3 conflicts already today

danjoa:
Nicht notwendig für die release, aber halt bald... da es unsere APIs betrifft

BobdenOs (approved):

danjoa:
Now, as we've all the other PRs in, would be great to merge that next as it affects Database Service Provider Interfaces / APIs (= SPIs)

BobdenOs:
If you enable auto merge it will trigger the tests again and if they pass will merge.

PR #129, opened 2023-07-23
[REVISIT:] Simplified Setup

danjoa:
- REVISIT: Including removed prettier task → need to re-discuss
- REVISIT: Why do postgres tests fail locally?
- REVISIT: Why is `npm t` that slow? → only one core?


BobdenOs:
> REVISIT: Why is npm t that slow? → only one core?

Here are some possible causes
1. [Github actions] the runner only has 2 cores [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)
2. [Locally] every `.test.js` file can only run sequentially making `lean-draft.test.js` and `compliance.test.js` often the last running tests.

BobdenOs:
@danjoa here I added an failing test to show what the github jest reporter does.

Adds test error message into files changed page (screenshot as it was removed when the test was removed)
image

Additionally on the action summary it has all the failing test errors listed [summary](https://github.com/cap-js/cds-dbs/actions/runs/5787013348?pr=129).
image

BobdenOs (approved):

Issue #127, opened 2023-07-21
@cap-js/sqlite: search with multiple words leads to error


sjvans:
- customer reports error `SQLite only supports single value arguments for $search`.
- fine on postgres.
- should be avoided by concatenating multiple search clauses with `OR`

danjoa:
Can you add an example?

sjvans:
https://github.com/cap-js/cds-dbs/pull/132

danjoa:
Thanks. We intentionally decided to provide only limited support for that, at least in SQLite, as the complexity and performance overhead would be significant, and the value rather low (most end users don't do advanced search expressions)

ArtyomAD:
Hello.

Can someone please clarify if this issue/missing functionality plan to be released soon?

We might require to disable main Search functionality in our Fiori application which is running on @cap-js/posgres exactly due to this issue.
![$seach from fiori elements app](https://github.com/cap-js/cds-dbs/assets/15830610/77e1ff38-6d1a-44d9-a180-b4cb3294476d)

gregorwolf:
Hi @ArtyomAD,

I would think that it would be better if you check if this issue also occurs when you use sqlite as then you could contribute the SQL statement that is created when you issue the query. Or you create a new issue which mentions the postgres driver.

ArtyomAD:
Hello @gregorwolf.

We face identical error in both cases with sqlite and postgres.

SQLite:
![sqlite logs](https://github.com/cap-js/cds-dbs/assets/15830610/612f8584-1703-4dfe-8a8b-ae940a7ddbec)
Trace without $search:
![sqlite trace without $search](https://github.com/cap-js/cds-dbs/assets/15830610/dfdae09e-8796-459c-8d72-e19c1aff3b70)
Trace with $search:
![sqlite trace with $search](https://github.com/cap-js/cds-dbs/assets/15830610/3b3fcbc3-0d74-40e5-b9ff-5c4ecebf1d15)

Postgres in BTP:
![postgres logs](https://github.com/cap-js/cds-dbs/assets/15830610/d4230226-f60a-4422-abc9-47b05454f2ed)
Confirmation from BTP as it's running with postgres:
![btp postgres db logs](https://github.com/cap-js/cds-dbs/assets/15830610/fea22b97-13ca-4364-b831-b64739a40a7a)

sebastianesch:
[...]

PR #132, opened 2023-07-24
feat: search with multiple values

sjvans:
possible solution for https://github.com/cap-js/cds-dbs/issues/127

- [ ] changelog


db-service/lib/cqn2sql.js
```diff
@@ -505,7 +505,13 @@ class CQN2SQLRenderer {
...
   static Functions = require('./cql-functions')
   func({ func, args }) {
-    args = (args || []).map(e => (e === '*' ? e : { __proto__: e, toString: (x = e) => this.expr(x) }))
+    args = (args || []).map(e => {
+      if (e === '*') return e
+      const arg = { __proto__: e, toString: (x = e) => this.expr(x) }
+      if ('xpr' in arg)
+        arg.xpr = arg.xpr.map(e => (typeof e === 'string' ? e : { __proto__: e, toString: (x = e) => this.expr(x) }))
+      return arg
+    })
```
danjoa, line 514: Why do we need that in addition to the `{xpr}` handling in `Functions.search()`?

db-service/lib/cql-functions.js
```diff
@@ -5,7 +5,11 @@ const StandardFunctions = {
   // length     : (x) => `length(${x})`,
   average: x => `avg(${x})`,
   search: function (ref, arg) {
-    if (!('val' in arg)) throw `SQLite only supports single value arguments for $search`
+    if ('xpr' in arg) {
+      let xpr = ''
+      for (const each of arg.xpr) xpr += typeof each === 'string' ? ` ${each} ` : this.search(ref, each)
```
danjoa, line 10: Don't we need an 'and' for strings?

db-service/lib/cqn2sql.js
sjvans: `this.tolower(arg)` called for xpr element. becomes `[object Object]` instead of `?` in the sql.

db-service/lib/cql-functions.js
sjvans: `$search=foo bar` (= `$search=foo AND bar`)
--> `xpr = [{ val: 'foo' }, 'and', { val: 'bar' }]`
--> sql snippet `x = ? and x = ?`, where `x = ?` is the result of `this.search(ref, each)`

patricebender:
Would be cool to have this, whats speaking against it?

schiwekM:
@danjoa @sjvans any news on this PR? With @cap-js/hana now being released this "regression" is now also introduced that search strings with whitespace in Fiori elements no longer work. In my opinion at least for HANA  /prod scenarios this needs to be enabled by default to not have this regression, which causes frustration on user side.

PR #136, opened 2023-07-25
feat: Dataset streams

BobdenOs:
## Examples

Insert `JSON` stream into table. Like this [file](https://github.com/cap-js/cds-dbs/pull/136/files#diff-a0eb472572b3a2189a19cf94ce0c0834be510ac9bb540aaacf562084eb4d03f0).

```js
const { Images } = cds.entities('test')
const stream = fs.createReadStream(path.join(__dirname, 'samples/data.json'))

const changes = await STREAM.into(Images).data(stream)
expect(changes).toEqual(2)
```

Select `JSON` stream from table. 

```js
const { Images } = cds.entities('test')
await cds.tx(async () => {
  const select = SELECT(['ID', 'data']).from(Images)
  const stream = await STREAM(['ID', 'data']).from(select)
  const result = await streamJson(stream)

  expect(result.length).toEqual(3)

  // Decode LargeBinary columns from base64
  expect(Buffer.from(result[0].data, 'base64').length).toEqual(7891)
  expect(result[1].data).toEqual(null)
  expect(Buffer.from(result[2].data, 'base64').length).toEqual(7891)
})
```

Pipe `JSON` stream into http response.

```js
this.on('READ', '*', async req => {
  const { query } = req // SELECT
  await cds.tx(async tx => {
    // OData only wraps arrays with value
    const wrapper = !query.SELECT.one 
    // Convert to STREAM query
    const stream = await tx.run(STREAM.from(query))

    // Prepare OData response
    if (wrapper) req.res.write(`{"@odata.context":"$metadata#TODO","value":`)
    // Write stream contents onto req.res socket
    for await (const chunk of stream) {
      req.res.write(chunk)
    }
    // Close req.res socket
    if (wrapper) req.res.end(`}`)
    else req.res.end()
  })
})
```

## todo

Adopt to new ql streaming approach, but the sqlite driver implementation should be re usable.

extracted from: #66


sqlite/lib/SQLiteService.js
```diff
@@ -44,13 +44,75 @@ class SQLiteService extends SQLService {
 
   prepare(sql) {
     try {
-      return this.dbc.prepare(sql)
+      const stmt = this.dbc.prepare(sql)
+      stmt._run = stmt.run
+      stmt.run = this._run.bind(stmt)
+      stmt._iterator = this._iterator
+      stmt.stream = this._stream.bind(stmt)
+      return stmt
```
danjoa, line 52: Looks like an ugly workaround for something... → for what?

postgres/lib/session.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$user.id": "CAP.APPLICATIONUSER",
+  "$user.locale": "CAP.LOCALE",
+  "$user.now": "CAP.NOW",
+  "$valid.from": "CAP.VALID_FROM",
+  "$valid.to": "CAP.VALID_TO"
+}
```
danjoa, line 7: We discussed to prefer all-lowercase names (especially for PostgreSQL) → @agoerler did you check whether that is possible without breaking changes? I.e. if `current_setting` is case-insensitive?

db-service/lib/cqn2sql.js
```diff
@@ -130,7 +130,7 @@ class CQN2SQLRenderer {
 
   SELECT(q) {
     let { from, expand, where, groupBy, having, orderBy, limit, one, distinct, localized } = q.SELECT
-    if (!expand) expand = q.SELECT.expand = has_expands(q) || has_arrays(q)
+    expand = q.SELECT.expand = expand ?? (has_expands(q) || has_arrays(q))
```
danjoa, line 133: Please avoid this `foo = foo || bar` pattern and prefer the `if (!foo) foo = bar` one

db-service/lib/cqn2sql.js
```diff
@@ -289,7 +289,7 @@ class CQN2SQLRenderer {
       .map(c => c.sql)
 
     this.entries = [[JSON.stringify(INSERT.entries)]]
-    return (this.sql = `INSERT INTO ${entity}${alias ? ' as ' + this.quote(alias) : ''} (${
+    return (this.sql = `INSERT INTO ${this.quote(entity)}${alias ? ' as ' + this.quote(alias) : ''} (${
```
danjoa, line 292: I'm not in favour of this inflationary use of the quite expensive `this.quote` function → we should plan a separate review on that

db-service/lib/common/DatabaseService.js
```diff
@@ -70,6 +70,10 @@ class DatabaseService extends cds.Service {
         get '$user.locale'() {
           return _set(this, '$user.locale', ctx.locale || cds.env.i18n.default_language)
         },
+        // REVISIT: should be decided in spec meeting for definitive name
+        get '$user.now'() {
+          return _set(this, '$user.now', (ctx.timestamp || new Date()).toISOString())
+        },
```
danjoa, line 76: Likely `$now` as it's already supported in the compiler. 
But why do we need that in the database?

danjoa: P.S. What is the costs for session context variables in HANA and PostgreSQL? → can we add them like hell, or is there a price tag?

db-service/lib/cqn2sql.js
```diff
@@ -423,21 +423,39 @@ class CQN2SQLRenderer {
...
 
   STREAM(q) {
-    let { from, into, where, column, data } = q.STREAM
+    let { from, into, where, column, columns, data } = q.STREAM
     let x, sql
     // reading stream
     if (from) {
       sql = `SELECT`
-      if (!_empty((x = column))) sql += ` ${this.quote(x)}`
+      if (!_empty((x = column))) {
```
danjoa, line 431: The implementation looks increasingly complex → hard to understands and maintain?

danjoa (commented):
I actually didn't understand, what this PR is doing... 👀

postgres/lib/session.json
BobdenOs: It is case insensitive and I will adjust accordingly

sqlite/lib/SQLiteService.js
BobdenOs: This is to support streams as parameters to the `run` function which SQLite doesn't support natively.

danjoa (commented):
Maybe we should have a call tomorrow with @BobdenOs and @sjvans

db-service/lib/cqn2sql.js
```diff
@@ -592,24 +592,64 @@ class CQN2SQLRenderer {
    * @returns {string} SQL
    */
   STREAM(q) {
```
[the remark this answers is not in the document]
BobdenOs: If we support `Readable` as values in `UPDATE`, `INSERT` and `UPSERT`. It should remove the need for `STREAM.into` cases. Which is why I changed the current internal logic of the STREAM handler to just translate them to the other types.

The only use case case I could for the STREAM event is reading a whole result set as raw json stream. This could also be done by adding a flag on the SELECT statement.

Additionally it would be required to adjust the `all` API so that the driver has a way of knowing what columns are of type LargeBinary as they should be returned as streams rather then full 2gib buffers.

sqlite/lib/SQLiteService.js
```diff
@@ -44,13 +44,76 @@ class SQLiteService extends SQLService {
...
   prepare(sql) {
     try {
-      return this.dbc.prepare(sql)
+      const stmt = this.dbc.prepare(sql)
+      return {
+        run: (..._) => this._run(stmt, ..._),
+        get: (..._) => stmt.get(..._),
+        all: (..._) => stmt.all(..._),
+        stream: (..._) => this._stream(stmt, ..._),
+      }
```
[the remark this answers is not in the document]
BobdenOs: Yes. 

The only reason I kept it outside of the prepare was to not create multiple instances of the function. In the hopes that V8 would have an easier timer optimizing it.

Also the reason why get and all don't accept streams is that HANA does not allow LOB types to be used in filters.

db-service/lib/cqn2sql.js
```diff
@@ -592,24 +592,64 @@ class CQN2SQLRenderer {
...
+          .where(where),
+      )
     } else {
-      // writing stream
-      const entity = this.name(q.target?.name || into.ref[0])
-      sql = `UPDATE ${this.quote(entity)}${into.as ? ` AS ${into.as}` : ``} SET ${this.quote(column)}=?`
-      this.entries = [data]
+      data.type = 'json'
+      // REVISIT: decide whether dataset streams should behave like INSERT or UPSERT
+      sql = this.UPSERT(cds.ql.UPSERT([{}]).into(into).forSQL())
```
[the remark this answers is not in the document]
BobdenOs: cqn2sql can't call cqn4sql currently as it does not have the original model argument any more. When doing the implementation for revert-STREAM-event this forSQL call should no longer be needed.

BobdenOs:
@danjoa reverted the `val` change so it still uses placeholders as it used to. Only `session_context` in Postgres will render the static strings when they are part of the known pseudo variables.

danjoa (approved):

PR #137, opened 2023-07-26
update is only noop if not managed

johannes-vogel:
@patricebender @BobdenOs Is there an appropiate test suite I could use for adding a test?


BobdenOs:
I would suggest https://github.com/cap-js/cds-dbs/blob/main/test/scenarios/bookshop/read.test.js#L33
Just have to find an entity without managed fields.

johannes-vogel:
> I would suggest https://github.com/cap-js/cds-dbs/blob/main/test/scenarios/bookshop/read.test.js#L33 Just have to find an entity without managed fields. I guess genres should do the trick.

are those executed in the voter?

BobdenOs (approved):

PR #141, opened 2023-07-28
[PARKED:] feat: deploy only tables

BobdenOs:
## Description

Deploying views takes time. So this PR solves that problem. By not deploying the views.

By not relying on the views it is possible to:

- Use rolling app deployments 
  - (e.g. two versions of the same app using the same persistence)
- Skip deployment when only views are changed
  - (or do schema evolution on tables)
- Host multiple different models in the same app 
  - (e.g. canary tenants)
- Flatten projections
  - (e.g. rename all columns only once)
- Optimize joins away
  - (e.g. SFlight with 7 layers of joins while a single column is selected)
- Cache query optimization in the app layer
  - (removing/reducing execution plan processing times)
- Define more views

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -555,9 +555,7 @@ class CQN2SQLRenderer {
         if (Buffer.isBuffer(val)) val = val.toString('base64')
         else val = this.regex(val) || this.json(val)
     }
-    if (!this.values) return this.string(val)
-    this.values.push(val)
-    return '?'
+    return this.string(val)
   }
 
   static Functions = require('./cql-functions')
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -138,7 +138,7 @@ class CQN2SQLRenderer {
       sql = `SELECT`
     if (distinct) sql += ` DISTINCT`
     if (!_empty((x = columns))) sql += ` ${x}`
-    if (!_empty((x = from))) sql += ` FROM ${this.from(x)}`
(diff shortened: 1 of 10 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -197,7 +197,7 @@ class CQN2SQLRenderer {
       sql = `SELECT`
     if (distinct) sql += ` DISTINCT`
     if (!_empty((x = columns))) sql += ` ${x}`
-    if (!_empty((x = from))) sql += ` FROM ${this.from(x)}`
+    if (!_empty((x = from))) sql += ` FROM ${this.from(x, q)}`
```
danjoa, line 200: btw: we don't always need to enhance args. this.cqn also gives access to top-level query

danjoa (commented):
Looks pretty promising ... → I'm interested to walk through the benefits above in detail in one of our next syncs 

Essentially we always had that for write ops by resolveView. and you add this now for SELECT as well?

db-service/lib/cqn2sql.js
BobdenOs: As the view is replaced with a select the from is applied to nested select queries. So wile it is possible to find the correct parent query of a from in the `this.cqn` it will probably be more complicated.

danjoa: These are cases where using `this.cqn` is not appropriate

BobdenOs:
The first try was just calling resolve view and tender the returned query to SQL, but this did not have the desired results.

By replacing the view with its query it is very simple bit of logic, but makes it basically impossible to create an optimized query. For which a resolve view approach would be better.

PR #142, opened 2023-07-28
update postgres and sqlite dependency to upcoming db-service version

johannes-vogel:


postgres/package.json
```diff
@@ -31,7 +31,7 @@
     "test": "npm run setup && jest --silent"
   },
   "dependencies": {
-    "@cap-js/db-service": "^1.0.1",
+    "@cap-js/db-service": "^1.1.0",
```
patricebender, line 34: could we also do smt like this, so that we dont have to bump the minor version each time?
```suggestion
    "@cap-js/db-service": "^1",
```

danjoa:
Why do we have to do that at all?
semver doesn't require that, does it?

sjvans:
> Why do we have to do that at all? semver doesn't require that, does it?

if you'd do `npm update @cap-js/sqlite`, then you wouldn't get a newer `@cap-js/db-service` if we don't bump, right? so i think it's always needed in case something new in `@cap-js/sqlite` needs something new in `@cap-js/db-service`, otherwise not.

sjvans (approved):

PR #143, opened 2023-07-28
[APIs] Introducing column_alias4 + more

danjoa:
@BobdenOs: as discussed I fixed the [column_alias for nested SELECTs](https://github.com/cap-js/cds-dbs/pull/143/commits/94da5e6c4d18bea2f1116cd2ab7306a6e9bbb408) → the PR would be ready for merge now


danjoa:
@BobdenOs : fails with typescript errors → I've no clue how to fix

db-service/lib/cqn2sql.js
```diff
@@ -266,32 +258,36 @@ class CQN2SQLRenderer {
...
+      return `extensions__->${this.string('$."' + x.element.name + '"')} as ${x.as || x.element.name}`
     }
+    ///////////////////////////////////////////////////////////////////////////////////////
     let sql = this.expr(x)
+    let alias = this.column_alias4(x)
+    if (alias) sql += ' as ' + this.quote(alias)
     return sql
   }
 
+  column_alias4(x) {
```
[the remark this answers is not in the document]
danjoa: Hmm, can we change the ts settings to not require us to document everything → that is not what we should do, will harm our ability to rafactor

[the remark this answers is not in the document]
BobdenOs: The typescript validation step is something we can turn off. It currently only considers APIs exported over the `index.js`, because `CQN2SQL` is part of the `SQLService` it is exported. Which means that all the functions part of the class are exported.

What I found is that when running the validation locally it is very handy to find gaps. So if we turn it off now and turn it on after we have refactored. It will just give a list of functions that require type definitions.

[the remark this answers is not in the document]
danjoa: I really dislike being forced into the corset of a statically typed language now. We would have chosen typescript if we wanted that. → My proposal:

1. When add jsdoc comments these go into .d.ts
2. When we don't the method is considered private 

Following our rules so far: Only the things we document explicitly are public API.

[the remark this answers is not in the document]
danjoa: Also needs a call, I think

db-service/lib/SQLService.js
```diff
@@ -91,7 +91,7 @@ class SQLService extends DatabaseService {
    */
   async onINSERT({ query, data }) {
     const { sql, entries, cqn } = this.cqn2sql(query, data)
-    if (!sql) return // Do nothing when there is nothing to be done
+    if (!sql) return // Do nothing when there is nothing to be done // REVISIT: When does this happen?
```
danjoa, line 94:
```suggestion
    if (!sql) return // Do nothing when there is nothing to be done // REVISIT: fix within mtxs
```

db-service/lib/SQLService.js
```diff
@@ -236,8 +237,7 @@ class SQLService extends DatabaseService {
   cqn2sql(q, values) {
     const cqn = this.cqn4sql(q)
 
-    // REVISIT: disable this for queries like (SELECT 1)
-    // Will return multiple rows with objects inside
+    // REVISIT: Why did we move that from onSELECT to here?
```
BobdenOs, line 240: In onSELECT the query is not yet inferred by cqn4sql.

postgres/lib/PostgresService.js
```diff
@@ -249,18 +249,9 @@ GROUP BY k
...
-      const unique = {}
-      return SELECT.columns
-        .map(x => `${this.column_expr(x)} as ${this.quote(this.column_name(x))}`)
-        .filter(x => {
-          if (unique[x]) return false
-          unique[x] = true
-          return true
-        })
+    column_alias4(x, q) {
+      if (!x.as && 'val' in x) return String(x.val)
```
BobdenOs, line 253: Is there a reason why `column_alias4` has to be overwritten here ? I did a local test run with it removed and the Postgres tests passed.

postgres/lib/PostgresService.js
```diff
@@ -254,6 +254,18 @@ GROUP BY k
       return super.column_alias4(x, q)
     }
 
+    SELECT_columns(q) {
+      if (!q.SELECT.columns) return '*'
+      // REVISIT: sflight tests fail without deduplication of columns
+      // The specific column which is defined twice is "DraftAdministrativeData_DraftUUID"
```
BobdenOs, line 260: @patricebender @David-Kunz any ideas why for these queries the column `DraftAdministrativeData_DraftUUID` is defined double ?

danjoa: Yes, please fix this duplicate columns. 
I actually saw this deduplication code in postgres and intentionally removed it to reveal and fix the root cause instead of going workarounds.

BobdenOs: I have found the root cause. Here is the proposed fix https://github.com/cap-js/cds-dbs/pull/151

postgres/lib/PostgresService.js
```diff
@@ -252,11 +252,16 @@ GROUP BY k
...
+      // REVISIT: possibly always quote all column aliases
+      // REVISIT: adjust all locations that reference column names (e.g. orderBy)
+      // REVISIT: exclude table alias when selecting from single source
+      return SELECT.columns
+        .map(x => `${this.column_expr(x)} as ${this.quote(this.column_name(x))}`)
+        .filter(x => {
+          if (unique[x]) return false
+          unique[x] = true
+          return true
+        })
```
danjoa: return super.SELECT_columns(q).filter(x => {
         if (unique[x]) return false
         unique[x] = true
         return true
       })

danjoa:
@BobdenOs did you have a chance to recheck this one, so we can reduce the number of loose ends?

BobdenOs (approved):

PR #147, opened 2023-08-01
Simplified managed

danjoa:


db-service/lib/cqn2sql.js
```diff
@@ -24,8 +24,8 @@ class CQN2SQLRenderer {
      * @type {import('@sap/cds/apis/services').ContextProperties}
      */
     this.context = cds.context || context
-    // REVISIT: find a way to make CQN2SQLRenderer work in SQLService as well
-    /** @type {CQN2SQLRenderer|unknown} */
+    // REVISIT: find a way to make CQN2SQLRenderer work in SQLService as well -> ???
+    // /** @type {CQN2SQLRenderer|unknown} */ -> that killed IntelliSense
```
arleytm, line 28: `unknown` isn't a valid JSDoc type.

BobdenOs: `unknown` is a typescript base type. It's the opposite of `any`.

PR #148, opened 2023-08-02
feat: forUpdate and forShareLock

BobdenOs:
## Description

`forUpdate` and `forShareLock` was not yet supported as SQLite does not have native support for user controlled locks of any kind. Postgres can do row level locking, but the `forUpdate` API is defined as column level locks using the `of` property. This implementation tries to solve the main issue that `cqn4sql` points the locking queries to the localized views. Which cause Postgres to throw a SQL error as the join condition can be null which is not allowed.


db-service/lib/cqn4sql.js
```diff
@@ -1828,7 +1828,12 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
    * @returns true if the given definition shall be localized
    */
   function isLocalized(definition) {
-    return inferred.SELECT?.localized && definition['@cds.localized'] !== false
+    return (
+      inferred.SELECT?.localized &&
+      definition['@cds.localized'] !== false &&
+      !inferred.SELECT.forUpdate &&
+      !inferred.SELECT.forShareLock
```
danjoa, line 1835: Does that mean that we receive queries like that, with `localized` and `forUpdate` set?: 
```js
SELECT.localized(Foo).foUpdate()
```
→ looks like a glitch in the caller, and should be fixed there?
At least as well

danjoa (approved):

db-service/lib/cqn4sql.js
BobdenOs: This was triggered by `cap/issues/issues/14177` which was facing exactly this scenario. I don't know the exact details, but should probably be followed up with the reporter.

johannes-vogel (approved):

PR #149, opened 2023-08-02
fix(draft): deep update on a managed draft entity without change should update the `modifiedAt` field on the parent entity

arleytm:
## Issue Background

```cds
entity Shipment : managed, cuid {
    shipmentDate : Composition of many ShipmentDate on shipmentDate.owner = $self
}

entity ShipmentDate : cuid, managed {
    owner         : Association to Shipment;
    dateID        : Association to ShipmentDateDesc;
    dateTimeStamp : DateTime
}

entity ShipmentDateDesc {
    key dateID      : Integer;
        description : localized String(240);
}
```

```cds
namespace cgt.filter.pocservice;

using { cgt.filter.poc as shipment } from '../db/schema';

service ManageShipmentService {
  annotate Shipment with  @odata.draft.enabled  @Common.SemanticKey: [ID];
  entity Shipment as projection on shipment.Shipment;

  entity ShipmentDate as projection on shipment.ShipmentDate {
    ID,
    owner : redirected to Shipment,
    dateID : redirected to ShipmentDateDesc,
    dateTimeStamp as timestamp,
    modifiedAt
  };

  entity ShipmentDateDesc as projection on shipment.ShipmentDateDesc;
}
```

In a draft scenario, after editing a draft-enabled entity with child entities and activating the draft without any modifications, the `modifiedAt` field of the parent entity should be updated. This means that when the user opens the Fiori Element-based application (List Report Floorplan), clicks on a list item to navigate to the Fiori Object Page Floorplan, and then presses the "Edit" button to edit the draft, if no changes are made to the child entities, and the "Save" button is pressed to activate the draft, the `modifiedAt` field should be modified.

This behavior remains consistent with both the old draft and the new lean draft implementations when utilizing the default database layer.

This pull request aims to align the current implementation with the expected behavior outlined above.


db-service/lib/deep-queries.js
```diff
@@ -240,7 +240,7 @@ const _getDeepQueries = (diff, target) => {
       queries.push(INSERT.into(target).entries(diffEntry))
     } else if (op === 'delete') {
       queries.push(DELETE.from(target).where(diffEntry))
-    } else if (op === 'update' || (op === undefined && subQueries.length && _hasManagedElements(target))) {
+    } else if (op === 'update' || (op === undefined && _hasManagedElements(target))) {
```
BobdenOs, line 243: This doesn't just impact drafts. This check was extra added to only update the parent if children where changed, because there are tests which expect exactly that.

So we have to make sure whether those tests are having the wrong expectation. As it is now no longer allowed to differentiate between draft and non draft entities.

@David-Kunz @johannes-vogel could you provide some clarification on the expected behavior?

arleytm: > This check was extra added to only update the parent if children where changed, because there are tests which expect precisely that.

The expectation regarding the deep update scenario in the draft context was outlined in the PR and discussed within the team. In a draft-enabled deep update scenario, the parent entity must always be updated after an edit operation, regardless of any changes made to the children's data. This decision was taken long ago, and to my knowledge, it has always been the case when utilizing the default database layer.

So the question is whether the outlined behavior is also expected in a non-draft deep update scenario.

BobdenOs: > So the question is whether the outlined behavior is also expected in a non-draft deep update scenario.

Exactly

danjoa:
@arleytm could you give an example of a deep update request payload for this use case?

arleytm:
> could you give an example of a deep update request payload for this use case?

You can take a look at this integration test in the CDS runtime located at `tests/_runtime/odata/__tests__/integration/draft-on-edit.test.js`.

```js
    test('it should not update the modifiedAt element without a change', async () => {
      let responseParent = await request(app)
        .get('/Parents(ID=68a34c2c-53c8-11ea-8d77-2e728ce88126,IsActiveEntity=true)')
        .set('Accept', 'application/json')
        .set('authorization', 'Basic dXNlci4xMjM6cGFzc3dvcmQ=')

      let responseChild = await request(app)
        .get('/Parents(ID=68a34c2c-53c8-11ea-8d77-2e728ce88126,IsActiveEntity=true)/compMany')
        .set('Accept', 'application/json')
        .set('authorization', 'Basic dXNlci4xMjM6cGFzc3dvcmQ=')

      const { modifiedAt: modifiedAtParent } = responseParent.body
      const { modifiedAt: modifiedAtChild } = responseChild.body.value[0]

      await request(app)
        .post(
          '/Parents(ID=68a34c2c-53c8-11ea-8d77-2e728ce88126,IsActiveEntity=true)/service.DraftAlias.draftEdit?$expand=compMany'
        )
        .send({ PreserveChanges: true })
        .set('Content-Type', 'application/json')
        .set('Accept', 'application/json')
        .set('authorization', 'Basic dXNlci4xMjM6cGFzc3dvcmQ=')
        .expect(201)

      await request(app)
        .post('/Parents(ID=68a34c2c-53c8-11ea-8d77-2e728ce88126,IsActiveEntity=false)/service.DraftAlias.draftActivate')
        .send({})
        .set('Content-Type', 'application/json')
        .set('Accept', 'application/json')
        .set('authorization', 'Basic dXNlci4xMjM6cGFzc3dvcmQ=')
        .expect(200)

      responseParent = await request(app)
        .get('/Parents(ID=68a34c2c-53c8-11ea-8d77-2e728ce88126,IsActiveEntity=true)')
        .set('Accept', 'application/json')

      responseChild = await request(app)
        .get('/Parents(ID=68a34c2c-53c8-11ea-8d77-2e728ce88126,IsActiveEntity=true)/compMany')
        .set('Accept', 'application/json')

      expect(responseParent.status).toBe(200)

      // FIXME <———————————————————————————————————-
      if (process.env.CDS_ENV !== 'better-sqlite') {
        expect(responseParent.body.modifiedAt).not.toStrictEqual(modifiedAtParent)
      }

      expect(responseChild.status).toBe(200)
      expect(responseChild.body.value[0].modifiedAt).toStrictEqual(modifiedAtChild)
    })
 ```

File: db-service/lib/deep-queries.js (hunk 1 above)
Comment (arleytm) [reply db-service/lib/deep-queries.js:243]: @BobdenOs

> because there are tests which expect exactly that.

Could you point out which runtime integration tests expect the opposite behavior? Why are the tests not already failing?

The code in question was added in this PR without any tests:
https://github.com/cap-js/cds-dbs/pull/15

Comment (BobdenOs) [reply db-service/lib/deep-queries.js:243]: Found the test that fails `cds/tests/_runtime/odata/__tests__/integration/etag.test.js` -> `composition to many`

Which expects that children with managed data that are not touched should not change.

File: db-service/lib/deep-queries.js [hunk 2]
```diff
@@ -189,7 +189,7 @@ const getDeepQueries = (query, dbData, target) => {
     diff = [diff]
   }
 
-  return _getDeepQueries(diff, target)
+  return _getDeepQueries(diff, target, true)
```
arleytm, line 192: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

BobdenOs (approved):

Issue #139, opened 2023-07-28
[cds] - error: cannot pass more than 100 arguments to a function


cforkin:
### Description of erroneous behaviour 

I am running the latest versions of CAP and @cap-js/postgres (see excerpt from `package.json` below)

```json
  "dependencies": {
    "@cap-js-community/odata-v2-adapter": "^1.11.4",
    "@cap-js/postgres": "^1.0.1",
    "@sap/cds": "^7",
    "express": "^4",
    "express-http-proxy": "^1.6.3"
  },
  "devDependencies": {
    "@cap-js/sqlite": "^1",
    "@sap/ux-specification": "^1.108.10"
  },
```

When executing a SELECT I'm hitting a hard limitation in the PostgreSQL function `json_build_object` (100 arguments to a function). This seems to be a hard-coded limitation within the PG client library

```text
[cds] - error: cannot pass more than 100 arguments to a function
    at test1/node_modules/pg/lib/client.js:526:17
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.all (test1/node_modules/@cap-js/postgres/lib/PostgresService.js:119:26)
    at async PostgresService.onSELECT (test1/node_modules/@cap-js/db-service/lib/SQLService.js:66:16)
    at async next (test1/node_modules/@sap/cds/lib/srv/srv-dispatch.js:79:17)
    at async next (test1/node_modules/@sap/cds/lib/srv/srv-dispatch.js:79:17)
    at async PostgresService.handle (test1/node_modules/@sap/cds/lib/srv/srv-dispatch.js:77:10)
    at async ReportService.<anonymous> (test1/node_modules/@sap/cds/libx/_runtime/common/generic/crud.js:67:16)
    at async next (test1/node_modules/@sap/cds/lib/srv/srv-dispatch.js:79:17)
    at async ReportService.handle (test1/node_modules/@sap/cds/lib/srv/srv-dispatch.js:77:10) {
  length: 118,
  severity: 'ERROR',
  code: '54023',
  detail: undefined,
  hint: undefined,
  position: '8',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_func.c',
  line: '136',
  routine: 'ParseFuncOrColumn',
[...]

PR #150, opened 2023-08-03
fix: column count limitation in Postgres Service

BobdenOs:
## Description

In the original implementation `json_build_object` was used to create the json objects. This function requires key value pairs as arguments to create the json object. This has the benefit that the property names are strings, but comes with the limitation that it is a function call with a maximum of 100 arguments.

This fix switches to using `row_to_json` as this function call only has a single parameter it should not be limited by the arguments limit. It is required to add a sub select to ensure that the property names are correct. As by default the column aliases are minimized to only alias when needed and only quote the alias when needed. This would result in property names being most of the time lower cased, but CAP is case sensitive which means that all columns require quoted aliases. To keep this change simple this is done in an additional sub select. 

## Example query

before

```sql
SELECT json_build_object('ID', Books.ID, 'title', Books.title) as _json_ from (
  SELECT ID, title FROM Books
) as Books
```

after

```sql
SELECT row_to_json(Books.*) as _json_ from (
  SELECT
    Books.ID as "ID",
    Books.title as "title"
  FROM (
    SELECT ID, title FROM Books
  ) as Books
) as Books
```

## Optimizations

It would be possible to optimize the query size by adjusting `SELECT_columns` and force quote all columns and inject the count columns when the query is an expand query.

## Related

reported issue: https://github.com/cap-js/cds-dbs/issues/139

### DIFF
--- a/postgres/lib/PostgresService.js
+++ b/postgres/lib/PostgresService.js
@@ -268,19 +268,24 @@ GROUP BY k
       const queryAlias = this.quote(SELECT.from?.as || (SELECT.expand === 'root' && 'root'))
       const cols = SELECT.columns.map(x => {
         const name = this.column_name(x)
-        let col = `${this.string(name)},${this.output_converter4(x.element, queryAlias + '.' + this.quote(name))}`
+        const outputConverter = this.output_converter4(x.element, `${queryAlias}.${this.quote(name)}`)
+        let col = `${outputConverter} as ${this.doubleQuote(name)}`
 
         if (x.SELECT?.count) {
           // Return both the sub select and the count for @odata.count
           const qc = cds.ql.clone(x, { columns: [{ func: 'count' }], one: 1, limit: 0, orderBy: 0 })
-          col += `, '${name}@odata.count',${this.expr(qc)}`
+          col += `,${this.expr(qc)} as ${this.doubleQuote(`${name}@odata.count`)}`
         }
         return col
       })
-      let obj = `json_build_object(${cols})`
+      let obj = `row_to_json(${queryAlias}.*)`
       return `SELECT ${
         SELECT.one || SELECT.expand === 'root' ? obj : `coalesce(json_agg(${obj}),'[]'::json)`
-      } as _json_ FROM (${sql}) as ${queryAlias}`
+      } as _json_ FROM (SELECT ${cols} FROM (${sql}) as ${queryAlias}) as ${queryAlias}`
+    }
+
+    doubleQuote(name) {
+      return `"${name.replace(/"/g, '""')}"`
     }
 
     INSERT(q, isUpsert = false) {


danjoa:
Do we need this two-fold nested SELECT or could we also go with this?:
```sql
SELECT row_to_json(Books.*) as _json_ from (
    SELECT ID, title FROM Books
)
```

danjoa:
And: is there something similar for SQLite?

BobdenOs:
The additional sub select does the following:
- Always alias the column with quotes to make the property name case sensitive
- Add the output converters where it applies
- Add count properties for expand columns with count

For SQLite there is no equivalent function

danjoa:
Ok, understood... so, could we leave out the inner select in favour of that?:

```sql
SELECT row_to_json(Books.*) as _json_ from (
  SELECT
    "ID",
    "title", 
    (SELECT count(*) from Books) as "$count" -- or alike in case of nested selects with $count
  FROM Books
) as Books
```

danjoa:
> I can adjust the PR to remove the added sub select. It might make `SELECT_columns` quite messy.

We can also do this subsequently, but pls create a REVISIT item for that then → it looks worth it. 
Combine that with the REVISITs for `this.quote()` and `this.column_alias4()`. Maybe we end up with a conclusion like that:

1. Because we anyways need to ensure quoted names for constructing JSONs...
2. We always read columns with quoted names, i.e.: 
   ```sql
   SELECT Books."ID", <expr> as "bar" , <val> as "bar" from Books JOIN Authors ...
   ```
4. For (sub) selects w/o joins we can skip the table aliases:
   ```sql
   SELECT "ID", <expr> as "bar" , <val> as "bar" from Books
   ```

danjoa:
A minor addition to above:  Do we need the outer table alias or is that also supported:
```sql
SELECT row_to_json(*) as _json_ from (
  SELECT... FROM Books
)
```

BobdenOs:
Postgres requires for every sub select an alias.

danjoa (approved):

BobdenOs:
@danjoa I have adjusted the code to not create the sub select. It has one failing test case because of the following SELECT.

```sql
SELECT
    row_to_json(Bookings.*) as _json_
FROM
    (
        SELECT
            count(distinct Bookings.ID) as "countBookings",
            Bookings.airline as "airline",
            Bookings.airlineName as "airlineName"
        FROM
            localized_AnalyticsService_Bookings as Bookings
        GROUP BY
            Bookings.airline,
            Bookings.airlineName
        ORDER BY
            countBookings DESC
        LIMIT
            1 OFFSET 0
    ) as Bookings
```

Because `countBookings` is quoted in the columns it is not found in the order by clause where it is not quotes.

danjoa:
> Because `countBookings` is quoted in the columns it is not found in the order by clause where it is not quotes.

Ok, means we would have to make `cqn2sql.orderBy()` use quoted columns as well when they refer to result set columns. 
Will likely not be that easy to add in a hurry... → Maybe revert that change and add that to [revisit above](https://github.com/cap-js/cds-dbs/pull/150#issuecomment-1663721798)

PR #151, opened 2023-08-03
fix: Always deduplicate columns

BobdenOs:
## Description

When a SELECT query is defined with the same column in different methods without explicit alias they are not deduplicated.

## Example

original

```cql
SELECT DraftAdministrativeData, DraftAdministrativeData_DraftUUID, TravelUUID FROM TravelService.Travel.drafts
```

result

```cql
SELECT DraftAdministrativeData_DraftUUID, DraftAdministrativeData_DraftUUID, TravelUUID FROM TravelService.Travel.drafts
```

expected

```cql
SELECT DraftAdministrativeData_DraftUUID, TravelUUID FROM TravelService.Travel.drafts
```

Query is created by `complementaryDrafts`


stewsk:
Hi @BobdenOs , could you add a test?

db-service/lib/cqn4sql.js
```diff
@@ -430,10 +430,11 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
 
       if (col.$refLinks.some(link => link.definition._target?.['@cds.persistence.skip'] === true)) return
 
+      const getName = col => col?.as || col?.ref?.at?.(-1)
```
danjoa, line 433:
```suggestion
      const getName = col => col.as || col.ref?.at(-1)
```
According to former impl `col` was always defined.
And if `col.ref` exists, function `col.ref.at()` exists as well.

db-service/lib/cqn4sql.js
```diff
@@ -430,10 +430,11 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
...
       const flatColumns = getFlatColumnsFor(col, { baseName, columnAlias, tableAlias })
       flatColumns.forEach(flatColumn => {
-        const { as } = flatColumn
-        if (!(as && transformedColumns.some(inserted => inserted?.as === as))) transformedColumns.push(flatColumn)
+        const name = getName(flatColumn)
+        if (!transformedColumns.some(inserted => getName(inserted) === name)) transformedColumns.push(flatColumn)
```
danjoa, line 437: We should make `transformedColumns` an object instead of an array. Then add to it by the like of:
```suggestion
        if (!(name in transformedColumns)) transformedColumns[name] = flatColumn
```
But after all: can we really just silently skip conflicting column names? 
What about that?: 
```sql
SELECT foo, bar as foo from Boo;
```
```

**Comment by danjoa** [comment]
@David-Kunz: pls double-check and fix the source of this query:
```sql
SELECT 
   DraftAdministrativeData, -- either remove this
   DraftAdministrativeData_DraftUUID, -- or remove this
   TravelUUID 
FROM TravelService.Travel.drafts
```

db-service/lib/cqn4sql.js
BobdenOs: I am not certain whether this behavior is ok or not, but my expectation is that the `foo, bar as foo` example would be caught by infer [here](https://github.com/cap-js/cds-dbs/blob/main/db-service/lib/infer/index.js#L292).

stewsk: If we let
```
SELECT DraftAdministrativeData, DraftAdministrativeData_DraftUUID, TravelUUID FROM TravelService.Travel.drafts
```
pass then the result data must be present both in the structure `DraftAdministrativeData` and in the plain field `DraftAdministrativeData_DraftUUID`
I think simply removing one of them is not an option.

BobdenOs: Currently `DraftAdministrativeData` turns into `DraftAdministrativeData_DraftUUID` which would indicate that there is no difference. If `DraftAdministrativeData` was an expand it would be different.

BobdenOs: Additionally in the test I added it is required to compile the model for `OData` as then the foreign keys can be selected. Where if the model is not recompiled the foreign key columns are not present in the model and can not be selected. So this might just be a side effect from the decision that it is allowed to explicitly select generated foreign key columns.

danjoa (approved):
Let's merge this to unblock the other PRs, but please follow up on 

1. the root causes in lean draft in addition → @David-Kunz 
2. the question of whether we really should do these silent skipping of duplicates, resulting from flattening → @stewsk .

PR #152, opened 2023-08-03
[REVISIT:] Using cds.context

danjoa:
As discussed in the review of #128 passing `this.context` when construction instances of `this.class.CQN2SQL` should not be necessary. Seems to have been added for some specific MTXs behaviour? → need to double-check


db-service/lib/SQLService.js
```diff
@@ -243,7 +243,7 @@ class SQLService extends DatabaseService {
       let target = q[cmd]._transitions?.[0].target
       if (target) q.target = target // REVISIT: Why isn't that done in resolveView?
     }
-    return new this.class.CQN2SQL(this.context).render(q, values) // REVISIT: Why do we need to pass in this.context? -> using cds.context down there should be fine, isn't it?
+    return new this.class.CQN2SQL.render(q, values) // REVISIT: Why do we need to pass in this.context? -> using cds.context down there should be fine, isn't it?
```
danjoa, line 246:
```suggestion
    return (new this.class.CQN2SQL).render(q, values) // REVISIT: Why do we need to pass in this.context? -> using cds.context down there should be fine, isn't it?
```

johannes-vogel:
This was my PR to change this: https://github.com/cap-js/sql/pull/235.
Stakeholders do manual transacation outside of any request context in test coding and expect that this is handled correctly.

danjoa:
> This was my PR to change this: [cap-js/sql#235](https://github.com/cap-js/sql/pull/235).
> Stakeholders do manual transacation outside of any request context in test coding and expect that this is handled correctly.

Ok, thanks → so I reverted the original changes in this PR to return using `srv.context` plus added comments why we do so

Issue #156, opened 2023-08-07
PostgreSQL Performance Improvement Request - Usage of "is not distinct from"


davidtillfcg:
### Queries to read and update, even by key, use "is not distinct from" which is performance intensive.

> Tested with CAP 7.1.1

> It seems in generating the queries to the DB, this part [here](https://github.com/cap-js/cds-dbs/blob/bfe4749c69f0fb41cfcd10bb0768c611ddd445b2/postgres/lib/PostgresService.js#L295) transforms the '=' operator to a "is not distinct from" statement.

```js
 if (x === '=') return 'is not distinct from'
```

> When running a simple query like

```sh
GET /odata/v4/order/orders?$filter=customerAccountNumber eq '12345' 


Subquery Scan on orders  (cost=22316.11..22316.13 rows=1 width=32) (actual time=174.156..180.259 rows=1 loops=1)
  ->  Limit  (cost=22316.11..22316.11 rows=1 width=94) (actual time=174.133..180.234 rows=1 loops=1)
        ->  Sort  (cost=22316.11..22316.11 rows=1 width=94) (actual time=174.132..180.232 rows=1 loops=1)
              Sort Key: orders_0.order_uuid
              Sort Method: quicksort  Memory: 25kB
              ->  Gather  (cost=1000.00..22316.10 rows=1 width=94) (actual time=174.122..180.224 rows=1 loops=1)
                    Workers Planned: 2
                    Workers Launched: 2
                    ->  Parallel Seq Scan on app_orders_orders orders_0  (cost=0.00..21316.00 rows=1 width=94) (actual time=128.641..163.187 rows=0 loops=3)
                          Filter: (NOT ((customeraccountnumber)::text IS DISTINCT FROM '203998123'::text))
                          Rows Removed by Filter: 346816
Planning Time: 0.185 ms
Execution Time: 180.303 ms
```

For example this table has 1 million records, it's large but not outside of possibilities.  This same query if you simply check field = value is 35ms. 

For those who might say just use keys, or this isn't realistic, note that even using the key, the same query is generated but it has a "limit 1" added to it. So it is still bad, but would depend on where in the table your record is. Same for PATCH/update requests, the update where condition uses is not distinct from. 

### Detailed steps to reproduce

> Setup any CAP project with PostgreSQL with the DB
> Observe the generated SQL query on filtering by a field or by the key

### Suggestion
[...]

PR #157, opened 2023-08-07
feat: Reduce distinct from usage - part 1

BobdenOs:
## Description

When using the `=` and `!=` cqn operator it should consider `NULL` as equal to `NULL`. Which means that for `SQL` it is required to use other operators then `=` and `!=`. For SQLite it is `IS` and `IS NOT`. For Postgres it is `is not distinct from` and `is distinct from`. Which in certain cases does not apply optimizations that `=` and `!=` have on the database.

With this change it is checked that both sides of the operation cannot contain `NULL` as result. Which allows for the usage of `=` and `!=`. This will mostly only be true when doing simple expressions like `... WHERE ID = 201`. For anything more complex like `... WHERE ID = (200 + 1)` it won't be capable of guaranteeing that both side will never be `NULL`.

> Why do both sides have to not be `NULL`

When an `=` or `!=` operator encounter a `NULL` on either side it will return `NULL`. This can cascade to any further operators.

```SQL
SELECT * FROM Books WHERE NOT(Books.ID = NULL) -- Books.ID = NULL -> NULL -> NOT(NULL) -> NULL -> false
SELECT * FROM Books WHERE NOT(Books.ID IS NULL) -- Books.ID IS NULL -> false -> NOT(false) -> true
```

## Issues

Reported by: https://github.com/cap-js/cds-dbs/issues/156

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -701,9 +701,10 @@ class CQN2SQLRenderer {
    * @returns {string} The correct operator string
    */
   operator(x, i, xpr) {
-    if (x === '=') return this.operator_has_null(i, xpr) ? 'is' : '='
-    if (x === '!=') return this.operator_has_null(i, xpr) ? 'is not' : '!='
-    else return x
+    // Convert into ANSI SQL NULL operators
+    if (x === '=' && xpr[i + 1]?.val === null) return 'is'
+    if (x === '!=' && xpr[i + 1]?.val === null) return 'is not'
+    return x
   }
 
   /**
@@ -726,9 +727,9 @@ class CQN2SQLRenderer {
     // Also catches when the val is null
     if (!left || !right) return true
 
-    left = left.key || left.notNull || left != null
-    right = right.key || right.notNull || left != null
-    return left && right
+    left = !(left.key || left.notNull)
+    right = !(right.key || right.notNull)
+    return left || right
   }
 
   /**
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -701,8 +701,8 @@ class CQN2SQLRenderer {
     return x
     function _not_null(operand) {
       if (!operand) return
-      operand = operand.element || operand.val
-      return operand?.key || operand?.notNull
+      const element = operand.element
+      return element?.key || element?.notNull || operand.val != null
     }
   }
 
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -695,9 +695,34 @@ class CQN2SQLRenderer {
    * @returns {string} The correct operator string
    */
   operator(x, i, xpr) {
-    if (x === '=' && xpr[i + 1]?.val === null) return 'is'
-    if (x === '!=') return 'is not'
+    if (x === '=') return this.operator_has_null(i, xpr) ? 'is' : '='
+    if (x === '!=') return this.operator_has_null(i, xpr) ? 'is not' : '!='
     else return x
+  }
+
+  /**
+   * Tests if the operator could contain a null value
+   * @param {Number} i Current index of the operator inside the xpr
+   * @param {import('./infer/cqn').predicate[]} xpr The parent xpr in which the operator is used
+   * @returns {boolean}
+   */
+  operator_has_null(i, xpr) {
+    let left = xpr[i - 1]
+    let right = xpr[i + 1]
+
+    // When there is no expression abort
+    if (!left || !right) return true
+
+    left = left.element || left.val
+    right = right.element || right.val
+
+    // When the type is not known abort
+    // Also catches when the val is null
+    if (!left || !right) return true
+
+    left = left.key || left.notNull || left != null
+    right = right.key || right.notNull || left != null
+    return left && right
   }
 
   /**
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -696,36 +696,18 @@ class CQN2SQLRenderer {
    * @returns {string} The correct operator string
    */
   operator(x, i, xpr) {
-    // Convert into ANSI SQL NULL operators
-    if (x === '=' && xpr[i + 1]?.val === null) return 'is'
-    if (x === '!=' && xpr[i + 1]?.val === null) return 'is not'
+    if (x === '=')  return xpr[i + 1]?.val === null ? 'is'     : _not_null(xpr[i-1]) && _not_null(xpr[i+1]) ? '='  : this.is_
+    if (x === '!=') return xpr[i + 1]?.val === null ? 'is not' : _not_null(xpr[i-1]) && _not_null(xpr[i+1]) ? '<>' : this.is_not_
     return x
+    function _not_null(operand) {
+      if (!operand) return
+      operand = operand.element || operand.val
+      return operand?.key || operand?.notNull
+    }
   }
 
-  /**
-   * Tests if the operator could contain a null value
-   * @param {Number} i Current index of the operator inside the xpr
-   * @param {import('./infer/cqn').predicate[]} xpr The parent xpr in which the operator is used
-   * @returns {boolean}
-   */
-  operator_has_null(i, xpr) {
-    let left = xpr[i - 1]
-    let right = xpr[i + 1]
-
-    // When there is no expression abort
-    if (!left || !right) return true
-
-    left = left.element || left.val
-    right = right.element || right.val
-
-    // When the type is not known abort
-    // Also catches when the val is null
-    if (!left || !right) return true
-
-    left = !(left.key || left.notNull)
-    right = !(right.key || right.notNull)
-    return left || right
-  }
+  get is_() { return 'is' }
+  get is_not_() { return 'is not' }
 
   /**
    * Renders an argument place holder into the SQL for prepared statements
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -702,10 +702,15 @@ class CQN2SQLRenderer {
     function _not_null(operand) {
       if (!operand) return false
       if (operand.val != null) return true // non-null values are not null
-      let element = operand.element
-      if (!element) return false
-      if (element.key) return true // primary keys usually should not be null
-      if (element.notNull) return true // not null elements cannot be null
+
+      // REVISIT: The below cannot be merged yet due to a glitch in cqn4sql
+      // which erroneously assigns the definition of Genre.ID as element to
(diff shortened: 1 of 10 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -701,11 +701,36 @@ class CQN2SQLRenderer {
    * @returns {string} The correct operator string
    */
   operator(x, i, xpr) {
-    if (x === '=' && xpr[i + 1]?.val === null) return 'is'
-    if (x === '!=') return 'is not'
+    if (x === '=') return this.operator_has_null(i, xpr) ? 'is' : '='
+    if (x === '!=') return this.operator_has_null(i, xpr) ? 'is not' : '!='
```
danjoa, line 705: That only cares for converting invalid native SQL expressions into valid `is null` / `is not null` expressions. 
For example:

```sql
SELECT 1 where 1 = null;
SELECT 1 where 1 != null;
```
→ 
```sql
SELECT 1 where 1 is null;
SELECT 1 where 1 is not null;
```

Note that this is plain old ANSI SQL, not to be confused with `is [not] distinct from` operators.

postgres/lib/PostgresService.js
```diff
@@ -290,11 +290,11 @@ GROUP BY k
       return ret === '?' ? `$${this.values.length}` : ret
     }
 
-    operator(x) {
+    operator(x, i, xpr) {
       if (x === 'regexp') return '~'
-      if (x === '=') return 'is not distinct from'
-      if (x === '!=') return 'is distinct from'
-      else return x
```
danjoa, line 297: It could be part of the problem, that we completely override the base class impl, hence loosing the `is null` / `is not null` handling. In particular, the example above: 

```sql
SELECT 1 where 1 = null;
SELECT 1 where 1 != null;
```

... would not resolve to this, as it should:

```sql
SELECT 1 where 1 is null;
SELECT 1 where 1 is not null;
```

... but to this, which might be less efficient:

```sql
SELECT 1 where 1 is not distinct from null;
SELECT 1 where 1 is distinct from null;
```

postgres/lib/PostgresService.js
```diff
@@ -290,11 +290,11 @@ GROUP BY k
...
+      if (x === '=') return this.operator_has_null(i, xpr) ? 'is not distinct from' : '='
+      if (x === '!=') return this.operator_has_null(i, xpr) ? 'is distinct from' : '!='
+      return x
```
danjoa, line 297: Keep calling `super.operator()`
```suggestion
      let y = super.operator('x', i, xpr)
      if (y !== x) return y
      if (x === '=') return 'is not distinct from'
      if (x === '!=') return 'is distinct from'
      else return x
```

danjoa (commented):
I think part of the issues is that we did override the base implementation completely...

db-service/lib/cqn2sql.js
BobdenOs: I have adjusted the logic to be ANSI SQL compliant. As far as I could find `IS NULL` and `IS NOT NULL` are the ANSI defined NULL operators. Where before the logic did not check the right hand side for the `!=` operator. Now it makes sure that the right hand side is always `{val:null}` to ensure that it would generate valid ANSI SQL.

danjoa:
Tried to simplify the implementation in https://github.com/cap-js/cds-dbs/pull/157/commits/1d6f08009141116cadd75f43e6cc243433e7adb8 → and if all that is correct, it reveals by the failing tests that the former implementation was incorrect, for example: 

Expected: 
```sql
SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE Foo.ID = ?
```
**But:** the value passed for `?` could be null

Expected: 
```sql
SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE Foo.x = :7
```
**But:** `Foo.x` could be null

danjoa:
In general, I'd still like to double-check whether the initially reported issue wasn't simply because we lost standard ANSI `is null` and `is not null` for PostgreSQL. 

And if so, if we really need these additional optimisations for elements marked as `key` or `not null`.

db-service/lib/cqn2sql.js
```diff
@@ -696,36 +696,18 @@ class CQN2SQLRenderer {
...
   operator(x, i, xpr) {
-    // Convert into ANSI SQL NULL operators
-    if (x === '=' && xpr[i + 1]?.val === null) return 'is'
-    if (x === '!=' && xpr[i + 1]?.val === null) return 'is not'
+    if (x === '=')  return xpr[i + 1]?.val === null ? 'is'     : _not_null(xpr[i-1]) && _not_null(xpr[i+1]) ? '='  : this.is_
+    if (x === '!=') return xpr[i + 1]?.val === null ? 'is not' : _not_null(xpr[i-1]) && _not_null(xpr[i+1]) ? '<>' : this.is_not_
     return x
+    function _not_null(operand) {
+      if (!operand) return
+      operand = operand.element || operand.val
```
BobdenOs, line 704:
```js
operand = operand.element || !(operand.val == null)
```
If `val` is `undefined` it would need to be something else. When the `val` is `0` or `''` it should should be not null. As we don't cache our prepared statements all instances of `{val:null}` will be checked and written into the `SQL`.

BobdenOs:
This optimization is not related to the ANSI SQL NULL checks. As the main target of the optimization is to have faster key look ups. Which specifically will never be `NULL`.

```sql
UPDATE ... WHERE ID = 201
UPDATE ... WHERE ID is 201
```

This optimization will also improve the logic for HANA

db-service/lib/cqn2sql.js
danjoa: Trying to understand... any specific reason for `!(operand.val == null)` vs `operand.val != null`?

danjoa: Btw: I took this `operand = operand.element || operand.val` from the original implementation:

```js
    left = left.element || left.val
    right = right.element || right.val
```

( So I thought at least )

BobdenOs: Mostly to be on the safe side. As I know for certain that `undefined == null` returns `true`, but was not sure wehter `undefined != null` would return `false`. Double checked it and they do behave the same.

BobdenOs: The original implementation was also wrong, because it did not consider the `0` and `''` cases.

BobdenOs: This PR shows how a small optimization like this can easily break things. Which is also why the original implementation was intentionally kept dumb.

danjoa:
> This optimization is not related to the ANSI SQL NULL checks. ... 

Understood. What I was saying above and here is: 

1. The current main impl always does `IS [NOT] DISTINCT FROM NULL` instead of using ANSI SQL's `IS [NOT] NULL` for postgres. 
2. And I'm asking myself if not that already caused performance degradation 
3. And if so, whether the additional optimisations for the `key`s / `not null` cases have a reasonable effect on top at all

BobdenOs:
> the result of a very simple query changes from a fluctuating 300-500ms response to ~60ms pretty constant. With ~30ms overhead in networking, this is really awesome.

My understanding of what was reported by https://github.com/cap-js/cds-dbs/issues/156. Is that when using the `=` and `!=` operation on Postgres it uses the index properly. While `IS [NOT] DISTINCT FROM NULL` always does a full table scan. So cutting the time for all key based where clauses down by ~80% sounds like a worth while optimization.

BobdenOs:
@danjoa I remember now why these queries expect `=` and `!=` instead of `IS` and `IS NOT`. The ANSI SQL does not have the generic `IS` and `IS NOT` operator. So to be ANSI compliant the `=` and `!=` operators have to be used. Even when these will behave incorrectly.

```sql
SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE Foo.ID = ? -- ANSI
SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE Foo.ID IS ? -- SQLite
```

I was not able to find an official link to the `F481` specifications, but here are the closest reference I could find:
[HANA](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/09940a1ed38c4ceb80770632997b286b.html?q=ANSI%20SQL%20Compliance%20Compliance%20With%20Mandatory%20ANSI%20SQL) -> [NULL predicate](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20fa62ef75191014a797dae048566adc.html)
[W3C](https://www.w3schools.com/sql/sql_null_values.asp)

danjoa:
I'm afraid we're still confusing ANSI SQL's _unary_ `IS [NOT] NULL` operator vs the _binary_ `IS [NOT] DISTINCT FROM` operator. → Your latest changes also a wrong in that case. 

Refer to this (run with postgres):
```sql 
SELECT 1 != null;      --> null
SELECT 1 = null;       --> null
SELECT 1 is not null;  --> true
SELECT 1 is null;      --> false
CREATE TABLE t ( x integer );
INSERT into t values (1);
INSERT into t values (null);
SELECT * from t where x != null;
SELECT * from t where x = null;      
SELECT * from t where x is not null;
SELECT * from t where x is null;
SELECT * from t where x is distinct from null;
SELECT * from t where x is not distinct from null;
```

BobdenOs:
I am confident that I am confused.

Maybe some quick clarifications:
1. Should default `cqn2sql` only generate valid ANSI SQL ?
2. Would `IS [NOT] DISTINCT FROM` be considered valid ANSI SQL ?
3. If `IS [NOT] DISTINCT FROM` is not valid ANSI SQL, what should the default be ?
4. is it correct that `x is not 5`  is invalid ANSI SQL ?

danjoa:
P.S. Note the the case mentioned in the issue #156 wouldn't benefit from all the things done in here, unless `customerAccountNumber` is declared with `not null`.

BobdenOs:
The difference between `parent.ID` vs `parent_ID` sounds like the other issues with generated foreign keys. Where cqn4sql is build with a focus on structured results which makes them equal.

As long as that foreign key is identified as a key it won't be possible to merge this PR.

danjoa (approved):
As discussed: [Parked the .key and .notNull-based optimizations](https://github.com/cap-js/cds-dbs/pull/157/commits/ebf60ee3a539a5f8bb642548fcecc09484e6e215) due to cqn4sql's `.elements` glitch

patricebender:
@BobdenOs @danjoa please have a look at #166 which identifies foreign keys in a flat model and attaches them rather than the target element

PR #160, opened 2023-08-08
feat: edge database service

BobdenOs:
## Description

This PR implements a PoC for how the pluggable database services can be wrapped together into a multi staged database service.

This implementation uses an `SQLiteService` as an in memory `edge` database and a `PostgresService` as `root` database. This is a very basic scenario, but also about how complex most edge database implementations go. With more database services implemented it would be possible to tier the data in more complex configurations. Going from `edge`, `hot`, `warm` to `cold`. Or potentially cross region using the same database service type for different levels, but ordered based upon latency from the service to the database instances.

## Examples

When writing to the database service a transaction is started in all databases. Only if all database layers are successful in the write operation will it be committed.

When reading from the database service this is done firstly from the edge database. If it is determined that the information is insufficient in the edge database the query will be routed further to the next database. This is repeated until all databases have been queried. It is very important to prevent cascading the requests down as this removes the benefit of having the `edge` database.

## Improvements

The more the requests to the instance in which the `edge` database resides are related to each other the less likely it is that the edge and root database de-synchronize. For example it would be possible to route all requests for a single tenant to the same service instance. This will mean that all changes made to the tenant database are reflected in the `edge` database. Making it impossible to have `cache` misses.

It would greatly improve the capabilities of the edge service when it would be possible to model data life cycles. This would provide the metadata required to fill the `edge` database with the most relevant information. This is especially important when it is not possible to load full tables into the service memory. When millions of records are accumulating into the `root` database. It would require a certain filter condition for the `edge` database to load the most relevant subset into the service. This would reduce service memory consumption and `cache` misses. Additionally this would allow for `warm` and `cold` database services to be introduced. Where the edge service would send queries to the lower level database instances and move data from `hot` to `cold`. Allowing for CAP services to handle potentially petabytes of datasets rather then the terabytes limits current cloud instances have. 

Additionally when the life cycle is known it is possible to union multiple database layers together. Only requesting data that is missing from the `edge` database. When the lifecycle column is a `timestamp` and the filter condition is within the `edge` database dataset. It will never need to query the following database layer. While when the filter condition falls outside of the `edge` database dataset it can skip the `edge` database and directly query the relevant database layer which would contain relevant data.

### DIFF
--- a/db-edge/CHANGELOG.md
+++ b/db-edge/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Change Log
+
+- All notable changes to this project are documented in this file.
+- The format is based on [Keep a Changelog](http://keepachangelog.com/).
+- This project adheres to [Semantic Versioning](http://semver.org/).
+
+## Version 1.2.0 - tbd
+
+- Initial release
--- a/db-edge/LICENSE
+++ b/db-edge/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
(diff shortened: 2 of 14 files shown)


danjoa:
Cool initiative @BobdenOs 
We should discuss that also with our Business Network contacts → they frequently have to deal with these kinds of data federation needs

PR #162, opened 2023-08-10
feat: Reduce distinct from usage - part 2

### LINKED ISSUE #156: PostgreSQL Performance Improvement Request - Usage of "is not distinct from"

danjoa:
Follow up to #157 
Fixes #156 

This PR cannot be merged yet due to a glitch in cqn4sql:

- https://github.com/cap-js/cds-dbs/pull/157#issuecomment-1671501230
- https://github.com/cap-js/cds-dbs/pull/157#issuecomment-1671533888

@patricebender


danjoa:
Can be merged now as well

danjoa:
I fixed the same in https://github.tools.sap/cap/cds/pull/3299

BobdenOs (approved):
Looks good. Just the test expectation should be updated.

PR #163, opened 2023-08-10
add requirements to readme

sjvans:


danjoa (approved):

danjoa:
More important would be to have this in the sub packages as their readmes are published to npm. This one not.

PR #166, opened 2023-08-11
cqn4sql: attach adjacent fk adjacent to assoc as element of column

### LINKED ISSUE #165: cqn4sql: Assign foreign key as `element` of column for flat model
Consider this query:

```
SELECT from bookshop.Books {
        ID,
        author, // transformed to `author_ID` -> attach foreign key `Books:author_ID` rather than `Author:ID`
      }
```

If the underlying model is compiled for nodejs, i.e. it is flat, the helper property `element` which is attached to every column, during the `cqn4sql` transformation must point to flat foreign keys which are always part of the `elements` of the containing entity. In this case, `Books` has the `elements.author_ID` in contrast to a structured model, where there only would be `elements.author`

reported in #157

patricebender:
for a flat model, we must attach the foreign key which is adjacent to it's association as the columns `element` property. If such a flat fk is found next to the association in the containing entity, it is used as the `element` rather than the key in the target of the association.

fixes #165

### DIFF
--- a/db-service/test/cqn4sql/column.element.test.js
+++ b/db-service/test/cqn4sql/column.element.test.js
@@ -107,7 +107,7 @@ describe('assign element onto columns with flat model', () => {
     `
     const { Authors } = model.entities
     expect(JSON.parse(JSON.stringify(query))).to.deep.eql(expected)
-    // foreign key is part of flat model
+
     expect(query.SELECT.columns[1].SELECT.columns[0]).to.have.property('element').that.eqls(Authors.elements.ID)
   })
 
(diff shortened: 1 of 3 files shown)


BobdenOs (approved):
In the tests I didn't see the opposite test case. That the key in an expand is still the key column. I expect it to work fine. Just an observation.

db-service/test/cqn4sql/column.element.test.js
```diff
@@ -63,3 +63,97 @@ describe('assign element onto columns', () => {
...
+        (SELECT from bookshop.Authors as author {
+          author.ID
+        } where Books.author_ID = author.ID) as author
+    }
+    `
+    const { Authors } = model.entities
+    expect(JSON.parse(JSON.stringify(query))).to.deep.eql(expected)
+    // foreign key is part of flat model
+    expect(query.SELECT.columns[1].SELECT.columns[0]).to.have.property('element').that.eqls(Authors.elements.ID)
+  })
```
patricebender, line 112: done

PR #167, opened 2023-08-11
chore(infer): add some type definitions

patricebender:
how can I make the defined types avaiable to all files in the same directory without having to import all typedefinitions all over again?


BobdenOs:
I had the same question. I didn't look into it yet, but .d.ts files are supposed to be global definitions. So the premise is that you define a namespace and then export all your types in there. Which would then look like this:
```js
/*
 * @type {namespace.type}
 */
const value = 'data'
```
All I know is that the detached `CQN` type was not resolved by `tsc`. So some experimenting has to be done to find what works. Which should be pretty straight forward just run the build command and ser whether it found the type or whether it define it as `any`.

danjoa (approved):

Issue #159, opened 2023-08-07
@cap-js/sqlite: nested properties in where clause of delete crash


sebastianesch:
Hi CAP Team!

When deleting using nested properties:

```js
 const result = await DELETE.from(GroupMembers).where({'user.externalUserId': user.externalUserId })
```

I get the following error:
```sh
TypeError: Cannot read properties of undefined (reading 'map')
    at _entityTransitionsForTarget (/.../node_modules/@sap/cds/libx/_runtime/common/utils/resolveView.js:593:19)
```

In resolveView the  `from` input for function `_entityTransitionsForTarget` is:

```js
{
  join: "left",
  args: [
    {
      ref: [
        "user.db.GroupMembers",
      ],
      as: "GroupMembers",
    },
    {
      ref: [
        "user.db.Users",
      ],
      as: "user",
    },
  ],
  on: [
    {
      ref: [
        "user",
        "ID",
      ],
    },
    "=",
    {
      ref: [
        "GroupMembers",
        "user_ID",
      ],
    },
  ],
}
```
but in line 593 an attribute `ref` is expected, that is not there in the DELETE case.

```js
return from.ref.map((f, i) => { ... } // line 593
```

`from.ref` is undefined in this case.

| Your Project Name | N/A |
|:------------------|---------------------------------------|
| OData version     | v4                              |
| Node.js version   | v18.16.1                               |
| @sap/cds          | 7.0.2                                 |
| @sap/cds-compiler | 3.9.2                                |
| @sap/cds-dk       | 7.1.0                                 |
| @cap-js/sqlite | 1.0.1 |

Kind regards,
Sebastian

patricebender:
I just tried it out and this still doesnt work. The db-service translates such a path expression to a join. However, if there are localized fields present, the whole `where` clause is put in a `from` filter which leads to a scoped query: `DELETE Books[author.name = 'Emily Bronte']:author` which doesnt work because non-fk traversals in infix filters are not allowed. If the localized fields are removed, we get the reported dump in `resolveView`:

```
> del =  DELETE.from(Books).where("author.name = 'Emily Brontë'");
Query {
  DELETE: {
    from: 'sap.capire.bookshop.Books',
    where: [ { ref: [ 'author', 'name' ] }, '=', { val: 'Emily Brontë' } ]
  }
}
> await del
Uncaught TypeError: Cannot read properties of undefined (reading 'map')
[...]

PR #168, opened 2023-08-11
fix: prevent resolve view from being called when using path expressions

BobdenOs:
details: https://github.com/cap-js/cds-dbs/issues/159

### DIFF
--- a/db-service/lib/deep-queries.js
+++ b/db-service/lib/deep-queries.js
@@ -44,6 +44,11 @@ async function onDeep(req, next) {
 const hasDeep = (query, target) => {
   if (handledDeep in query) return
   if (query.DELETE) {
+    // REVISIT: Resolve view does not support from.join
+    // But deep delete is capable of handling this query
+    // Therefor this has to return true to prevent resolve view
+    // from being called and throwing an error
+    if (!query.DELETE.from.ref) return true
     for (let c in target?.compositions) return true
     return false
   }


patricebender (approved):

danjoa (approved):

BobdenOs:
This is happening in most cases as expected (like: https://github.com/cap-js/cds-dbs/pull/170#issue-1849524125). I am not sure why it is not happening in the reported query.

patricebender:
> @BobdenOs @patricebender Shouldn't cqn4sql differentiate between READ and UPDATE/DELETE? For UPDATE/DELETE, a path expression should be rendered to a `where exists` clause, shouldn't it?

No for an `UPDATE`  there is an error emitted if path expressions are used. See https://github.com/cap-js/cds-dbs/pull/97 for details. `DELETE`s have a from clause which might also have join nodes hence path expressions should work.

johannes-vogel:
> > @BobdenOs @patricebender Shouldn't cqn4sql differentiate between READ and UPDATE/DELETE? For UPDATE/DELETE, a path expression should be rendered to a `where exists` clause, shouldn't it?
> 
> No for an `UPDATE` there is an error emitted if path expressions are used. See #97 for details. `DELETE`s have a from clause which might also have join nodes hence path expressions should work.

Now I understand the current limitation. Long term, path expressions should also be usable in WHERE clauses of both, UPDATE and DELETE, shouldn't it?

PR #169, opened 2023-08-11
Replacing assoc name with alias in where when different

schiwekM:
Hi,

when having a data model like 

```cds
entity Foo : cuid {
   owner : Composition of many Owner on owner.foo = $self;
   activeOwners : Association to many ActiveOwner on activeOwners.foo = $self;
   owner2 : Composition of many Owner2 on owner2.foo = $self;
   specialOwners : Association to many SpecialOwner2 on specialOwners.foo = $self;

   boos: Association to many Boo on boos.foo = $self;
}

entity ActiveOwner as projection on Owner where validFrom <= $now and validTo >= $now;
entity SpecialOwner2 as projection on Owner2 where validFrom <= $now and validTo >= $now and isSpecial = true; 

entity Owner2 : cuid {
   foo : Association to one Foo;
   owner2 : Association to one Employees;
   isSpecial : Bool default false;
   validFrom: Date;
   validTo: Date;
}

entity Owner : cuid {
   foo : Association to one Foo;
   owner: Association to one Employees;
   validFrom: Date;
   validTo: Date;
}

entity Employees {
   key userID: String;
}

entity Boo : cuid {
   foo_ID: UUID;
   foo : Association to one Foo on foo.ID = foo_ID;
}
```

and `Boo` has the `@restrict` where clause
`'exists foo.specialOwners[owner2_userID = $user.id] or exists foo.activeOwners[owner_userID = $user.id]'`

it currently crashes when calling `GET /service/Foo(ID=<some ID>)/boos` because the second exists clause will get an alias `foo2` however does not have the alias in the where (the internal where however does).

Test in PR is currently wrong as it throws another error when doing a SELECT the way it is in the test.


db-service/test/cqn4sql/where-exists.test.js
```diff
@@ -1240,6 +1240,14 @@ describe('Path expressions in from combined with `exists` predicate', () => {
           AND EXISTS ( SELECT 1 from bookshop.Genres as parent where parent.ID = genre.parent_ID )
       `)
   })
+
+  it('MUST ... two EXISTS both on same path in where', () => {
+    let query = cqn4sql(CQL`SELECT from bookshop.Books { ID } where exists genre.children[code = 'ABC'] or genre.children[code = 'DEF']`, model)
```
patricebender, line 1245: If you want to express to parallel exists you gotta write it like that
```suggestion
    let query = cqn4sql(CQL`SELECT from bookshop.Books { ID } where exists genre.children[code = 'ABC'] or exists genre.children[code = 'DEF']`, model)
```

patricebender: and then the correct, resulting cqn would be:

```
SELECT from bookshop.Books as Books { Books.ID }
          WHERE EXISTS (
            SELECT 1 from bookshop.Genres as genre where genre.ID = Books.genre_ID
              and EXISTS ( SELECT 1 from bookshop.Genres as children where children.parent_ID = genre.ID and children.code = 'ABC' )
          )
          or  EXISTS (
            SELECT 1 from bookshop.Genres as genre2 where genre2.ID = Books.genre_ID 
            and EXISTS ( SELECT 1 from bookshop.Genres as children2 where children2.parent_ID = genre2.ID and children2.code = 'DEF' )
          )
        `)

```

patricebender:
Thanks for contributing Marten. I can confirm the issue and will look into it :)

schiwekM:
@patricebender thanks for having a look. The proposed fix at least resolves the issue in the productive application - however the test case does not really cover it, because the deep comparison also compares against the wrong cqn. 
Is there a way to execute the cqn against the db also in the cqn4sql tests? Because the current way is not ideal, as we only test if the generation leads to same results with different syntax not that the result is correct.

patricebender:
> @patricebender thanks for having a look. The proposed fix at least resolves the issue in the productive application - however the test case does not really cover it, because the deep comparison also compares against the wrong cqn. Is there a way to execute the cqn against the db also in the cqn4sql tests? Because the current way is not ideal, as we only test if the generation leads to same results with different syntax not that the result is correct.

I see. We have end2end tests as well. Those are found in the `cqn2sql`, `sqlite` and `postgres` test suite. However, by looking at the transformations here in the cqn4sql tests, we can at least make sure that aliasing and semantic transformation happened properly. So lets focus on having the same expectation first. I have added your test exampe in [this branch](https://github.com/cap-js/cds-dbs/compare/broken-exists?expand=1) see "real life example test". My expectation matches with the result from cqn4sql. Can you confirm that this cqn looks good to you? If that is the case, we can progress and add end 2 end tests and see if the bug is further down the line.

schiwekM:
> > @patricebender thanks for having a look. The proposed fix at least resolves the issue in the productive application - however the test case does not really cover it, because the deep comparison also compares against the wrong cqn. Is there a way to execute the cqn against the db also in the cqn4sql tests? Because the current way is not ideal, as we only test if the generation leads to same results with different syntax not that the result is correct.
> 
> I see. We have end2end tests as well. Those are found in the `cqn2sql`, `sqlite` and `postgres` test suite. However, by looking at the transformations here in the cqn4sql tests, we can at least make sure that aliasing and semantic transformation happened properly. So lets focus on having the same expectation first. I have added your test exampe in [this branch](https://github.com/cap-js/cds-dbs/compare/broken-exists?expand=1) see "real life example test". My expectation matches with the result from cqn4sql. Can you confirm that this cqn looks good to you? If that is the case, we can progress and add end 2 end tests and see if the bug is further down the line.

Looks good to me 👍

patricebender:
@schiwekM I have added an end 2 end test in [this commit](https://github.com/cap-js/cds-dbs/commit/b1fc7e7c9bbccf8c50c1fba1a2f58bf943871128), could you have a look? If you wish to execute this test by yourself on sqlite, do it like this: `CDS_ENV=better-sqlite mocha test/scenarios/cap_issue/read.test.js`

schiwekM:
> @schiwekM I have added an end 2 end test in [this commit](https://github.com/cap-js/cds-dbs/commit/b1fc7e7c9bbccf8c50c1fba1a2f58bf943871128), could you have a look as this does not yet cause a crash? If you wish to execute this test by yourself on sqlite, do it like this: `CDS_ENV=better-sqlite mocha test/scenarios/cap_issue/read.test.js`

Hi Patrice,
in the setup it is important that `'exists foo.specialOwners[owner2_userID = $user.id] or exists foo.activeOwners[owner_userID = $user.id]'` is injected into the query. Either with @restrict or a handler, else it won't crash

patricebender:
> > @schiwekM I have added an end 2 end test in [this commit](https://github.com/cap-js/cds-dbs/commit/b1fc7e7c9bbccf8c50c1fba1a2f58bf943871128), could you have a look as this does not yet cause a crash? If you wish to execute this test by yourself on sqlite, do it like this: `CDS_ENV=better-sqlite mocha test/scenarios/cap_issue/read.test.js`
> 
> Hi Patrice, in the setup it is important that `'exists foo.specialOwners[owner2_userID = $user.id] or exists foo.activeOwners[owner_userID = $user.id]'` is injected into the query. Either with @restrict or a handler, else it won't crash

thanks for the hint, I added the restrict clause to the model, fired the query again -> no crash.

schiwekM:
> > > @schiwekM I have added an end 2 end test in [this commit](https://github.com/cap-js/cds-dbs/commit/b1fc7e7c9bbccf8c50c1fba1a2f58bf943871128), could you have a look as this does not yet cause a crash? If you wish to execute this test by yourself on sqlite, do it like this: `CDS_ENV=better-sqlite mocha test/scenarios/cap_issue/read.test.js`
> > 
> > 
> > Hi Patrice, in the setup it is important that `'exists foo.specialOwners[owner2_userID = $user.id] or exists foo.activeOwners[owner_userID = $user.id]'` is injected into the query. Either with @restrict or a handler, else it won't crash
> 
> thanks for the hint, I added the restrict clause to the model, fired the query (observing that the exists clause is now part of the query) again -> no crash.

I assume the test case does not have authorisation enabled https://cap.cloud.sap/docs/node.js/authentication

patricebender:
> n you confirm that this cqn looks good to you? If that is the case, we can progress and add end 2 end tests and see if the bug is further down the line.

Hm, okay. I'm not really fluent when it comes to those auth concepts. However, If I do not pass the `admin` user to my `GET` request, I get an `401` so I assumed auth is already enabled.

db-service/lib/cqn4sql.js
```diff
@@ -1605,6 +1605,11 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
       res = transformedOn
     } else if (on) {
       res = calculateOnCondition(on)
+      if (assocRefLink.alias !== assocRefLink.definition.name)
+        res.forEach(r => {
+          if (r.ref && r.ref[0] === assocRefLink.definition.name) 
+            r.ref[0] = assocRefLink.alias
+        })
```
danjoa, line 1612:
```suggestion
        for (let r of res) {
          if (r.ref?.[0] === assocRefLink.definition.name) 
            r.ref[0] = assocRefLink.alias
        }
```

danjoa (approved):

patricebender:
> it currently crashes when calling GET /service/Foo(ID=<some ID>)/boos because the second exists clause will get an alias foo2 however does not have the alias in the where (the internal where however does).

Just to make sure: ist the error emmited from sqlite, or is it a real crash in our stack? If so, could you provide the error log?

schiwekM:
> > it currently crashes when calling GET /service/Foo(ID=)/boos because the second exists clause will get an alias foo2 however does not have the alias in the where (the internal where however does).
> 
> Just to make sure: ist the error emmited from sqlite, or is it a real crash in our stack? If so, could you provide the error log?

Only sqlite complains

patricebender:
Hi @schiwekM,

with the step-by-step you provided in the cap/issue, I was able to reproduce the malformed sql. I was also able to come up with a [fix](https://github.com/cap-js/cds-dbs/compare/main...broken-exists) would you please confirm that this works on your side as well? I also have a fix in the backlog for the other dump in the `calculatedOnCondition` function which you have found. However, lets first check if this works for you.

schiwekM:
> Hi @schiwekM,
> 
> with the step-by-step you provided in the cap/issue, I was able to reproduce the malformed sql. I was also able to come up with a [fix](https://github.com/cap-js/cds-dbs/compare/main...broken-exists) would you please confirm that this works on your side as well? I also have a fix in the backlog for the other dump in the `calculatedOnCondition` function which you have found. However, lets first check if this works for you.

Tried it out and does the trick - thanks

patricebender:
Closing in favor of #173 thanks!

PR #172, opened 2023-08-15
test: Split test script into each package again

BobdenOs:
Change the approach of doing multi workspace testing

```sh
npm t -ws # runs all test in the workspaces in order of the package.json workspaces array
npm t -w postgres -w hana # will run the test command for postgres and hana
```

Which allows each package to do their `start` script before running their tests. Also it allows for all the packages to have their own local `jest.config.js` files. Which allows `sqlite` to have a smaller `testTimeout` then `postgres` and `hana`.

### DIFF
--- a/db-service/package.json
+++ b/db-service/package.json
@@ -25,6 +25,9 @@
     "node": ">=16",
     "npm": ">=8"
   },
+  "scripts": {
+    "test": "jest --silent"
+  },
   "peerDependencies": {
     "@sap/cds": ">=7.1.1"
   },
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "author": "SAP SE (https://www.sap.com)",
   "workspaces": [
     "db-service",
-    "postgres",
     "sqlite",
+    "postgres",
     "hana"
   ],
   "devDependencies": {
@@ -22,7 +22,7 @@
     "express": "^4"
   },
   "scripts": {
-    "test": "npm start -w postgres && npm run setup -w hana && CDS_JEST_MEM_FIX=1 npx jest --silent --colors",
+    "test": "CDS_JEST_MEM_FIX=1 npm t -w db-service -w sqlite",
     "lint": "npx eslint ."
   },
   "license": "SEE LICENSE"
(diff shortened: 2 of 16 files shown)


db-service/package.json
```diff
@@ -25,6 +25,9 @@
     "node": ">=16",
     "npm": ">=8"
   },
+  "scripts": {
+    "test": "jest --silent"
```
danjoa, line 29:
```suggestion
    "test": "jest --silent --colors"
```

package.json
```diff
@@ -22,7 +22,7 @@
     "express": "^4"
   },
   "scripts": {
-    "test": "npm start -w postgres && npm run setup -w hana && CDS_JEST_MEM_FIX=1 npx jest --silent --colors",
+    "test": "CDS_JEST_MEM_FIX=1 npm t -w db-service -w sqlite",
```
danjoa, line 25:
```suggestion
    "test": "CDS_JEST_MEM_FIX=1 npm t -w db-service -w sqlite --colors",
```

postgres/package.json
```diff
@@ -27,8 +27,8 @@
     "npm": ">=8"
   },
   "scripts": {
-    "start": "docker-compose -f pg-stack.yml up -d",
-    "test": "npm start && jest --silent"
+    "test": "npm start && jest --silent",
```
danjoa, line 30:
```suggestion
    "test": "npm start && jest --silent --colors",
```

sqlite/package.json
```diff
@@ -26,6 +26,9 @@
     "node": ">=16",
     "npm": ">=8"
   },
+  "scripts": {
+    "test": "jest --silent"
```
danjoa, line 30:
```suggestion
    "test": "jest --silent --colors"
```

danjoa (approved):

db-service/package.json
BobdenOs: I don't like the pattern to always define `--colors` in every `jest` call. My understanding is that this is only to have color in the Github action. Therefor I have added the environment variable `FORCE_COLOR` which will enable it explicitly only in the Github action. Assuming that it will behave correctly in all other scenarios.

PR #173, opened 2023-08-16
cqn4sql: make sure to always use calculated alias in custom on conditions

patricebender:
Inspired by #169

---

the issue here was that we had a where condition like

`where exists foo[id=1] or exists foo[id=2]` with foo being an association `foo : Association to one Foo on foo.ID = foo_ID;`.

While building up the where exists subqueries, we calculate unique table aliases for `foo`, which results in a table alias `foo2` for the second condition of the initial where clause. Now, if we incorporate the on-condition into the where clause of the second where exists subquery, we must replace the table alias `foo` from the on-condition with `foo2`.

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -1689,7 +1689,7 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
               lhs.$refLinks[0].definition ===
               getParentEntity(assocRefLink.definition).elements[assocRefLink.definition.name]
             )
-              result[i].ref = [result[i].ref[0], lhs.ref.slice(1).join('_')]
+              result[i].ref = [assocRefLink.alias, lhs.ref.slice(1).join('_')]
             // naive assumption: if the path starts with an association which is not the association from
             // which the on-condition originates, it must be a foreign key and hence resolvable in the source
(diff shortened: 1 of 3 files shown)


stewsk:
The test case provided here is (if I'm not mistaken) the same as in https://github.com/cap-js/cds-dbs/pull/169. But that test didn't discover the problem described in https://github.com/cap-js/cds-dbs/pull/173#issue-1852975234 - or did it?

patricebender:
> Reference in new issue

nope it didn't, im still working on that.

sjvans:
> > Reference in new issue
> 
> nope it didn't, im still working on that.

for this pr?

patricebender:
> > > Reference in new issue
> 
> > 
> 
> > nope it didn't, im still working on that.
> 
> 
> 
> for this pr?

Yes, I will add it as part of this pr hopefully today

patricebender:
@stewsk @sjvans please have a look at #173 which showcases the error which was reported in the cap issue.

stewsk:
Is it possible to have a dedicated cqn4sql test for the situation described in https://github.com/cap-js/cds-dbs/pull/174 ?

patricebender:
> Is it possible to have a dedicated cqn4sql test for the situation described in #174 ?

yes you are right, this way it is much clearer what the issue is. I replaced the end-2-end test with a proper cqn4sql test illustrating the issue. :)

PR #174, opened 2023-08-17
cqn4sql: add test showcasing issue from #169

patricebender:
reproduced `AxiosError: 500 - no such column: foo.ID in:`

---

see inline comments and #169 for details


patricebender:
@David-Kunz thanks for helping yesterday, this scenario successfully reproduces the error and is to-be-merged into the fix branch #173

patricebender:
All tests in `test/scenario` are not executed anymore in the pipeline. @danjoa @BobdenOs please have a look, maybe it is related to the error shown in the action:

```log
Using prepared HXE image
Creating hxe_hana_1 ... 
Creating hxe_hana_1 ... error

ERROR: for hxe_hana_1  Cannot create container for service hana: Conflict. The container name "/hxe_hana_1" is already in use by container "56155d69aa19006652ec09a1a1fdc79b75fe311cfd557f70613a96168234c6ba". You have to remove (or rename) that container to be able to reuse that name.

ERROR: for hana  Cannot create container for service hana: Conflict. The container name "/hxe_hana_1" is already in use by container "56155d69aa19006652ec09a1a1fdc79b75fe311cfd557f70613a96168234c6ba". You have to remove (or rename) that container to be able to reuse that name.
Encountered errors while bringing up the project.
Error response from daemon: No such container: hxe_hana_1
```

Locally, executing only this test works and yields the desired error.

patricebender:
```log
  650 passing (2s)
  21 pending
  1 failing

  1) cap issue
       MUST ... two EXISTS both on same path in where with real life example:

      AssertionError: expected { SELECT: { from: { …(3) }, …(2) } } to deeply equal { …(2) }
      + expected - actual

                     }
                     "where": [
                       {
                         "ref": [
      -                    "foo"
      +                    "foo2"
                           "ID"
                         ]
                       }
                       "="
```

stewsk (approved):

PR #176, opened 2023-08-17
cqn4sql: skip assocs from wildcard expansion instead of fks in flat mode

patricebender:
this reverts the cqn4sql part of the hotfix in #154

---

The problem was, that in a flat model, the foreign keys were not part of the result anymore, in case that the respective association was overwritten by the smart wildcard mechanism. E.g. by an expand.

The reason was, that in case of a flat model, we skipped the foreign keys and still flattened the association. This caused the bug where overwritting an association caused the lack of the foreign key in the result.

The fix here is simple, skip the association from the wildcard expansion instead. The compiler already, correctly generates the foreign key elements for us. This way, we have the correct "smart wildcard" mechanisms well known from the compiler while persisting the foreign keys in flat mode.

@sjvans @BobdenOs @stewsk please have look.


db-service/test/cqn4sql/wildcards.test.js
```diff
@@ -168,6 +168,31 @@ describe('wildcard expansion and exclude clause', () => {
...
+          Books.currency_code,
+          Books.dedication_addressee_ID,
+          Books.dedication_text,
+          Books.dedication_sub_foo,
+          Books.dedication_dedication,
+          Books.coAuthor_ID_unmanaged,
+        }
+      `)
+  })
+  it('expand after wildcard combines assoc from wildcard expansion', () => {
```
stewsk, line 184:
```suggestion
  it('expand after wildcard combines assoc from wildcard expansion (flat mode)', () => {
```

stewsk (approved):

db-service/lib/cqn4sql.js
```diff
@@ -911,7 +911,7 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
       const { index, tableAlias } = inferred.$combinedElements[k][0]
       const element = tableAlias.elements[k]
       // ignore FK for odata csn / ignore blobs from wildcard expansion
-      if (isODataFlatForeignKey(element) || (element['@Core.MediaType'] && !element['@Core.IsURL'])) return
+      if (isManagedAssocInFlatMode(element) || (element['@Core.MediaType'] && !element['@Core.IsURL'])) return
```
patricebender, line 914: rather filter out duplicates than to rely on some meta information w.r.t. the model

patricebender:
I will merge this and apply the enhancement w.r.t. the de-duplication later. This requires a little more thinking and this change is definitely an improvement compared to the previous state.

PR #178, opened 2023-08-18
pg: add more reserved words

patricebender:
Postgres keywords, used for smart quoting.
Taken from https://www.postgresql.org/docs/current/sql-keywords-appendix.html


BobdenOs:
I tried out some of the claimed to be reserved words and some of them worked fine without quoting. Probably make sure that this list is aligned with the reserved words of the cds compiler.

patricebender:
> I tried out some of the claimed to be reserved words and some of them worked fine without quoting. Probably make sure that this list is aligned with the reserved words of the cds compiler.

I forgot to mention, this is exactly the list from the cds-compiler

BobdenOs (approved):

PR #182, opened 2023-08-18
db-service CHANGELOG.md

patricebender:
please don't merge, yet.

Let us first gather all changes which will be part of the next release

Issue #180, opened 2023-08-18
@cap-js/postgres - multiple expands not working


nitiiz:
### Description of erroneous behaviour 
Hi,

I noticed that when entity has multiple associations or compositions and trying to call them in single OData call is failing as "Internal Server Error" and in the service-srv logs I can see that the following error is occuring when calling the odata entity with multiple expands.
Entity has more than 40 fields (mixed Boolean and String) and 4 compositions to another entities

Entity?$expand=CustomerContacts,VendorContacts => fails
Entity?$expand=CustomerContacts => works
Entity?$expand=VendorContacts => works
does not matter which association is used since only one works

![image](https://github.com/cap-js/cds-dbs/assets/14157177/2ed58e5d-3ccd-4fed-935a-e94c55ca60e6)
![image](https://github.com/cap-js/cds-dbs/assets/14157177/f5b8be37-72f5-40d0-bdf3-2ab3dcaac5c1)
![image](https://github.com/cap-js/cds-dbs/assets/14157177/7145fdcb-f883-4789-b5d3-87c83a19a2fa)

"cds] - error: cannot pass more than 100 arguments to a function"

### Detailed steps to reproduce

> 1. create a single odata entity with multiple compositions as in picture
> 2. call them using Entity?$expand=FirstExpand,SecondExpand

### Details about your project

|:------------------|---------------------------------------|
| OData version     | v4 / v2                               |
| Node.js version   | v18.14.2                               |
| @sap/cds          | 7.1.1                                 |
| @sap/cds-dk       | 7.1.1                                 |
| @cap-js/postgres | 1.1.0 |

patricebender:
related issue #139, @BobdenOs could you have a look?

nitiiz:
when this fix will be released?

johannes-vogel:
> when this fix will be released?

at latest beginning of september. is that sufficient?

nitiiz:
> > when this fix will be released?
> 
> at latest beginning of september. is that sufficient?

I think it's okay we can work around somehow without fetching data because currently the database fetching is unusable because we have lot of fields and even when trying to call OData create we get the same error because it tries to return in the response all the fields and that limit 100 fields throws an error. OData get works if we limit the fields using $select but in odata create there's no that possibility.

For us as soon as possible the better we can continue working on our tasks.
Thank you!

nitiiz:
@patricebender not sure if its related to this issue but I noticed that field type array of String is not working when trying to set data to that field.

[...]

PR #184, opened 2023-08-21
fix: pass json as parameter instead of sql string

BobdenOs:
When updating a `many` column type. The `json` value is converted into a `SQL` string, but is send as a parameter.


danjoa (approved):

patricebender (commented):
is this for #180? If so could you add a changelog entry [here](https://github.com/cap-js/cds-dbs/pull/182)?
Also, would be nice if we had a test for this :)

PR #186, opened 2023-08-22
Using axios 1

danjoa:


BobdenOs:
Should this be updated in the `package-lock.json` ?

BobdenOs (approved):

PR #189, opened 2023-08-23
fix: Don't skip unknown columns in INSERT.columns

BobdenOs:
## Description

When inserting `rows` and `values` the `columns` order is directly linked to the data order. As it is required to properly infer the column type all unknown columns are filtered out. This results in the index of the column shifting and the data from removed columns being inserted into the remaining columns. 

This is mostly a side effect of in correctly defined `csv` data files.

```js
const q = {INSERT:{
  columns: ['id','title'], // should be ID
  rows: [
    [1,'title']
  ]
}}

await q 
// inserts 1 into the title column
// with the fix it will insert 'title' into the title column
```

## Related

reported by `cap/14304`


db-service/lib/cqn2sql.js
```diff
@@ -435,15 +435,16 @@ class CQN2SQLRenderer {
       throw cds.error`Cannot insert rows without columns or elements`
     }
     let columns = INSERT.columns || (elements && ObjectKeys(elements))
+    const columnIndex = columns.reduce((m, c, i) => ((m[c] = i), m), {})
     if (elements) {
       columns = columns.filter(c => c in elements && !elements[c].virtual && !elements[c].isAssociation)
     }
```
danjoa, line 441: Isn't the problem stemming from here? 
If we are called with a specified list of columns then this list of columns has to be valid or lead to an error, but not silent skipping → we should simply remove these lines imho instead 
```suggestion
```

danjoa (changes requested):

danjoa (approved):

PR #191, opened 2023-08-23
cqn4sql: for `$self` replacements, do not use table alias for flat name

patricebender:
easy fix, just check if `ref[0]` is an entity, if so, it must not be part of the flat name.

---

@stewsk please have a look


db-service/test/cqn4sql/table-alias.test.js
```diff
@@ -231,6 +216,31 @@ describe('table alias access', () => {
...
+      )
+      expect(query).to.deep.equal(CQL`
+      SELECT from bookshop.Books as ![FROM]
+      {
+        ![FROM].title as group,
+      }
+      where ![FROM].title = 'foo' 
+      group by ![FROM].title
+      having ![FROM].title = 'foo' 
+      order by ![FROM].title
```
stewsk, line 241: wasn't the plan to _not_ replace `group` by `![FROM].title` in the ORDER BY clause if the expression is a single ref?

stewsk (approved):

PR #195, opened 2023-08-25
[REVISIT:] Managed Data

danjoa:
This PR fixes these issues:

1. [Using `session_context()` with literals instead of binding parameters](https://github.com/cap-js/cds-dbs/pull/195/commits/3f063b9b328c22c92f1eac6002bc33087ff7dd89)

    Before, a query like that:
    ```js
    UPDATE(Books,201).with({stock:111})
    ```
    ... endet up in a SQL query like that:
    ```sql
    UPDATE 
      sap_capire_bookshop_Books AS Books 
    SET
      stock = 111,
      modifiedAt = coalesce(ISO(NULL), session_context(?)),
      modifiedBy = coalesce(NULL, session_context(?))
    WHERE ID=201;
    ```
    ... with these binding parameters:
    ```js
    [ '$now', '$user.id' ]
    ```

    With this fix, the session variables are sent as inline literals instead:
    ```sql
    UPDATE 
      sap_capire_bookshop_Books AS Books 
    SET
      stock = 111,
      modifiedAt = coalesce(ISO(NULL), session_context('$now')),
      modifiedBy = coalesce(NULL, session_context('$user.id'))
    WHERE ID=201;
    ```
    
2. [Apply Input Converters before `coalesce()` for managed data elements](https://github.com/cap-js/cds-dbs/pull/195/commits/0b19a341b5c8cd0b2b3a424d0e2123fa1720525b)

    Before, expressions for writing managed data elements looked like that in the SQL output:
    ```sql
    modifiedAt = ISO(coalesce(NULL, session_context('$now'))),
    ```
    With this fix, the Input Converters, i.e. `ISO()` function is applied before the `coalesce(<>,$now)`
    ```sql
    modifiedAt = coalesce(ISO(NULL), session_context('$now')),
    ```

3. [Allow pseudo vars everywhere](https://github.com/cap-js/cds-dbs/pull/195/commits/ded1fd44905b0188852c99eba7a2cd12b3beb323)

   With this fix queries like that are supported now, which wasn't the case before:
   ```sql
   SELECT.from(Books).where(`createdAt = $now`)
   ```

4. Simplify UPDATE of managed data (todo)

    Before, a query like that:
    ```js
    UPDATE(Books,201).with({stock:111})
    ```
    ... endet up in a SQL query like that:
    ```sql
    UPDATE 
      sap_capire_bookshop_Books AS Books 
    SET
      stock = 111,
      modifiedAt = coalesce(ISO(NULL), session_context('$now')),
      modifiedBy = coalesce(NULL, session_context('$user.id'))
    WHERE ID=201;
    ```
    
    With this fix this is optimised to that SQL:
    ```sql
    UPDATE 
      sap_capire_bookshop_Books AS Books 
    SET
      stock = 111,
      modifiedAt = ?,
      modifiedBy = ?
    WHERE ID=201;
    ```
    ... with binding parameters like that:
    ```js
    [ 
      req.data.modifiedAt || req.timestamp,
      req.data.modifiedBy || req.user.id,
    ]
    ```


danjoa:
@BobdenOs we need to follow up on this: [Fragile xtra condition for streaming in PostgreSQL](https://github.com/cap-js/cds-dbs/pull/195/commits/43234aede5f46056cc4bbca8732993103db5a847) 

→ we can't keep that like that.

If I remove this extra condition about leading '$', postgres test/streaming.test.js will fail

db-service/lib/cqn2sql.js
```diff
@@ -751,25 +754,19 @@ class CQN2SQLRenderer {
...
-        }
-        if (Buffer.isBuffer(val)) val = val.toString('base64')
-        else val = this.regex(val) || this.json(val)
+        if (val instanceof Readable) ; // go on with default below
+        else if (Buffer.isBuffer(val)) val = val.toString('base64')
+        else if (is_regexp(val)) val = val.source
+        else val = JSON.stringify(val)
+      case 'string': // eslint-disable-line no-fallthrough
     }
-    if (!this.values) return this.string(val)
```
BobdenOs, line 772: This is to render `CREATE` statements. Which are not allowed to have binding values.

db-service/lib/cqn2sql.js
```diff
@@ -795,25 +792,7 @@ class CQN2SQLRenderer {
   }
 
   /**
-   * Renders a Regular Expression into its string representation
```
BobdenOs, line 798: This is currently being called by the HANA service in the tests. Should simply be adjusting the `val` overwrite in the HANA Service.

db-service/lib/cql-functions.js
```diff
@@ -1,4 +1,8 @@
 const StandardFunctions = {
+
+
+  session_context: x => `session_context('${x.val}')`,
```
BobdenOs, line 4: This does the same as in [#188](https://github.com/cap-js/cds-dbs/pull/188/files#diff-d9ef4ae327f7e5954027eae8acf2866e34a6a365a11e60a3dc5ff535a210c33fR258). To prevent the possibility to do SQL injection it only applies this to known variables.

INSERT INTO.sql
```diff
@@ -0,0 +1,52 @@
...
+  coalesce(
+    ISO(value ->> '$."createdAt"'),
+    session_context('$now')
+  ),
+  coalesce(
+    value ->> '$."createdBy"',
+    session_context('$user.id')
+  ),
+  coalesce(
+    ISO(value ->> '$."modifiedAt"'),
```
BobdenOs, line 27: Having the converter around the `coalesce` simplifies the casting for typed databases. As the inpurt converter for Postgres is `cast` it will correct the `session_context` string type into the correct column type.

db-service/lib/cqn2sql.js
```diff
@@ -741,7 +739,12 @@ class CQN2SQLRenderer {
    * @returns {string} SQL
    */
   ref({ ref }) {
-    return ref.map(r => this.quote(r)).join('.')
+    switch (ref[0]) {
+      case '$now': return this.func({ func: 'session_context', args: [{ val: '$now' }]})
+      case '$user': return this.func({ func: 'session_context', args: [{ val: '$user.id' }]})
```
BobdenOs, line 744:
```suggestion
      case '$user':
```

BobdenOs:
I gave `default` and `generated` values a try with HANA, but it is very limited in which functions are allowed. Additionally it rejects any sub select as an invalid "user defined function". Which greatly reduces the chances of circumventing any limitations.

db-service/lib/cql-functions.js
danjoa: Ok. Yes I've seen that in postgres. Risk of SQL injection might be 0 as no user input ever goes into there, I think.

db-service/lib/cqn2sql.js
danjoa: Ah ok, will restore

db-service/lib/cqn2sql.js
danjoa: Let's revisit in that context with HANA currently copying the whole impl of `val()`

BobdenOs (approved):

Issue #192, opened 2023-08-23
`concat` function returns unexpected result.


dinurp:
### Description of erroneous behaviour 

`concat` function returns unexpected result.

For query q2 below year, month and correction are integers and have the values 2023, 8 and 0 respectively. 
```
> await SELECT.from(q2).columns`concat(year, (month * 2 - correction) ) as something`
[ { something: 40476 } ]
>  await SELECT.from(q2).columns`year || (month * 2 - correction) as something`
[ { something: '202316' } ]
>  await SELECT.from(q2).columns`year || month * 2 - correction as something`
[ { something: 40476 } ]
```
Expected result is `'202316'` but the returned result is `40476` when using `concat`.
It appears translation of `concat` is missing a set of parenthesis around its arguments. 

### Detailed steps to reproduce

1. `cds init capv7-eval`
2. `cd capv7-eval`
3. `npm add @cap-js/sqlite`
4. `npm i`
5. create file `db\Logs.cds` with the following contents:
  ```
  entity Logs {
      key ID: String(30);
       CreatedAt: Timestamp ;//@cds.on.insert : $now;
  }
  ```
6. `cds repl`
7. Following are the commands in `repl` session
```
> cds.test()
> await INSERT.into('Logs').entries( { ID:"1", CreatedAt:'2023-08-23T12:10:00.000Z'} );
> await SELECT.from('Logs')
> let q1 =  SELECT.from('Logs').columns`
	cast (substring( CreatedAt, 0, 4 ) as Integer) as year, 
	cast (substring( CreatedAt, 5, 2 ) as Integer) as month, 
	cast (substring( CreatedAt, 8, 2 ) as Integer) as day`;
> let q2 = SELECT.from(q1).columns`*, case when day < 16  then 1 else 0 end as correction`;

> await SELECT.from(q2).columns`concat(year, (month * 2 - correction) ) as something`
[ { something: 40476 } ]
>  await SELECT.from(q2).columns`year || (month * 2 - correction) as something`
[ { something: '202316' } ]
>  await SELECT.from(q2).columns`year || month * 2 - correction as something`
[ { something: 40476 } ]
```

### Details about your project

| capv7-eval | <Add your repository here> |
|:---------------------- | ----------- |
| Node.js                | v18.16.0    |
| @sap/cds               | 7.1.2       |
| @sap/cds-compiler      | 4.1.2       |
| @sap/cds-dk            | -- missing  |
[...]

PR #196, opened 2023-08-25
cqn2sql: dont forget to wrap `xpr`in parens

patricebender:
```log
sqlite> SELECT  2023||8 * 2 - 0 as something FROM Person;
40476
```
vs.
```log
sqlite> SELECT  2023|| (8 * 2 - 0) as something FROM Person;
202316
```

--> the xpr needs to be wrapped in parentheses:

```json
{
  "func": "concat",
  "args": [
    { "val": 2023 },
    {
      "xpr": [
        { "val": 8 },
        "*",
        { "val": 2 },
        "-",
        { "val": 0 }
      ]
    }
  ],
  "as": "something"
}
```

### DIFF
--- a/db-service/lib/cql-functions.js
+++ b/db-service/lib/cql-functions.js
@@ -40,7 +40,7 @@ const StandardFunctions = {
    * @returns {string}
    */
   concat: (...args) => args.map(a => `(${a})`).join('||'), // More brackets
-  // concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'), // Possible errors
+  // concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'), // Possible errors?
   /**
    * Generates SQL statement that produces a boolean value indicating whether the first string contains the second string
    * @param  {...string} args
--- a/db-service/lib/cql-functions.js
+++ b/db-service/lib/cql-functions.js
@@ -40,6 +40,7 @@ const StandardFunctions = {
    * @returns {string}
    */
   concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join(' || '),
+  
   /**
    * Generates SQL statement that produces a boolean value indicating whether the first string contains the second string
    * @param  {...string} args
--- a/db-service/lib/cql-functions.js
+++ b/db-service/lib/cql-functions.js
@@ -39,8 +39,8 @@ const StandardFunctions = {
    * @param  {...string} args
    * @returns {string}
    */
-  concat: (...args) => args.map(a => `(${a})`).join('||'), // More brackets
-  // concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'), // Possible errors?
+  // concat: (...args) => args.map(a => `(${a})`).join('||'), // More brackets
+  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'), // Possible errors?
   /**
    * Generates SQL statement that produces a boolean value indicating whether the first string contains the second string
    * @param  {...string} args
--- a/db-service/lib/cql-functions.js
+++ b/db-service/lib/cql-functions.js
@@ -40,7 +40,7 @@ const StandardFunctions = {
    * @returns {string}
    */
   // concat: (...args) => args.map(a => `(${a})`).join('||'), // More brackets
-  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'), // Possible errors?
+  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'),
   /**
    * Generates SQL statement that produces a boolean value indicating whether the first string contains the second string
    * @param  {...string} args
--- a/db-service/lib/cql-functions.js
+++ b/db-service/lib/cql-functions.js
@@ -256,7 +256,7 @@ const StandardFunctions = {
    * @param {string} x session variable name or SQL expression
    * @returns {string}
    */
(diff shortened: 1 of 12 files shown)


db-service/lib/cql-functions.js
```diff
@@ -51,7 +51,7 @@ const StandardFunctions = {
    * @param {string} x
    * @returns {string}
    */
-  count: x => `count(${x || '*'})`,
+  count: x => `count${(x?.xpr && x ) || (x && `(${x})`) || '(*)'}`,
```
danjoa, line 54:
```suggestion
  count: x => !x ? 'count(*)' : x.xpr ? `count${x}` : `count(${x})`,
```

danjoa: What is an example of such expressions in `count()`?

danjoa: Looks like this is a follow up fix for always wrapping `xpr` in `expr()`? → likely not good

db-service/lib/cqn2sql.js
```diff
@@ -664,7 +664,7 @@ class CQN2SQLRenderer {
     if ('param' in x) return wrap(this.param(x))
     if ('ref' in x) return wrap(this.ref(x))
     if ('val' in x) return wrap(this.val(x))
-    if ('xpr' in x) return wrap(this.xpr(x))
+    if ('xpr' in x) return wrap(`(${this.xpr(x)})`)
```
danjoa, line 667: Looks like far too much parenthesis

danjoa (changes requested):
I don't understand, why exactly we need that.
And if so, there has to be a better way

patricebender:
> I don't understand, why exactly we need that. And if so, there has to be a better way

https://github.com/cap-js/cds-dbs/issues/192

db-service/lib/cqn2sql.js
BobdenOs: I double checked and this is indeed not the correct place to do this. It is already being done where it should be done [here](https://github.com/cap-js/cds-dbs/blob/main/db-service/lib/cqn2sql.js#L684).

While constructing a reproduction query I found another issue.
```cql
-- CQL input
SELECT (1.0+2.0)*(1.0/2.0);
SELECT (1+2)*(1/2); -- is equal in cqn

-- SQL expectation
SELECT (1.0 + 2.0) * (1.0 / 2.0); -- 1.5

-- SQL output
SELECT (1 + 2) * (1 / 2); -- 0
```

danjoa:
The problem is the implementation of the concat() function for SQLite. 
The changes in PR are too broad for that.

BobdenOs:
@danjoa the implementation was adjusted to only change the `concat` implementation.
[here](https://github.com/cap-js/cds-dbs/pull/196/files#diff-d9ef4ae327f7e5954027eae8acf2866e34a6a365a11e60a3dc5ff535a210c33fR42)

danjoa:
> @danjoa the implementation was adjusted to only change the `concat` implementation.
> [here](https://github.com/cap-js/cds-dbs/pull/196/files#diff-d9ef4ae327f7e5954027eae8acf2866e34a6a365a11e60a3dc5ff535a210c33fR42)

Yes, this is the correct fix. 
```js
  concat: (...args) => args.map(x => x.xpr ? `(${x})` : x).join(' || '),
```

Should work as well I think. We don't need to wrap gals and refs into parentheses.

db-service/lib/cql-functions.js
```diff
@@ -39,7 +39,8 @@ const StandardFunctions = {
    * @param  {...string} args
    * @returns {string}
    */
-  concat: (...args) => args.join('||'),
+  // concat: (...args) => args.map(a => `(${a})`).join('||'), // More brackets
+  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'), // Possible errors?
```
patricebender, line 43:
```suggestion
  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join('||'),
```

db-service/test/cqn2sql/expression.test.js
```diff
@@ -27,6 +27,27 @@ describe('expressions', () => {
     const { sql, values } = cqn2sql(cqn)
     expect({ sql, values }).toMatchSnapshot()
   })
+  it.only('ref, String operator and value', () => {
```
patricebender, line 30:
```suggestion
  xtest('ref, String operator and value', () => {
```

db-service/lib/cql-functions.js
```diff
@@ -39,7 +39,8 @@ const StandardFunctions = {
    * @param  {...string} args
    * @returns {string}
    */
-  concat: (...args) => args.join('||'),
+  // concat: (...args) => args.map(a => `(${a})`).join('||'), // More brackets
+  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join(' || '),
```
danjoa, line 43:
```suggestion
  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join(' || '),
```

db-service/lib/cql-functions.js
```diff
@@ -39,7 +39,7 @@ const StandardFunctions = {
    * @param  {...string} args
    * @returns {string}
    */
-  concat: (...args) => args.join('||'),
+  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join(' || '),
```
danjoa, line 42:
```suggestion
  concat: (...args) => args.map(a => a.xpr ? `(${a})` : a).join(' || '),
  
```

danjoa (approved):

PR #199, opened 2023-08-28
fix: deep insert with backlinks as key in children

johannes-vogel:
I've added a test to reproduce the problem. Still needs to be fixed.

### DIFF
--- a/db-service/lib/fill-in-keys.js
+++ b/db-service/lib/fill-in-keys.js
@@ -23,6 +23,11 @@ const generateUUIDandPropagateKeys = (target, data, event) => {
     }
 
     if (elements[element].is2one || elements[element].is2many) {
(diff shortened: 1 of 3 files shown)


db-service/lib/fill-in-keys.js
```diff
@@ -23,6 +23,11 @@ const generateUUIDandPropagateKeys = (target, data, event) => {
     }
 
     if (elements[element].is2one || elements[element].is2many) {
+      // propagate own foreign keys to propagate further to sub data
+      propagateForeignKeys(element, data, elements[element]._foreignKeys, elements[element].isComposition, {
```
johannes-vogel, line 27: does propagateForeignKeys propagate one level or the complete business object?

Issue #201, opened 2023-08-30
@cap-js/postgres - array of String field type not able to update using OData


nitiiz:
### Description of erroneous behaviour 

I noticed that field type array of String is not working when trying to set data to that field.

Tried to set it to ["one", "two"]

Error when trying to OData Update those fields
<img width="731" alt="image" src="https://github.com/cap-js/cds-dbs/assets/14157177/ac5ea7fa-b03a-4bd8-ac26-371a45e5cb7f">

### Detailed steps to reproduce

> For example (→ replace by appropriate ones for your case):
> 1. Create OData entity and add one field which type is "array of String" as in picture and deploy
> 2. Call OData create to create an entry and set that field value to ["one"]
> 3. Call update to that created entry and set field value to ["one", "two"]
> 4. See error as in picture

### Details about your project

|:------------------|---------------------------------------|
| OData version | v4 / v2 |
| Node.js version | v18.14.2 |
| @sap/cds | 7.1.1 |
| @sap/cds-dk | 7.1.1 |
| @cap-js/postgres | 1.1.0 |

patricebender:
Hi @nitiiz,

again, thanks for reporting. I have been able to reproduce the error:

`git checkout v1.1.0`: 

```js
  it.only('Update array of', async () => {
    // create book
    const insert = INSERT.into('sap.capire.bookshop.Books')
      .columns(['ID', 'footers'])
      .values([150, ['one']])
    await cds.run(insert)

    const update = await PUT(
      '/admin/Books(150)', // UPSERT new footnotes
      {
        descr: 'UPDATED',
        footnotes: ['one', 'two']
      },
      admin,
    )
    
    expect(update.status).to.be.eq(200)
    expect(update.data.footnotes).to.be.eql(['one', 'two'])
  })
```

which results in:

```log
 FAIL  test/bookshop.test.js
  ● postgres › Bookshop - Update › Update array of

    AxiosError: 22P02 - invalid input syntax for type json
```

However, this is resolved with the current codeline, which is soon to be released. I added a test which shows it in #202.

nitiiz:
> Hi @nitiiz,
> 
> again, thanks for reporting. I have been able to reproduce the error:
> 
> `git checkout v1.1.0`:
> 
> ```js
>   test('Update array of', async () => {
>     // create book
>     const insert = INSERT.into('sap.capire.bookshop.Books')
>       .columns(['ID'])
>       .values([150])
>     await cds.run(insert)
> 
[...]

PR #202, opened 2023-08-30
test: add scenario for upsert `array of`

patricebender:
there was an issue with `postgres v1.1.0` - reported in https://github.com/cap-js/cds-dbs/issues/201 - which has been resolved in the meantime. Adding a test for bookkeeping.

### DIFF
--- a/test/scenarios/bookshop/update.test.js
+++ b/test/scenarios/bookshop/update.test.js
@@ -27,20 +27,20 @@ describe('Bookshop - Update', () => {
   test('Update array of', async () => {
     // create book
     const insert = INSERT.into('sap.capire.bookshop.Books')
-      .columns(['ID', 'footnotes'])
-      .values([150, ['one']])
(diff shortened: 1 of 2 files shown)


test/scenarios/bookshop/update.test.js
```diff
@@ -24,6 +24,24 @@ describe('Bookshop - Update', () => {
     expect(res.data.author_ID).to.be.eq(201)
     expect(res.data.descr).to.be.eq('UPDATED')
   })
+  test('Update array of', async () => {
+    // create book
+    const insert = INSERT.into('sap.capire.bookshop.Books')
+      .columns(['ID', 'footers'])
```
stewsk, line 30: Should `footers` be `footnotes`?

patricebender: thanks for catching that. It seems like the inital creation of the `footnotes` was not even necessary to resproduce the issue.

PR #131, opened 2023-07-24
[REVISIT:] Input/Output Converters

danjoa:
- [x] InputConverters streamlined → using `ISO` function; none for `Date` and `Time`
- [x] OutputConverters only where necessary → not when InputConverters did the job already
- [x] DateTimes and Timestamps are canonicalised to ms precision on write
- [x] DateTimes ms are cut off on read → using `substr` instead of `strftime`
- [x] [OutputConverters are applied when using jsons only ?!](#issuecomment-1647320690)
- [x] [INSERT from json_each() → relevant for HANA, but not for SQLite?](#issuecomment-1647323730)

PR #206, opened 2023-09-01
reproduce issue with timestamp as key

johannes-vogel:
Oliver has switched the odata-v2 proxy to use the new db layer in his tests.
This reveals a problem with datetime elements as key.

On first glance, the ISO function formats datetime values as timestamps which cannot be found anymore as we don't apply the same function on where clauses. Alternatively, we must not use the function for datetime elements.

caused by #131

### DIFF
--- a/sqlite/test/general/datetime.test.js
+++ b/sqlite/test/general/datetime.test.js
@@ -7,9 +7,11 @@ describe('datetime handling', () => {
     let res
     const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
     res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
-    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() })) // this finds the value
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() }))
     expect(res).toMatchObject(payload)
-    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: payload.dt})) // this returns undefined
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() }))
+    expect(res).toMatchObject(payload)
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: payload.dt}))
     expect(res).toMatchObject(payload)
   })
 })
--- a/sqlite/test/general/datetime.test.js
+++ b/sqlite/test/general/datetime.test.js
@@ -1,0 +1,14 @@
+const cds = require('../../../test/cds.js')
+
+cds.test(__dirname, 'testModel.cds')
+
+describe('datetime handling', () => {
+  test('datetime elements as key', async () => {
+    let res
+    const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
+    res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: payload.dt})) // this returns undefined
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: '2020-12-31T01:02:03.000Z'})) // this finds the value but shouldn't
+    expect(res).toMatchObject(payload)
+  })
+})
--- a/sqlite/test/general/datetime.test.js
+++ b/sqlite/test/general/datetime.test.js
@@ -7,7 +7,7 @@ describe('datetime handling', () => {
     let res
     const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
     res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
-    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(dt).toISOString() })) // this finds the value
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() })) // this finds the value
     expect(res).toMatchObject(payload)
     res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: payload.dt})) // this returns undefined
     expect(res).toMatchObject(payload)
--- a/sqlite/test/general/datetime.test.js
+++ b/sqlite/test/general/datetime.test.js
@@ -7,7 +7,7 @@ describe('datetime handling', () => {
     let res
     const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
     res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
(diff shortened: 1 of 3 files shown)


sqlite/test/general/datetime.test.js
```diff
@@ -0,0 +1,14 @@
+const cds = require('../../../test/cds.js')
+
+cds.test(__dirname, 'testModel.cds')
+
+describe('datetime handling', () => {
+  test('datetime elements as key', async () => {
+    let res
+    const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
+    res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: payload.dt})) // this returns undefined
```
BobdenOs, line 10:
```suggestion
    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: payload.dt})) // this returns undefined
    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload.dt)}))
```

This should functions the same as the next line. I don't think it is a valid solution, but wanted to make sure this scenario is also included. There is also a test currently skipped in `cds` which encountered the same issue.

johannes-vogel: update the test and comment

sqlite/test/general/datetime.test.js
```diff
@@ -0,0 +1,15 @@
+const cds = require('../../../test/cds.js')
+
+cds.test(__dirname, 'testModel.cds')
+
+describe('datetime handling', () => {
+  test('datetime elements as key', async () => {
+    let res
+    const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
+    res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(dt).toISOString() })) // this finds the value
```
sjvans, line 10:
```suggestion
    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() })) // this finds the value
```

sqlite/test/general/datetime.test.js
BobdenOs: The `toISOString` should not be required as `cqn2sql` should call it on any `Date` object. 

Also it should be possible to compare a `DateTime` with a `Timestamp` and find a match. As typed databased will do an implicit cast between the two types and match accordingly.

```SQL
TO_SECONDDATE('2020-12-31T01:02:03Z')     = TO_TIMESTAMP('2020-12-31T01:02:03.000Z') --> TRUE
TO_SECONDDATE('2020-12-31T01:02:03.123Z') = TO_TIMESTAMP('2020-12-31T01:02:03.000Z') --> TRUE
TO_SECONDDATE('2020-12-31T01:02:03Z')     = TO_TIMESTAMP('2020-12-31T01:02:03.123Z') --> FALSE
TO_TIMESTAMP('2020-12-31T01:02:03.123Z')  = TO_SECONDDATE('2020-12-31T01:02:03Z')    --> FALSE
```

HANA seems to cast to `Timestamp` when comparing `DateTime` with `Timestamp`. As the precision of the `Timestamp` can result in a mismatch.

sqlite/test/general/datetime.test.js
```diff
@@ -0,0 +1,15 @@
+const cds = require('../../../test/cds.js')
+
+cds.test(__dirname, 'testModel.cds')
+
+describe('datetime handling', () => {
+  test('datetime elements as key', async () => {
+    let res
+    const payload = { dt: '2020-12-31T01:02:03Z', int: 4711}
+    res = await cds.db.run(INSERT.into('DateTimeEntity').entries(payload))
+    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() })) // this finds the value
```
danjoa, line 10:
```suggestion
    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt).toISOString() })) // this finds the value
    res = await cds.db.run(SELECT.one.from('DateTimeEntity').where({dt: new Date(payload .dt) })) // this should work as well
```

sqlite/test/general/datetime.test.js
danjoa: The failing one is exactly one of the cases [documented](https://pages.github.tools.sap/cap/docs/guides/databases-sqlite#comparing-datetimes-timestamps) to fail:

For example, all these work:

```js
SELECT.from(Foo).where `someTimestamp = anotherTimestamp`
SELECT.from(Foo).where `someTimestamp = someDateTime`
SELECT.from(Foo).where `someTimestamp = ${new Date}`
SELECT.from(Foo).where `someTimestamp = ${req.timestamp}`
SELECT.from(Foo).where `someTimestamp = ${'2022-11-11T11:11:11.123Z'}`
```

While these would fail, because the input values don't comply to the rules:

```js
SELECT.from(Foo).where `createdAt = ${'2022-11-11T11:11:11+02:00'}` // non-zulu time zone
SELECT.from(Foo).where `createdAt = ${'2022-11-11T11:11:11Z'}` // missing 3-digit fractions
```

This is because we never can reliable infer the types of input to where clause expressions → it's strongly recommended to use always use `Date` objects there, e.g.: 

```js
SELECT.from(Foo).where { createdAt: new Date('2022-11-11T11:11:11Z') }`
```

danjoa:
The failing one is exactly one of the cases [documented](https://pages.github.tools.sap/cap/docs/guides/databases-sqlite#comparing-datetimes-timestamps) to fail:

For example, all these work:

```js
SELECT.from(Foo).where `someTimestamp = anotherTimestamp`
SELECT.from(Foo).where `someTimestamp = someDateTime`
SELECT.from(Foo).where `someTimestamp = ${new Date}`
SELECT.from(Foo).where `someTimestamp = ${req.timestamp}`
SELECT.from(Foo).where `someTimestamp = ${'2022-11-11T11:11:11.123Z'}`
```

While these would fail, because the input values don't comply to the rules:

```js
SELECT.from(Foo).where `createdAt = ${'2022-11-11T11:11:11+02:00'}` // non-zulu time zone
SELECT.from(Foo).where `createdAt = ${'2022-11-11T11:11:11Z'}` // missing 3-digit fractions
```

This is because we never can reliable infer the types of input to where clause expressions → it's strongly recommended to use always use `Date` objects there, e.g.: 

```js
SELECT.from(Foo).where { createdAt: new Date('2022-11-11T11:11:11Z') }`
```

johannes-vogel:
> The failing one is exactly one of the cases [documented](https://pages.github.tools.sap/cap/docs/guides/databases-sqlite#comparing-datetimes-timestamps) to fail:

I've seen this and therefore mentioned that we either need to support it on sqlite level or change the runtime to generate a compliant query. As of now, the runtime always sends ISO strings without ms precision.

johannes-vogel:
just tested and also works with current runtime implementation :)

BobdenOs (approved):

PR #207, opened 2023-09-04
fix: Ensure safe rollback and release

BobdenOs:
## Description

When a database connection encounters an error in a query it will trigger the default rollback mechanism. As all queries including rollback are asynchronous it is possible that queries are in progress or in the process of being send. Which can mean that the rollback happens before a parallel query. Which allows the query to escape the transaction. Additionally it is possible to send a query to the database service while it is in the process of rolling back. All new queries to a rollback database transaction should be rejected.

To prevent the database connection to be released back into the pool with queries that are in progress this PR implements a tracker of all returned promises. Which is awaited before the rollback is triggered. After which the database connection is safely released back to the connection pool.

```js
this.on(['*'], () => {
  const prom = next()
  this.activeQueries.push(prom)
  return prom
})
```

### DIFF
--- a/db-service/lib/common/DatabaseService.js
+++ b/db-service/lib/common/DatabaseService.js
@@ -35,6 +35,21 @@ class DatabaseService extends cds.Service {
     throw '2b overridden by subclass'
   }
 
+  init() {
+    this.on(['*'], this.onRequest)
+    return super.init()
+  }
+
+  /**
+   * Handler for SELECT
+   * @type {Handler}
+   */
+  async onRequest(req, next) {
+    const prom = next()
+    this.activeQueries.push(prom)
+    return prom
+  }
+
   /**
    * Acquires a pooled connection and starts a session, including setting
    * session context like `$user.id` or `$user.locale`, and starting a
@@ -55,6 +70,7 @@ class DatabaseService extends cds.Service {
 
     // Acquire a pooled connection
     this.dbc = await this.acquire()
+    this.activeQueries = []
 
     // Begin a session...
     try {
@@ -83,6 +99,7 @@ class DatabaseService extends cds.Service {
     if (!this.dbc) return
     else
       try {
+        await Promise.allSettled(this.activeQueries)
         await this.send('ROLLBACK')
       } finally {
         this.release()
--- a/db-service/lib/SQLService.js
+++ b/db-service/lib/SQLService.js
@@ -20,6 +20,7 @@ const cqn4sql = require('./cqn4sql')
 class SQLService extends DatabaseService {
 
   init() {
+    const ret = super.init()
     this.on(['SELECT'], this.transformStreamFromCQN)
     this.on(['UPDATE'], this.transformStreamIntoCQN)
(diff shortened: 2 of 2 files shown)


db-service/lib/common/DatabaseService.js
```diff
@@ -55,6 +70,7 @@ class DatabaseService extends cds.Service {
 
     // Acquire a pooled connection
     this.dbc = await this.acquire()
+    this.activeQueries = []
```
sjvans, line 73: shouldn't the active queries be tracked per connection?

db-service/lib/common/DatabaseService.js
```diff
@@ -35,6 +35,21 @@ class DatabaseService extends cds.Service {
     throw '2b overridden by subclass'
   }
 
+  init() {
+    this.on(['*'], this.onRequest)
```
sjvans, line 39: fyi, you could use [prepend](https://cap.cloud.sap/docs/node.js/core-services#srv-prepend) to register before others and, hence, keep `return super.init()` in SQLService

db-service/lib/common/DatabaseService.js
```diff
@@ -83,6 +99,7 @@ class DatabaseService extends cds.Service {
     if (!this.dbc) return
     else
       try {
+        await Promise.allSettled(this.activeQueries)
```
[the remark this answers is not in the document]
BobdenOs: Hanging queries is a possibility, but often caused by other side effects. Like the bug found in hdb [PR](https://github.com/SAP/node-hdb/pull/216#issue-1864835295).

An alternative solution would be to not release the database connection, but to disconnect/ destroy it in case of a rollback.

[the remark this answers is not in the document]
sjvans: > An alternative solution would be to not release the database connection, but to disconnect/ destroy it in case of a rollback.

i wouldn't be opposed to that, if the destroy doesn't await the stmt.exec. it's simple and effective, and rollbacks should happen very rarely, so very little impact.

sjvans (approved):
🤙

db-service/lib/common/DatabaseService.js
```diff
@@ -85,7 +85,8 @@ class DatabaseService extends cds.Service {
       try {
         await this.send('ROLLBACK')
       } finally {
-        this.release()
+        // Destroy connection as it might be tainted by parallel operations
+        this.destroy()
```
danjoa, line 89: Should we really destroy the connection on each rollback? 
Only when an error happened in acquire or begin, isn't it?

danjoa (changes requested):

danjoa:
Thinking twice about the approach to destroy connections on each rollback, to avoid reusing 'poisoned' `dbc`s on subsequent requests ('poisened' meaning: dangling operations in a promise chain). 

We might frequently rollback transactions due to failed input validations, after some validations read from the database. → in consequence, we would see many transactions destroyed and re-connected (each connect with a cost of ~40ms for HANA).

db-service/lib/common/DatabaseService.js
sjvans: no, this shall prevent executing statements prepared before the rollback.

PR #210, opened 2023-09-05
fix: better-sqlite3@8.6.0

BobdenOs:
`better-sqlite3@8.6.0` includes an upgrade of `sqlite3` which contains a performance optimization for `json` functions ([changelog](https://www.sqlite.org/releaselog/3_43_0.html#:~:text=Performance%20enhancements%20to%20JSON%20processing%20results%20in%20a%202x%20performance%20improvement%20for%20some%20kinds%20of%20processing%20on%20large%20JSON%20strings.))

The performance optimization seems to take the trade of that it uses more memory. Which made the `lean-draft` tests run into `SQLITE_NOMEM` (out of memory) errors. After some investigating this seems to be related to the solution used to support 50+ columns in select statements. This PR switches to use `json_patch` which is equal to `Object.assign({...},{...})`. Which seems to make the behavior as expected again. Might even be a small general performance increase.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -235,19 +235,20 @@ class CQN2SQLRenderer {
     if (!elements) return sql // REVISIT: Above we say this is an error condition, but here we say it's ok?
     let cols = SELECT.columns.map(x => {
       const name = this.column_name(x)
-      let col = `'$."${name}"',${this.output_converter4(x.element, this.quote(name))}`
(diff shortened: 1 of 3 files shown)


johannes-vogel:
Is that also compatible with 8.5.2?

BobdenOs:
Yes no changes where made to the `json` function APIs. Just the internals changed.

johannes-vogel (approved):

PR #212, opened 2023-09-05
Reopen Optimized deep delete #170

johannes-vogel:
This still causes some runtime tests to fail. We need to analyze before merging as this causes the runtime PRs to be blocked.

- [x] requires https://github.com/cap-js/cds-dbs/pull/213
- [x] requires https://github.com/cap-js/cds-dbs/pull/214 to be fixed
- [x] revert to hierarchy support and throw error for cycles
- [x] adapt runtime tests with cyclic compositions

PR #213, opened 2023-09-05
cqn4sql: pass context for association expansion

johannes-vogel:
When having a scoped query, as in the test example, we must pass the context for an infix filter (that is the entity where the filter is resolvable in) if the filter is not part of the last `ref` and hence not part of the `elements` of the query.

---

@patricebender FYI.
When testing #212 with the current runtime implementation, we observed that a request like this causes a crash in cqn4sql.
If you have some time, you could have a look. Assumption is that the association in the filter is flattened to its foreign key.

### DIFF
--- a/db-service/test/cqn4sql/tupleExpansion.test.js
+++ b/db-service/test/cqn4sql/tupleExpansion.test.js
@@ -260,6 +260,24 @@ describe('Structural comparison', () => {
       expect(query).to.deep.equal(CQL(expectedQueryString))
     })
   })
+
+  it('expands comparison also in exists subquery', () => {
+      const queryString = `SELECT from bookshop.AssocWithStructuredKey[toStructuredKey = null]:accessGroup { ID }`
+      let query = cqn4sql(CQL(queryString), model)
+      const expectedQueryString = `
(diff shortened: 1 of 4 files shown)


db-service/test/cqn4sql/tupleExpansion.test.js
```diff
@@ -260,6 +260,24 @@ describe('Structural comparison', () => {
...
+          SELECT 1 from bookshop.AssocWithStructuredKey as AssocWithStructuredKey
+          where AssocWithStructuredKey.accessGroup_ID = accessGroup.ID and
+              AssocWithStructuredKey.toStructuredKey_struct_mid_leaf        = null and
+              AssocWithStructuredKey.toStructuredKey_struct_mid_anotherLeaf = null and
+              AssocWithStructuredKey.toStructuredKey_second                 = null
+        )
+      `
+      expect(query).to.deep.equal(CQL(expectedQueryString))
+  })
+
```
patricebender, line 280: added another test which includes an expansion

patricebender (approved):

stewsk (approved):

PR #218, opened 2023-09-06
replace npm test:all script in release workflow

johannes-vogel:


.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           node-version: 18
           registry-url: https://registry.npmjs.org/
       - run: npm i
-      - run: npm run test:all
+      - run: npm test -w db-service -w sqlite -w postgres -- --maxWorkers=1
```
BobdenOs, line 23: once hana is also released it should be possible to replace it with:
`npm test -ws -- --maxWorkers=1`

BobdenOs (approved):

Issue #229, opened 2023-09-08
Postgres Quoted/Upper-case Column Alias in SELECT output doesn't seem correct


jung-thomas:
### Description of erroneous behaviour 

I'm testing on CAP  7.2.0 and @cap-js/postgres 1.2.1

I'm performing a SELECT and want to alias the column names to upper case. In both HANA and SQLite the query works just by passing  upper case literal to the as parameter

SQLite example that works as expected
```js
let dbQuery = SELECT
            .columns({ref:["name"],as:'TABLE_NAME'})
            .from("sqlite_schema")
            .where({ type: 'table', name: { like: tableName } })
            .limit(super.getPrompts().limit)
        base.debug(JSON.stringify(dbQuery))
        let results = await this.getDB().run(dbQuery)
```
https://github.com/SAP-samples/hana-developer-cli-tool-example/blob/8b6eab19d38361c5b59cf7a82f2ded8ed27b7e7b/utils/database/sqlite.js#L15

Name is in the results as TABLE_NAME
![image](https://github.com/cap-js/cds-dbs/assets/12159356/5ea48530-d9e0-47e5-84b8-4b1107f650a2)

But doing the same when targeting Postgres doesn't work the same.  The resulting column alias is still lower case.  So, I tried quoting within the literal with `"`.  That did produce an all-uppercase column name but, that name also contains the `"` in the output.  I can't get the same output as HANA and SQLite with the column alias in this example.

Postgres example
```js
     let dbQuery = SELECT
            .columns(
                {ref:["table_schema"],as: `"SCHEMA_NAME"` },
                {ref:["table_name"],as:'TABLE_NAME'} )
            .from("tables")
            .where({ table_schema: this.#schema, table_name: { like: tableName }, table_type: 'BASE TABLE' })
            .limit(super.getPrompts().limit)

            base.debug(JSON.stringify(dbQuery))
        let results = await this.getDB().run(dbQuery)
```
https://github.com/SAP-samples/hana-developer-cli-tool-example/blob/8b6eab19d38361c5b59cf7a82f2ded8ed27b7e7b/utils/database/postgres.js#L18

TABLE_NAME is returned all lower case and SCHEMA_NAME is upper but has the `"` in the output column:
![image](https://github.com/cap-js/cds-dbs/assets/12159356/c7d4ed4c-9aa8-4cfe-9ba5-5d5dd8e4f0c1)
 

### Details about your project

[...]

PR #230, opened 2023-09-11
fix(aliasing): alias name must be kept as requested

johannes-vogel:
Trying to reproduce https://github.com/cap-js/cds-dbs/issues/229

@BobdenOs Do you know by any chance whether this can be depending on pg version?


BobdenOs:
@johannes-vogel from the original issue it looks like it is being selected from the native table. Which means that it is not going through the same process as tables which are part of the model. Especially as the original issue uses `"` in the alias which looks like it is trying to prevent SQL syntax errors. Which would no longer be needed with the new database service as it would quote when needed.

johannes-vogel:
> @johannes-vogel from the original issue it looks like it is being selected from the native table. Which means that it is not going through the same process as tables which are part of the model. Especially as the original issue uses `"` in the alias which looks like it is trying to prevent SQL syntax errors. Which would no longer be needed with the new database service as it would quote when needed.

Thanks, now I can reproduce it.

BobdenOs:
This is a side effect of our minimal quoting approach. Where Postgres assumes every column to be lower case. Like how HANA assumes every column to be upper case.

johannes-vogel:
> This is a side effect of our minimal quoting approach. Where Postgres assumes every column to be lower case. Like how HANA assumes every column to be upper case.

I don't think that's what someone would expect. If I provide an alias within CQN, I'd expect that the result contains the property as specified in the CQN (regardless whether we know the entity or not).

johannes-vogel:
@BobdenOs I think we should allow selects the same way as we do for insert/update. I've updated the test to be executable on any db, but cqn2sql doesn't work with unknown entities yet. Could you please have a quick look on that?

patricebender:
I believe this can be closed, at least the alias is now kept as requested.

PR #233, opened 2023-09-11
calculated elements: fixes and enhancements

patricebender:
This change contains several fixes for the calculated elements on read:
- If multiple calculated elements share the same join node, make sure that refs are properly rewritten to the correct TA
- make sure that infix filter arguments in calculated elements are properly massaged into join arguments
  - make sure that such an infix filter can also be a calculated element by itself
- provide a base context for function arguments making sure that aliases are properly added to the refs

other changes:

- add test - which is currently skipped - that shows that calculated elements are currently not properly rejected after `where exists` arguments

### DIFF
--- a/sqlite/lib/SQLiteService.js
+++ b/sqlite/lib/SQLiteService.js
@@ -16,7 +16,6 @@ class SQLiteService extends SQLService {
         dbc.function('regexp', { deterministic: true }, (re, x) => (RegExp(re).test(x) ? 1 : 0))
         dbc.function('ISO', { deterministic: true }, d => d && new Date(d).toISOString())
         if (!dbc.memory) dbc.pragma('journal_mode = WAL')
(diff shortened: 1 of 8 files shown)


sqlite/lib/SQLiteService.js
```diff
@@ -16,6 +16,7 @@ class SQLiteService extends SQLService {
         dbc.function('regexp', { deterministic: true }, (re, x) => (RegExp(re).test(x) ? 1 : 0))
         dbc.function('ISO', { deterministic: true }, d => d && new Date(d).toISOString())
         if (!dbc.memory) dbc.pragma('journal_mode = WAL')
+        dbc.pragma('foreign_keys = ON')
```
patricebender, line 19:
```suggestion
```

stewsk (approved):

PR #235, opened 2023-09-12
`cds build` plugin support for Postgres

swaldmann:
- [x] auto-pulls build task for `cds build --production` if `@cap-js/postgres` is in `dependencies`
- [x] postgres build task can be explicitly configured using `{"for": "postgres"}`
- [x] postgres build task can be executed using CLI `cds build --for postgres`
- [x] requires `"@sap/cds-dk": "^7.3.0-2023092113412679-SNAPSHOT"`
- [x] Check `cds-dk` behavior as optional peer dependency → silently not installed with `npm i`, just what we need
- [x] End-to-end tests for Nodejs and Java missing including cloud deployment → will come with `cds add postgres`
- [x] adapt Capire documentation → https://github.com/cap-js/docs/pull/413
According to [Capire doc](https://cap.cloud.sap/docs/guides/databases-postgres#create-the-shell-script) files have to be created at `gen/pg/db`. This seems to be wrong - correct path (consistent with hana) would be `gen/db/pg` instead


swaldmann:
@patricebender @simonoswald Can you review this? This PR introduces native `cds build` support for Postgres, which automatically pulls once the plugin is loaded.

Essentially this replaces steps 1-4 here with a plain `cds build`:
https://pages.github.tools.sap/cap/docs/guides/databases-postgres#with-a-deployer-app

swaldmann:
Then let's do it :)

patricebender (approved):
hopefully now my review is counted as code owner approval

swaldmann:
Can be merged once cap/cds-dk/pull/2186 is merged.

johannes-vogel (approved):

juergenravnik:
The peer dependency `@sap/cds-dk` introduced by this pull request drastically increases build file size after `mbt build`. The build file size of a very basic CAP project using with that peer dependency (i.e. `@cap-js/postgres@1.3.1`) is 78MB.

An older project (using `@cap-js/postgres@1.2.1`) has a build file size of 7MB. So that new version apparently more than 10-folds the file size.

I am aware of using `ignore: ["node_modules/"]` in the `build-parameters` of the mta module... but I am unsure if that has further side effects.

Is there a recommended way to approach this and have the build file size be smaller without any side effects?

BobdenOs:
@juergenravnik the peer dependency should only be needed when doing the build which should allow it to be defined as a `devDependency` which allows you to `npm` install for production which should remove the `devDependencies` again.

For an MTA which only depends on public `node modules` it would be possible to ignore them from the upload. Which often makes the upload faster, but can introduce some side effects when not using `package lock` files properly. Which can also be done improperly by dependencies. The original recommendation has been to upload the `node modules` along with the application.

juergenravnik:
@BobdenOs it is not about the `devDependencies` in my own `package.json`, but rather the `peerDependencies` in the `package.json` of `@cap-js/postgres`.

`npm install --production` is the default command executed using the `npm` builder in the MTA config (see https://github.com/SAP/cloud-mta-build-tool/blob/master/configs/builder_type_cfg.yaml#L19) - so any dev dependencies shoud already be omitted with `mbt build`. I should have given more info about my project setup, sorry.

However, I just noticed the `peerDependenciesMeta` in the `package.json` of `@cap-js/postgres` in which the `@sap/cds-dk` module is marked as optional... or so I thought... shouldn't it be `@sap/cds-dk` and not just `cds-dk` here? https://github.com/cap-js/cds-dbs/blob/964f0dcbe7e7bb817bc9eb2796eac89eb591165d/postgres/package.json#L42

I am not sure if that relates to my problem, since I am not familiar with how npm resolves that kind of dependencies.

My current workaround is to just exclude `@sap/cds-dk` in the `build-parameters` of my MTA config: `ignore: ["node_modules/@sap/cds-dk/"]`. That way other node modules should stay in the build along with the application and the file size is down to 4 MB.

patricebender:
> However, I just noticed the peerDependenciesMeta in the package.json of @cap-js/postgres in which the @sap/cds-dk module is marked as optional... or so I thought... shouldn't it be @sap/cds-dk and not just cds-dk here?

yes, fixed with https://github.com/cap-js/cds-dbs/pull/305

PR #238, opened 2023-09-14
fix: Retain null columns for sqlite

BobdenOs:
Switching back to using `json_insert` / `json_set` for results sets that contain more then `~50` columns. As they would overflow the function arguments limit of `100`. To retain `null` valued columns. As the `Object.assign({a:[]},a:null)` returns `{a:null}` in javascript. For SQLite it follows [RFC-7396](https://datatracker.ietf.org/doc/html/rfc7396#:~:text=Null%20values%20in%20the%20merge) which defines `null` on the right hand side as a `delete` operation.


johannes-vogel:
@BobdenOs can we add a test to ensure regression protection?

db-service/lib/cqn2sql.js
```diff
@@ -233,16 +233,25 @@ class CQN2SQLRenderer {
...
+
       if (x.SELECT?.count) {
         // Return both the sub select and the count for @odata.count
         const qc = cds.ql.clone(x, { columns: [{ func: 'count' }], one: 1, limit: 0, orderBy: 0 })
-        return [col, `'${name}@odata.count',${this.expr(qc)}`]
+        cols.push(`${wrap(name + '@odata.count')},${this.expr(qc)}`)
       }
-      return col
-    }).flat()
+    }
```
soccermax, line 254: this preserves null values but leads to the situation that every column after the 49 column count looks the following pattern in the response:
`$."columnName"` this should be --> `columnName`

johannes-vogel (approved):

PR #240, opened 2023-09-15
cqn4sql: reject calc elements \w exists predicate

patricebender:
also streamline error handling to always use `cds.error`.

### DIFF
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -101,19 +101,19 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
       if (ref.length > 1) {
         target = from.ref.slice(1).reduce((d, r) => {
           const next = d.elements[r.id || r]?.elements ? d.elements[r.id || r] : d.elements[r.id || r]?._target
-          if (!next) cds.error(`No association "${r.id || r}" in ${d.kind} "${d.name}": ${d}`)
+          if (!next) cds.error(`No association “${r.id || r}” in ${d.kind} “${d.name}”`)
           return next
         }, target)
       }
       if (target.kind !== 'entity' && !target._isAssociation)
-        throw new Error(/Query source must be a an entity or an association/)
+        cds.error('Query source must be a an entity or an association')
 
       attachRefLinksToArg(from) // REVISIT: remove
       const alias =
         from.uniqueSubqueryAlias ||
         from.as ||
         (ref.length === 1 ? first.match(/[^.]+$/)[0] : ref[ref.length - 1].id || ref[ref.length - 1])
-      if (alias in querySources) throw new Error(`Duplicate alias "${alias}"`)
+      if (alias in querySources) cds.error(`Duplicate alias "${alias}"`)
       querySources[alias] = target
     } else if (from.args) {
       from.args.forEach(a => inferTarget(a, querySources))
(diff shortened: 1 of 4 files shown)


johannes-vogel (commented):
Is there a reason to switch to `cds.error`?
`throw new Error` also works fine.

patricebender:
> Is there a reason to switch to `cds.error`? `throw new Error` also works fine.

I had both styles mixed up and wanted to get rid of one. Would you prefer `throw new Error`?

danjoa:
I'd say use `cds.error` only if you want one of its advanced features, not supported by `new Error` like enriching existing error objects, adding details objects as a secondary argument or as a throwing fallback. 

If you use `cds.error` always add a `throw` before it unless using it as a conditional throwing fallback:

```js
let x = argx || cds.error(`Expected argx to be defined`)
```

db-service/test/cqn4sql/calculated-elements.test.js
```diff
@@ -606,12 +606,34 @@ describe('Unfolding calculated elements in select list', () => {
...
 
-  // TODO
-  it.skip('where exists cannot leverage calculated elements', () => {
+  it('exists cannot leverage calculated elements', () => {
     // at the leaf of a where exists path, there must be an association
     // calc elements can't end in an association, hence this does not work, yet.
-    let query = cqn4sql(CQL`SELECT from booksCalc.Books { ID } where exists youngAuthorName`, model)
-    expect(query).to.deep.throw()
+    expect(() => cqn4sql(CQL`SELECT from booksCalc.Books { ID } where exists youngAuthorName`, model)).to.throw(
+      'Unexpected calculated element “youngAuthorName” in path preceded by “exists” predicate',
```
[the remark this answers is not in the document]
danjoa: Btw: is that also (to be) handled in path expressions? 
E.g.
```sql
exists someAssoc.anotherAssoc.someCalculatedElement
```

danjoa (changes requested):

db-service/test/cqn4sql/calculated-elements.test.js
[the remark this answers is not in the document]
patricebender: > Btw: is that also (to be) handled in path expressions? E.g.
> 
> ```sql
> exists someAssoc.anotherAssoc.someCalculatedElement
> ```

yes it is :) I have locally added another test with path expressions.

I will incorporate your feedback regarding the error message. Also I will from now on use `throw new Error` until I find a good reason to use `cds.error`.

db-service/lib/infer/index.js
```diff
@@ -97,16 +97,16 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
     if (ref) {
       const first = ref[0].id || ref[0]
       let target = getDefinition(first, model)
-      if (!target) cds.error(`"${first}" not found in the definitions of your model`)
+      if (!target) throw new Error(`"${first}" not found in the definitions of your model`)
       if (ref.length > 1) {
         target = from.ref.slice(1).reduce((d, r) => {
           const next = d.elements[r.id || r]?.elements ? d.elements[r.id || r] : d.elements[r.id || r]?._target
-          if (!next) cds.error(`No association "${r.id || r}" in ${d.kind} "${d.name}": ${d}`)
+          if (!next) throw new Error(`No association “${r.id || r}” in ${d.kind} “${d.name}”`)
```
johannes-vogel, line 104:
```suggestion
          if (!next) throw new Error(`No association "${r.id || r}" in ${d.kind} "${d.name}"`)
```

Or is there a reason for using `”`?

patricebender: Just personal taste. Also, we use those unicode quotes throughout the compiler messages, hence I used them in most error messages. With this change, I also commited to one quotation style. I dont mind using `"` instead, if you wish.

johannes-vogel: I just stumbled over it and I think this is something we should align.

patricebender: I agree, that is why I wanted to use the same quotation style as in the compiler. Maybe we need to bring this up for a broader audience (runtimes, tooling etc). Besides that, is this change okay for now?

johannes-vogel (approved):

PR #242, opened 2023-09-18
fix: do not create case when stmts for comparison operators

johannes-vogel:


BobdenOs (approved):

danjoa (commented):
Let's pls review in sync meeting

hana/lib/HANAService.js
```diff
@@ -646,21 +646,22 @@ class HANAService extends SQLService {
...
-        '<=': null,
-        '!<': null,
-        '!>': null,
+        // '>': null,
+        // '<': null,
+        // '<>': null,
+        // '>=': null,
+        // '<=': null,
+        // '!<': null,
+        // '!>': null,
```
danjoa, line 659: Would that change specified behavior?

BobdenOs: This change only impacts the usage of these operators where `HANA` does not allow for them.

```SQL
SELECT 1>0 as gt FROM DUMMY
```

Currently considering to close this PR as https://github.com/cap-js/cds-dbs/pull/426 does the same with more fixes.

johannes-vogel:
obsolete due to #426

PR #243, opened 2023-09-18
Fix: don't release to pool twice

danjoa:
Replacing https://github.com/cap-js/cds-dbs/pull/207 until we clarified the idea of destroying connections on each rollback.


patricebender:
is there a changelog entry for this?

johannes-vogel:
it's not yet released

PR #244, opened 2023-09-18
Less Typescript artefacts

danjoa:


db-service/lib/cqn2sql.js
```diff
@@ -910,4 +910,3 @@ const _empty = a => !a || a.length === 0
  */
 module.exports = (q, m) => new CQN2SQLRenderer().render(cqn4sql(q, m), m)
 module.exports.class = CQN2SQLRenderer
-module.exports.classDefinition = CQN2SQLRenderer // class is a reserved typescript word
```
BobdenOs, line 913: As commented `class` is a reserved word. Which should not be a problem, but it turns out that the `.d.ts` files it generates tries to define the property name as the variable name.

```typescript
import { class } from './cqn2sql'
```

danjoa: Didn't we agree to not generate `.d.ts` files out of our `.js` sources? 
This PR is in effect of that

PR #246, opened 2023-09-21
Fixed loading cds.env before cds.test()

danjoa:
Many tests erroneously loaded `cds.env` from arbitrary folder before switching to the target folder with `cds.test()`. 
Some of these tests silently ran with the old sqlite service.


db-service/test/deep/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@cap-js/sqlite": "latest"
+  }
+}
```
patricebender, line 5:
```suggestion
}

```

patricebender (approved):
thanks

johannes-vogel (approved):

PR #249, opened 2023-09-25
feat(temporal data): add time slice key to conflict clause

sjvans:
- [x] test
- [x] ~changelog~ will happen automatically once merged


patricebender (commented):
@sjvans  Could you add a test which shows the effects of this change?

db-service/lib/cqn2sql.js
```diff
@@ -862,8 +865,8 @@ class CQN2SQLRenderer {
       if (converter && sql[0] !== '$') sql = converter(sql, element)
 
       let val = _managed[element[annotation]?.['=']]
+        || !isUpdate && (element['@cds.valid.from'] ? '$valid.from' : element['@cds.valid.to'] ? '$valid.to' : null)
```
johannes-vogel, line 868: I think we did not support this so far as we cannot achieve this on SAP HANA. We do an `UPSERT... WITH PRIMARY KEY` and do not list the specific columns.

BobdenOs: @johannes-vogel I am not 100% certain what your concern is.

The current HANA UPSERT implementation creates a join condition. [here](https://github.com/cap-js/cds-dbs/blob/main/hana/lib/HANAService.js#L641)

Which should look something like:
```SQL
UPSERT "TABLE" (
  A,B,C
) SELECT
  NEW.A,
  COALESCE(NEW.B,OLD.B) AS B,
  CASE OLD.A IS NULL THEN NEW.C ELSE OLD.C END
FROM (SELECT ... FROM JSON_TABLE(...))
LEFT JOIN "TABLE" AS OLD ON NEW.A=OLD.A
```

nocin:
Hi all,

thanks for moving this topic forward! Am I right, that only one check is missing until it could be merged? That would be awesome!

BR, Nico

patricebender (approved):

johannes-vogel:
> Hi all,
> 
> thanks for moving this topic forward! Am I right, that only one check is missing until it could be merged? That would be awesome!
> 
> BR, Nico

@nocin Is it an option for you to make `@cds.valid.from` element a key? That should already work today.

nocin:
> > Hi all,
> > thanks for moving this topic forward! Am I right, that only one check is missing until it could be merged? That would be awesome!
> > BR, Nico
> 
> @nocin Is it an option for you to make `@cds.valid.from` element a key? That should already work today.

Hi @johannes-vogel,

thanks for the hint! I now spend some time thinking about this approach, but it would mean a lot of work in backend, frontend and another external program which pushes data in the application, as in our case it is the "main" table of the project. Therefore, I would prefer to stick to a single UUID as key and will wait for this feature, until I continue working on my feature request.

BR, Nico

hana/lib/HANAService.js
```diff
@@ -837,19 +841,10 @@ class HANAService extends SQLService {
...
-            break
-          case '$now':
-            managed = this.func({ func: 'session_context', args: [{ val: '$user.now' }] })
-            break
-          default:
-            managed = undefined
-        }
-
+        const val = _managed[element[annotation]?.['=']]
+          || !isUpdate && (element['@cds.valid.from'] ? '$valid.from' : element['@cds.valid.to'] ? '$valid.to' : null)
```
johannes-vogel, line 845: Why is `|| !isUpdate && (element['@cds.valid.from'] ? '$valid.from' : element['@cds.valid.to'] ? '$valid.to' : null)` required?

db-service/lib/cqn2sql.js
johannes-vogel: I was referring to the old db implementation but the concern is invalid anyway as the `@cds.valid.from` column is marked as table key anyway.

hana/lib/HANAService.js
sjvans: otherwise you cannot insert/ upsert data without specifying both `validFrom` and `validAt` in the payload.
in other words:
`coalesce(ISO(value->>'$."validFrom"'), session_context('$valid.from'))`
instead of just
`ISO(value->>'$."validFrom"')`.

sjvans: @BobdenOs @johannes-vogel 
was the default `validTo` changed? i now get `$now` but i would have expected something like `9999-12-31T23:59:59.999Z`...

BobdenOs: `session_context('$now')` will return the timestamp from the `tx.timestamp`. Or is the `$now` part of the `entries || data` ?

sjvans: https://github.com/cap-js/cds-dbs/blob/87208af735941369a0f926873f5ddb950734b917/db-service/lib/common/session-context.js#L27 defaults to `$now` (logically)

BobdenOs: I see what you mean. I don't know what the behavior used to be. 
I am also not sure whether the max timestamp default is correct in the temporal context as I would expect that any new insert with the same key values should become the `validTo` of the last value.

```js
await INSERT({ID: 1/*, validFrom: $now, validTo: timestamp_max */}).into(temporal)
await sleep(...)
await Promise.all([
  INSERT({ID:1/*, validFrom: $now, validTo: timestamp_max */}).into(temporal),
  UPDATE(temporal).data({validTo: $now}).where(`ID = ${1} and validTo = ${timestamp_max}`),
])
```

johannes-vogel: I think we did not support/document how to write temporal data so far.
For now I think it's fine that the input is required by the user → #344

johannes-vogel (approved):

db-service/lib/cqn2sql.js
```diff
@@ -874,7 +874,6 @@ class CQN2SQLRenderer {
       if (converter && sql[0] !== '$') sql = converter(sql, element)
 
       let val = _managed[element[annotation]?.['=']]
-        || !isUpdate && (element['@cds.valid.from'] ? '$valid.from' : element['@cds.valid.to'] ? '$valid.to' : null)
```
BobdenOs: This does not look correct. If the session context contains the `from` and `to` it should be considered.

PR #251, opened 2023-09-26
feat: SELECT returns LargeBinaries as streams unless feature flag "stream_compat" is set

vkozyura:


patricebender:
whats the state of this PR?

sqlite/lib/SQLiteService.js
```diff
@@ -232,6 +232,31 @@ class SQLiteService extends SQLService {
       throw _not_unique(err, 'UNIQUE_CONSTRAINT_VIOLATION') || err
     }
   }
+
+  _convertStreamValues(values) {
+    let any
+    values.forEach((v, i) => {
+      if (v && typeof v.pipe === 'function') {
```
BobdenOs, line 239: Checking for the `pipe` function can be a bit dangerous. As the follow up code is using `on` to the value. Which might not be there. By checking for the `Readable` inheritance it should guarantee most of the expected APIs to be available. 

```suggestion
      if (v && v instanceof Readable) {
```

sqlite/test/general/stream.test.js
```diff
@@ -84,208 +77,4 @@ describe('STREAM', () => {
...
-      test('WRITE stream property on view', async () => {
-        const { ImagesView } = cds.entities('test')
-        const stream = fs.createReadStream(path.join(__dirname, 'samples/test.jpg'))
-
-        const changes = await STREAM.into(ImagesView, 1).column('renamedData').data(stream)
-        expect(changes).toEqual(1)
-        await readStream(1, ImagesView)
-      })
-
-      test('WRITE dataset from json file stream', async () => {
```
BobdenOs, line 235: It would be nice to retain this functionality. Probably should be revisited after the clean up.

BobdenOs (commented):
`db-service/test/cqn4sql/stream.test.js` should also be adjusted with this PR. As it is testing the `STREAM` apis.

db-service/lib/SQLService.js
```diff
@@ -79,7 +39,16 @@ class SQLService extends DatabaseService {
...
       if (cqn.SELECT.expand) rows = rows.map(r => (typeof r._json_ === 'string' ? JSON.parse(r._json_) : r._json_ || r))
-    if (cqn.SELECT.count) rows.$count = await this.count(query, rows)
+    if (cqn.SELECT.count) rows.$count = await this.count(query, rows) 
+    if (cds.env.features.compat_stream_cqn) {
+      if (query._streaming) {
+        this._changeToStreams(cqn, rows, true)
+        return rows.length ? { value: Object.values(rows[0])[0] } : undefined
+      } 
+    } else {  
+      this._changeToStreams(cqn, rows)
```
BobdenOs, line 49: Calling `_changeToStreams` should not be done in the base `SQLService` as it is only implemented in `SQLite`.

sqlite/lib/SQLiteService.js
```diff
@@ -232,6 +232,31 @@ class SQLiteService extends SQLService {
...
+        any = values[i] = new Promise(resolve => {
+          const chunks = []
+          v.on('data', chunk => chunks.push(chunk))
+          v.on('end', () => resolve(Buffer.concat(chunks)))
+          v.on('error', () => {
+            v.removeAllListeners('error')
+            v.push(null)
+          })
+        })
```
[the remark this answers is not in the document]
vkozyura: @BobdenOs: Are you sure that also all error listeners are removed in this case?

sqlite/test/general/stream.test.js
vkozyura: tests are changed, the idea is left as a comment

db-service/lib/SQLService.js
vkozyura: kind of solution implemented

sqlite/lib/SQLiteService.js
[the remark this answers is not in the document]

db-service/lib/SQLService.js
```diff
@@ -27,46 +26,45 @@ class SQLService extends DatabaseService {
...
-    const elements = query._target?.elements || target?.elements
-    if (!elements) next()
-    for (const key in elements) {
-      const element = elements[key]
-      if (element['@Core.MediaType'] && data[key]?.pipe) col = key
-      if (element['@Core.IsMediaType'] && data[key]) type = key
-      if (element['@odata.etag'] && data[key]) etag = key
+    // REVISIT: remove after removing compat_stream_cqn feature flag 
+    if (compat) { 
+      rows[0][Object.keys(rows[0])[0]] = this._stream(Object.values(rows[0])[0])
```
etimr, line 38: `Object.values(rows[0])[0]` instead of `rows[0][Object.keys(rows[0])[0]]`

etimr (approved):

vkozyura (commented):
Add changelog

patricebender (changes requested):
Please adhere to [conventional commit](https://github.com/cap-js/cds-dbs/blob/main/CONTRIBUTING.md#committing) convention, with this, the changelog will be automatically created

db-service/lib/SQLService.js
```diff
@@ -27,62 +26,77 @@ class SQLService extends DatabaseService {
...
+      }
+    })    
   }
 
   /**
    * Handler for SELECT
    * @type {Handler}
    */
   async onSELECT({ query, data }) {
+    query.SELECT.expand = 'root'
```
danjoa, line 79:
```suggestion
```

vkozyura: @danjoa: We moved `expand = 'root'` from cqn2sql back to onSELECT, because only there we are able to modify it in case of streaming. Otherwise expand is always set to 'root', which leads to building a json, which breaks streaming.

postgres/test/service-types.test.js
```diff
@@ -2,6 +2,8 @@ const { resolve } = require('path')
 const cds = require('../../test/cds.js')
 if (cds.env.fiori) cds.env.fiori.lean_draft = true
 else cds.env.features.lean_draft = true
+// REVISIT: remove after streaming PR is merged in cds
+cds.env.features.compat_stream_cqn = true
```
johannes-vogel, line 6: flag name

BobdenOs (approved):

patricebender (approved):

PR #248, opened 2023-09-22
chore(search): get rid of dependency to internal API

patricebender:
this basically is a copy of the functionality of the coding of the orignal function. Removed deprecated code and simplified it a little

---

also a prerequisite to support deep search with path expressions

PR #252, opened 2023-09-26
feat(cqn4sql/infer): support for deep search

patricebender:
support search via path expressions:

- deep search via associations
- include calculated elements if explicitly requested
- keep existing caching mechanisms

based on #248


patricebender:
Closed in favor of #590

PR #254, opened 2023-09-26
$now on sqlite

sjvans:
- [ ] changelog
- [ ] check deploy to postgres

---

given
```cds
entity Foo {
  $now as now : Timestamp
}
```
old `ISO(CURRENT_TIMESTAMP)` results in `2023-09-26T19:26:14.000Z` -> 2 hours off and no millisecond resolution
whereas new `STRFTIME('%Y-%m-%dT%H:%M:%fZ','NOW')` results in `2023-09-26T21:24:10.672Z`

alternatively, we could look into something like
```js
dbc.function('ISO', { deterministic: true }, d => {
  if (!d) return d
  if (!d.match(/Z$/)) d += 'Z'
  return d && new Date(d).toISOString()
})
```
but i believe `CURRENT_TIMESTAMP` will never have ms precision on sqlite.

where else would we need to adjust?

would it be possible to create a function `CURRENT_UTCTIMESTAMP` like we (should) use on hana?


db-service/lib/SQLService.js
```diff
@@ -170,7 +170,7 @@ class SQLService extends DatabaseService {
       // REVISIT: this is a hack the target of $now might not be a timestamp or date time
       // Add input converter to CURRENT_TIMESTAMP inside views using $now
       if(/^CREATE VIEW.* CURRENT_TIMESTAMP[( ]/is.test(query)) {
-        query = query.replace(/CURRENT_TIMESTAMP/gi, 'ISO(CURRENT_TIMESTAMP)')
+        query = query.replace(/CURRENT_TIMESTAMP/gi, "STRFTIME('%Y-%m-%dT%H:%M:%fZ','NOW')")
```
johannes-vogel, line 173: What's the difference between `CURRENT_TIMESTAMP` and `NOW`?

The `ISO` function is a js plugin and simply creates a `new Date().toISOString()`

sjvans: `CURRENT_TIMESTAMP` only has s precision, afaik

sjvans: > The `ISO` function is a js plugin and simply creates a `new Date().toISOString()`

yes, see addition to description for alternative solution by modifying the `ISO` function.

db-service/lib/SQLService.js
```diff
@@ -170,7 +170,7 @@ class SQLService extends DatabaseService {
       // REVISIT: this is a hack the target of $now might not be a timestamp or date time
       // Add input converter to CURRENT_TIMESTAMP inside views using $now
       if(/^CREATE VIEW.* CURRENT_TIMESTAMP[( ]/is.test(query)) {
```
sjvans, line 172:
```suggestion
      if(this.options.kind === 'sqlite' && /^CREATE VIEW.* CURRENT_TIMESTAMP[( ]/is.test(query)) {
```
i need this locally to not apply the patch during deployment to postgres... why hasn't anybody else run into this?

johannes-vogel: I guess we don't test the replacement yet.

sjvans:
@danjoa @BobdenOs @patricebender any opinions?

patricebender:
> @danjoa @BobdenOs @patricebender any opinions?

If the requirement is to get to `ms` precision, this is definitely an improvement. Before this change, it looks like we used `UTC` time. Do we want to present the date as local time in this case?

sjvans:
> > @danjoa @BobdenOs @patricebender any opinions?
> 
> If the requirement is to get to `ms` precision, this is definitely an improvement. Before this change, it looks like we used `UTC` time. Do we want to present the date as local time in this case?

original problem: `SELECT CURRENT_TIMESTAMP` -> `YYYY-MM-DD HH:MM:SS` in utc but without zulu appendix. `ISO()` does not ensure zulu but `new Date()` then assumes it's local time -> we get a timestamp with the offset doubled

ms precision is additional, but prevents a fix in `ISO()` (cf. https://github.com/cap-js/cds-dbs/pull/254#issue-1914324847)

db-service/lib/SQLService.js
johannes-vogel: I've double checked and the function `onPlainSQL` is overridden in postgres.

johannes-vogel:
> > @danjoa @BobdenOs @patricebender any opinions?
> 
> If the requirement is to get to `ms` precision, this is definitely an improvement. Before this change, it looks like we used `UTC` time. Do we want to present the date as local time in this case?

We always assume that the timestamps are in zulu timezone. the current behavior is unexpected as the current_timestamp of a view definition looks like zulu time zone but is two hours off. We definetely must fix the time zone. ms precision is a nice side effect imo.

db-service/lib/SQLService.js
sjvans: then we should probably check why deploy to postgres reached there...

johannes-vogel (approved):

PR #264, opened 2023-10-06
Update CODEOWNERS

patricebender:
Adding the product owners as code owners to the more specific paths, so that they are actually able to approve changes in those locations as well


patricebender:
`CODEOWNER` files are [parsed top to bottom and the last match for a given file wins](https://github.blog/2017-07-06-introducing-code-owners/).

johannes-vogel (approved):

PR #265, opened 2023-10-09
search: generic error for search with multiple arguments

patricebender:
for #127


db-service/lib/cql-functions.js
```diff
@@ -14,18 +14,6 @@ const StandardFunctions = {
...
    */
   average: x => `avg(${x})`,
-  /**
-   * Generates SQL statement that produces a boolean value indicating whether the search term is contained in the given columns
-   * @param {string} ref
-   * @param {string} arg
-   * @returns {string}
-   */
-  search: function (ref, arg) {
-    if (!('val' in arg)) throw `SQLite only supports single value arguments for $search`
```
johannes-vogel, line 24: Why don't you make the error here more generic? It shares the implementation for all databases.

patricebender: I thought of supporting search with multiple args for postgres as shown in #132 so separating the functions seemed like a good first step

patricebender: ... applied your suggestion :)

johannes-vogel (approved):

PR #267, opened 2023-10-09
db-service: enable anonymous expand on (implicit) table alias

patricebender:
enable queries like:

```cds
select from Books {
  { * } as FOO // wildcard elements grouped into `Books.elements.FOO.elements`
}


select from Books as Bar {
  Bar { * } // wildcard elements grouped into `Bar.elements.Bar.elements`
}
```

that gives users the ability to group the elements coming from the wildcard


patricebender:
getting `cds.error`Query was not inferred and includes expand. For which the metadata is missing.`, need to investigate

stewsk (approved):

PR #269, opened 2023-10-09
add test to reproduce deep delete issue with mixin compositions

johannes-vogel:


SamuelBrucksch:
Will be fixed in cds.

PR #307, opened 2023-10-17
chore: Add api tests for affected rows

SamuelBrucksch:
Added these tests to make sure, all DBs behave exactly the same for affectedRows.

We already have these kind of tests spread all over the place, should we remove them and just keep this new test, where we have them all together in one file?


test/compliance/api.test.js
```diff
@@ -0,0 +1,35 @@
...
+    // InsertResult
+    expect(affectedRows).to.include({ affectedRows: 1 }) // lastInsertRowid not available on postgres
+  })
+
+  test('Update returns affected rows', async () => {
+    const authors = await cds.db.run(SELECT.from('complex.Books'))
+    
+    const affectedRows = await cds.db.run(UPDATE.entity('complex.Books').data({title: 'Book'}))
+    expect(affectedRows).to.be.eq(authors.length)
+  })
```
danjoa, line 29:
```suggestion
    const { count } = await SELECT.one`couint(*)`.from('complex.Books')
    
    const affectedRows = await UPDATE.entity('complex.Books').data({title: 'Book'})
    expect(affectedRows).to.be.eq(count)
  })
```

test/compliance/api.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+  test('Upsert returns affected rows', async () => { 
+    const affectedRows = await cds.db.run(UPSERT.into('complex.Books').entries({ID: 9999999, title: 'Book'}))
```
danjoa, line 32:
```suggestion
    const affectedRows = await UPSERT.into('complex.Books').entries({ID: 9999999, title: 'Book'})
```

test/compliance/api.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Tests explicitely, that all DBs behave exactly the same for affected rows
+ */
+describe('affected rows', () => {
+const { expect } = cds.test(__dirname + '/resources')
+
+  test('Delete returns affected rows', async () => {
+    const del = DELETE.from('complex.Books').where('ID = 4712')
+    const affectedRows = await cds.db.run(del)
```
danjoa, line 11:
```suggestion
    const affectedRows = await DELETE.from('complex.Books').where('ID = 4712')
```

danjoa (approved):

PR #310, opened 2023-10-19
fix(search): check calculated columns at any depth

BobdenOs:
# Description

Search was being applied to calculated columns which where defined on the source of a projection. Rather then on the searched projection.


patricebender:
@BobdenOs @danjoa was that the reason the `cds` tests were failing?

patricebender (approved):

PR #311, opened 2023-10-19
fix: Add countdistinct to hana functions

BobdenOs:
The SFlight tests started to fail with `countdistinct` not being an HANA aggregate function.


johannes-vogel:
done in https://github.com/cap-js/cds-dbs/pull/332

PR #312, opened 2023-10-19
deps: upgrade to better-sqlite@9?

danjoa:


johannes-vogel:
According to https://github.com/WiseLibs/better-sqlite3/releases/tag/v9.0.0
and https://www.sqlite.org/releaselog/3_43_2.html
look fine to me.

johannes-vogel (approved):

PR #319, opened 2023-10-23
feat: adapt build plugin implementation to new API

swaldmann:
requires `@sap/cds-dk` 7.3.2


postgres/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 - The format is based on [Keep a Changelog](http://keepachangelog.com/).
 - This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Version 1.3.2 - tbd
+
+### Changed
+
+- Using the new `cds build` plugin API (requires `@sap/cds-dk` 7.3.1)
```
patricebender, line 11: You dont have to take care of the changelog here, just follow the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) convention and the changelog will be automagically generated for you. :)

patricebender (changes requested):

swaldmann:
Holding off for `cds-dk 7.3.2`

postgres/lib/build.js
```diff
@@ -1,37 +1,12 @@
...
-  ({ BuildPlugin } = require('@sap/cds-dk/lib/build'))
-} catch (e) {
-  if (e.code === 'ENOTFOUND') throw `No build plugin mechanism for @sap/cds-dk found. Please install @sap/cds-dk for development using 'npm i -D @sap/cds-dk@^7.3.0'`
-  else throw e
-}
-
-const { fs, path, rimraf } = cds.utils
-
-module.exports = class PostgresBuildPlugin extends BuildPlugin {
-  static hasTask() { // REVISIT: should be unnecessary -> plugin mechanism knows what to pull
```
LotharBender, line 14: The `hasTask` impl has ensured that a postgres build task is only executed if a postgres database is used according to the effective cds configuration. E.g. a `sqlite` database might have been configured for the `development` profile. Of course, this might be an edge case as cds build will normally only be executed for the `production` and not for the `development` profile...

swaldmann: Yes, unfortunately I also don't see another way to make it work properly w/ the profile technique where you switch between HANA/Postgres based on config profiles. I re-added the `hasTask` again.

LotharBender: ...unfortunately? - I think this is a perfect solution :)

swaldmann: I think it's close to perfect, but ideally the plugin doesn't handle what the plugin orchestrator in `cap/cds` should do, deciding which plugins to pull.

I'd rather have opt-out mechanism like this instead, which is then also combinable with profiles:
```json
"cds": {
  "requires": {
      "plugins": {
        "[pg]":  {
           "@sap/cds-hana": false
        },
        "[hana]": {
          "@cap-js/postgres": false
        }
     }
  }
}
```

Opting out in non-centralized places like the build plugin introduces inconsistent behavior, e.g. the plugin will still show up as loaded with `DEBUG=plugins`.

Also, it reduces the number of functions you have to implement from 2 to 1, which is huge imo.

LotharBender: But isn't this more a question of different lifecycles for `cds plugin` versus `cds build task`. Loading the `cds plugin` does not necessarily mean that a corresponding build task respectively its BuildPlugin counterpart will get executed.

I think the `cds add plugin` would behave similar. A `cds plugin` might get loaded, but the concrete `cds facet plugin` might not, as this is determined based on the concrete `cds add` command line parameters or some other facet dependencies.

postgres/cds-plugin.js
```diff
@@ -1,9 +1,10 @@
...
 
 if (!cds.env.fiori.lean_draft) {
   throw new Error('"@cap-js/postgres" only works if cds.fiori.lean_draft is enabled. Please adapt your configuration.')
 }
 
-if (cds.cli.command === 'build') {
-  module.exports = require('./lib/build')
-}
+cds.build?.register?.('postgres', {
+  impl: __dirname + '/lib/build.js',
```
LotharBender, line 8: Is there a reason for using `__dirname + '/lib/build.js'` instead of `@cap-js/postgres/lib/build.js`?
`require` should correctly resolve...

postgres/cds-plugin.js
```diff
@@ -1,9 +1,10 @@
...
+  taskDefaults: { src: cds.env.folders.db, dest: 'pg' }
```
LotharBender, line 9: I think configuring a default for the `dest` folder is not a good idea as this will create the build output in the wrong folder for Java apps. Instead the `pg` folder has to be passed to the corresponding `copy.to` or `write.to` methods 
Node.js: `./gen/pg/**` (build.target folder is `gen`)
Java: `./db/pg/gen/db/**` (build.target folder is `.`)

**Note:** Paths are resolved in the following way - `<cds.root>/<build.target>/<task.dest|task.src>/gen`

LotharBender (changes requested):

LotharBender (approved):

LotharBender:
Would be great if we could ship this version soon as the current postgres plugin version does not support @sap/cds-dk 7.3.2. See https://github.tools.sap/cap/issues/issues/14748

swaldmann:
Open points re simplifications to be discussed in follow-up.

swaldmann:
@patricebender @johannes-vogel Can we ignore HANA tests for changes in `cds-dbs/postgres` in GHA? Or rather an opt-in and just run HANA integration tests for changes in `db-service`, `test`, and `hana`?

patricebender (approved):

PR #322, opened 2023-10-26
fix: time function casting

larslutz96:
The time functions are now correctly casted.

fix: add validations and reject certain previously supported queries.
      Added to align the behavior of Postgres, SQLite and Hana.

### DIFF
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - `cds build`-relevant files are now correctly packaged into the release. #266
+- The time functions are now correctly casted
 
 ## Version 1.3.0 - 2023-10-06
 
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -9,7 +9,6 @@
 ### Fixed
 
 - `cds build`-relevant files are now correctly packaged into the release. #266
-- The time functions are now correctly casted
 
 ## Version 1.3.0 - 2023-10-06
 
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -4,7 +4,7 @@
 - The format is based on [Keep a Changelog](http://keepachangelog.com/).
 - This project adheres to [Semantic Versioning](http://semver.org/).
 
-## Version 1.3.1 - 2023-10-10
+## [1.3.1](https://github.com/cap-js/cds-dbs/compare/v1.3.0...v1.3.1) (2023-10-10)
 
 ### Fixed
 
--- a/postgres/lib/func.js
+++ b/postgres/lib/func.js
@@ -1,4 +1,5 @@
 const session = require('./session.json')
+const cds = require('../../cds')
 
 const StandardFunctions = {
   session_context: x => {
@@ -10,16 +11,63 @@ const StandardFunctions = {
   contains: (...args) => `(coalesce(strpos(${args}),0) > 0)`,
   indexof: (x, y) => `strpos(${x},${y}) - 1`, // sqlite instr is 1 indexed
   startswith: (x, y) => `strpos(${x},${y}) = 1`, // sqlite instr is 1 indexed
-  endswith: (x, y) => `substr(${x},length(${x}) + 1 - length(${y})) = ${y}`,
+  endswith: (x, y) => `substr(${x},length(${x}) + 1 - length(${y})) = ${y}`,  
 
   // Date and Time Functions
-  year: x => `date_part('year',(${x})::TIMESTAMP)`,
-  month: x => `date_part('month',(${x})::TIMESTAMP)`,
-  day: x => `date_part('day',(${x})::TIMESTAMP)`,
-  hour: x => `date_part('hour',(${x})::TIMESTAMP)`,
-  minute: x => `date_part('minute',(${x})::TIMESTAMP)`,
-  second: x => `date_part('second',(${x})::TIMESTAMP)`,
+  year: x => {
+    if (isTime(x.val)) throw new cds.error ({
+      message: `Is time but date expected`,
+      code: 400,
+    })
+    return `date_part('year',${`${x}${!x.val ? '' : '::TIMESTAMP' }`})`
+  },
+  month: x => {
+    if (isTime(x.val)) throw new cds.error ({
+      message: `Is time but date expected`,
+      code: 400,
+    })
+    return `date_part('month',${`${x}${!x.val ? '' : '::TIMESTAMP' }`})`
+  },
+  day: x => {
+    if (isTime(x.val)) throw new cds.error ({
+      message: `Is time but date expected`,
+      code: 400,
+    })
+    return `date_part('day',${`${x}${!x.val ? '' : '::TIMESTAMP' }`})`
+  },
+  hour: x => {
+    if (isDate(x.val)) throw new cds.error ({
+      message: 'Is date but time expected',
+      code: 400,
+    })
+    return `date_part('hour',${`${x}${!x.val ? '' : (isTime(x.val) ? '::TIME' : '::TIMESTAMP') }`})`
+  },
+  minute: x => {
+    if (isDate(x.val)) throw new cds.error ({
+      message: 'Is date but time expected',
+      code: 400,
+    })
+    return `date_part('minute',${`${x}${!x.val ? '' : (isTime(x.val) ? '::TIME' : '::TIMESTAMP') }`})`
+  },
+  second: x => {
+    if (isDate(x.val)) throw new cds.error ({
+      message: 'Is date but time expected',
+      code: 400,
+    })
+    return `date_part('second',${`${x}${!x.val ? '' : (isTime(x.val) ? '::TIME' : '::TIMESTAMP') }`})`
+  },
 }
+
+function isTime(input) {
+  const timePattern = /^(?:\d{2}|\d{1}):(\d{2}|\d{1}):(\d{2}|\d{1})$/
+  return timePattern.test(input)
+}
+
+function isDate(input) {
+  const datePattern = /^(?:(\d{2}|\d{1})\/(\d{2}|\d{1})\/(\d{4})|(\d{2}|\d{1})\.(\d{2}|\d{1})\.(\d{4})|(\d{4})-(\d{2}|\d{1})-(\d{2}|\d{1}))$/
+  return datePattern.test(input)
+}
+
 
 const HANAFunctions = {
   // https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/f12b86a6284c4aeeb449e57eb5dd3ebd.html
--- a/postgres/lib/func.js
+++ b/postgres/lib/func.js
@@ -25,14 +25,12 @@ const StandardFunctions = {
 const getTimePart = (x, part) => {
   if ((part === 'hour' || part === 'minute' || part === 'second') && isDate(x.val)) {
     throw new cds.error({
-      message: 'Is date but time expected',
-      code: 400,
+      message: 'Is date but time expected'
     })
   }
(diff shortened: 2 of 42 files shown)


postgres/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Fixed
 
 - `cds build`-relevant files are now correctly packaged into the release. #266
+- The time functions are now correctly casted
```
patricebender, line 12:
```suggestion
```

no need to manually tweak the changelog, our release automation takes care of that

postgres/lib/func.js
```diff
@@ -10,17 +11,64 @@ const StandardFunctions = {
...
 
   // Date and Time Functions
-  year: x => `date_part('year',(${x})::TIMESTAMP)`,
-  month: x => `date_part('month',(${x})::TIMESTAMP)`,
-  day: x => `date_part('day',(${x})::TIMESTAMP)`,
-  hour: x => `date_part('hour',(${x})::TIMESTAMP)`,
-  minute: x => `date_part('minute',(${x})::TIMESTAMP)`,
-  second: x => `date_part('second',(${x})::TIMESTAMP)`,
+  year: x => {
+    if (isTime(x.val)) throw new cds.error ({
```
BobdenOs, line 18: To much duplicate code. Also DB service does not create errors with HTTP error codes.

BobdenOs (changes requested):

postgres/lib/func.js
```diff
@@ -10,17 +11,43 @@ const StandardFunctions = {
...
+  day: x => getTimePart(x, 'day'),
+  hour: x => getTimePart(x, 'hour'),
+  minute: x => getTimePart(x, 'minute'),
+  second: x => getTimePart(x, 'second'),
 }
 
+const getTimePart = (x, part) => {
+  if ((part === 'hour' || part === 'minute' || part === 'second') && isDate(x.val)) {
+    throw new cds.error({
+      message: 'Is date but time expected'
```
patricebender, line 28:
```suggestion
      message: 'Expected "time" but found "date"'
```

BobdenOs (changes requested):
With the optimized regex it is really no longer needed to wrap them inside functions. Also as each function call would create a new `RegExp` Instance.

test/scenarios/bookshop/funcs.test.js
```diff
@@ -155,6 +155,82 @@ describe('Bookshop - Functions', () => {
...
+    const testCases = [
+      { value: '1970-03-30T01:01:01.000Z', expectedResultPostgres: 5, expectedResult: 5 },
+      { value: 'foo', expectedResultPostgres: `invalid input syntax for type timestamp: "foo"`, expectedResult: 0, },
+    ]
+
+    const testCasesYearMonthDay = [
+      { value: '1970-03-30', expectedResultPostgres: 5, expectedResult: 5 },
+      { value: '03/30/1970', expectedResultPostgres: 5, expectedResult: 0 },
+      { value: '03.30.1970', expectedResultPostgres: 5, expectedResult: 0 },
+      { value: '01:01:01', expectedResultPostgres: `Is time but date expected`, expectedResult: 0 },
```
BobdenOs, line 168: `expectedResultPostgres` is against the fundamental purpose of having the new db service. Where everything behaves the same. The error message might differ, but just don't expect them to exactly match. Just expect them to fail or succeed.

postgres/lib/func.js
larslutz96: Thanks for the hint.

larslutz96:
const testCases = [
      { value: '1970-03-30T01:01:01.000Z', expectReqToSucceed: true },
      { value: 'foo', expectReqToSucceed: false }, // true without postgres
    ]

    const testCasesYearMonthDay = [
      { value: '1970-03-30', expectReqToSucceed: true },
      { value: '03/30/1970', expectReqToSucceed: true },
      { value: '03.30.1970', expectReqToSucceed: true },
      { value: '01:01:01', expectReqToSucceed: false }, // true without postgres
      { value: '1:1:1', expectReqToSucceed: false },  // true without postgres
    ]

    const testCasesHourMinuteSecond = [
      { value: '1970-03-30', expectReqToSucceed: false },
      { value: '03/30/1970', expectReqToSucceed: true },
      { value: '03.30.1970', expectReqToSucceed: true },
      { value: '01:01:01', expectReqToSucceed: true },
      { value: '1:1:1', expectReqToSucceed: true },
    ]

    test.each([...testCases, ...testCasesYearMonthDay])('where day %s', async ({ value, expectReqToSucceed }) => {
      let reqSuccessfull = true
      try {
        await SELECT.from('sap_capire_bookshop_Books').where `day(${value}) = 30`
      } catch (e) {
        reqSuccessfull = false
      }
      expect(reqSuccessfull).to.be.eq(expectReqToSucceed)
    })
    
    for the testCase with the value "foo" i get back the value true when i send the request with the db layer sqlite. When i send it with postgres i get back the value false. How can i improve this test so i can only use just one expectedReqToSucceed value

test/scenarios/bookshop/funcs.test.js
larslutz96: But for example: when i send `where day('foo') = 30` with postgres the request runs into the catch block. When i send it without postgres i just get back an empty array (request succeeded). How can i check without checking on the db.

BobdenOs: The problem is that the databases respond differently. So rather then changing the expectation in the test. It is required to adjust the behavior to be aligned.

johannes-vogel:
Is there still something open or is it ready for review?

BobdenOs:
I think this PR is ready for review. All database services should now throw errors when the wrong combination of data/time functions are being used. It might require some more extensive test suites to cover all scenarios, but this state should cover the original task.

patricebender:
@larslutz96 please no internal links in public github, thanks :)

sqlite/lib/SQLiteService.js
```diff
@@ -15,6 +15,19 @@ class SQLiteService extends SQLService {
         dbc.function('session_context', key => dbc[$session][key])
         dbc.function('regexp', { deterministic: true }, (re, x) => (RegExp(re).test(x) ? 1 : 0))
         dbc.function('ISO', { deterministic: true }, d => d && new Date(d).toISOString())
+
+        // define date and time functions in js to allow for throwing errors
+        const isTime = /^\d{1,2}:\d{1,2}:\d{1,2}$/
+        const hasTimezone = /([+-]\d{1,2}:?\d{0,2}|Z)$/
+        const throwNaN = n => { if (Number.isNaN(n)) { throw new Error(`Value does not contain a date "${n}"`) } else { return n } }
+        const toDate = d => new Date(isTime.test(d) ? `1970-01-01T${d}Z` : hasTimezone.test(d) ? d : d + 'Z')
+        dbc.function('year', { deterministic: true }, d => d && throwNaN(new Date(d).getUTCFullYear()))
```
johannes-vogel, line 24: What happens for falsy values like empty string?

sqlite/lib/SQLiteService.js
```diff
@@ -15,6 +15,19 @@ class SQLiteService extends SQLService {
         dbc.function('session_context', key => dbc[$session][key])
         dbc.function('regexp', { deterministic: true }, (re, x) => (RegExp(re).test(x) ? 1 : 0))
         dbc.function('ISO', { deterministic: true }, d => d && new Date(d).toISOString())
+
+        // define date and time functions in js to allow for throwing errors
+        const isTime = /^\d{1,2}:\d{1,2}:\d{1,2}$/
+        const hasTimezone = /([+-]\d{1,2}:?\d{0,2}|Z)$/
+        const throwNaN = n => { if (Number.isNaN(n)) { throw new Error(`Value does not contain a date "${n}"`) } else { return n } }
```
johannes-vogel, line 22: I'd expect  from the function name that this function always throws.
Does it make sense to extract the `isNaN` check?

sqlite/lib/SQLiteService.js
```diff
@@ -15,6 +15,19 @@ class SQLiteService extends SQLService {
...
+        const toDate = d => new Date(isTime.test(d) ? `1970-01-01T${d}Z` : hasTimezone.test(d) ? d : d + 'Z')
+        dbc.function('year', { deterministic: true }, d => d && throwNaN(new Date(d).getUTCFullYear()))
+        dbc.function('month', { deterministic: true }, d => d && throwNaN(new Date(d).getUTCMonth()) + 1)
```
johannes-vogel, line 25: What happens for invalid date values? Doesn't the new Date constructor reject that?

postgres/lib/func.js
```diff
@@ -13,14 +13,18 @@ const StandardFunctions = {
...
+  month: x => `date_part('month', ${castVal(x)})`,
+  day: x => `date_part('day', ${castVal(x)})`,
+  hour: x => `date_part('hour', ${castVal(x)})`,
+  minute: x => `date_part('minute', ${castVal(x)})`,
+  second: x => `date_part('second', ${castVal(x)})`,
 }
 
+const isTime = /^\d{1,2}:\d{1,2}:\d{1,2}$/
+const isVal = x => x && 'val' in x
+const castVal = (x) => `${x}${isVal(x) ? isTime.test(x.val) ? '::TIME' : '::TIMESTAMP' : ''}`
```
johannes-vogel, line 26: Is there a reason to only cast vals? hour(deliveryDate) is also allowed, isn't it?

johannes-vogel (commented):
Please also pick an appropiate PR title. This does not only fix time functions on postgres. It also rejects certain previously supported queries on sqlite.

sqlite/lib/SQLiteService.js
BobdenOs: If it was not clear from the function name I would rename the function to `throwsIfNaN`.

sqlite/lib/SQLiteService.js
BobdenOs: Quick overview of the `new Date` behavior.
```js
new Date('') // Invalid
new Date(0) // 1970
new Date(null) // 1970
new Date(false) // 1970
new Date('0') // 2000
```

sqlite/lib/SQLiteService.js
BobdenOs: `throwNaN` is what make Invalid Dates throw. Other wise it will just become `NaN`.

postgres/lib/func.js
BobdenOs: Only `val` can have no types. `hour(deliveryDate)` will have a type already.

sqlite/lib/SQLiteService.js
johannes-vogel, line 22: I don't think that this gives a good error message. If I provide an invalid time or date, it'll end up in an error message `Value does not contain a date "NaN"`. it's not clear whether the user input was a time or what was the invalid value.

johannes-vogel: I'd prefer to rename it.

patricebender:
> Please also pick an appropiate PR title. This does not only fix time functions on postgres. It also rejects certain previously supported queries on sqlite.

You can add multiple conventional commits to the pr description:

```
Fix: time function casting

The time functions are now correctly casted.

fix: add validations and reject certain previously supported queries.
  … explain why this is a fix and not a breaking change
```

sqlite/lib/SQLiteService.js
larslutz96: So we should differentiate between year, month, day (which require a date) and hour, minute second (which require a time)? @BobdenOs do you agree?

BobdenOs: The error does not need to be related to the function trying to parse the value as the value is only `NaN` if the original value is an invalid `Date`. So even before it would be possible to apply the extraction it already failed.

johannes-vogel: Yes, but now we have the possibility to raise a better error message, don't we?

johannes-vogel (approved):

PR #323, opened 2023-10-27
feat: use place holders for update and delete

BobdenOs:
# Details

Before all `val` values where printed into the `SQL` statement. To improve overall performance this is not longer done. Which allows common `DELETE` and `UPDATE` statements to be compiled only once. Additionally the statements are cached on the connection reducing the network usage.

- side effect of the statement cache is that `@sap/hana-client` cannot run the same statement multiple times at the same time. So the `deep-queries` logic is slightly adjusted to not generate the same `INSERT` queries multiple times, but instead to send them all as a single `INSERT` statement. As all `INSERT` statements to the same entity with `entries` will produce the same `SQL` statements.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -64,12 +64,11 @@ class HANAService extends SQLService {
   }
 
   async set(variables) {
-    const columns = Object.keys(variables).map(
-      k => `SET '${k.replace(/'/g, "''")}'='${(variables[k] + '').replace(/'/g, "''")}';`,
-    )
-    const sql = `DO BEGIN ${columns.join('')} END;`
+    // REVISIT: required to be compatible with generated views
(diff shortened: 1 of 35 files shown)


hana/lib/HANAService.js
```diff
@@ -528,6 +530,7 @@ class HANAService extends SQLService {
     }
 
     INSERT_entries(q) {
+      this.values = undefined
```
johannes-vogel, line 533: In which situations do we need to reset the values?

BobdenOs: By setting the `values` to `undefined` it informs the `val` function that it should just inline the values instead of pushing them onto the `values` array. When the `INSERT` statement is being rendered it is important to only have the `JSON` payload be the parameters. This also prevents `session_context` and other functions from being rendered with place holders while their values should be rendered into the sql directly instead.

johannes-vogel (approved):

PR #324, opened 2023-10-27
feat: use a simple select for flat queries

BobdenOs:
# Details

Without this change every query is wrapped inside a `DO BEGIN` which allows for multiple queries to be executed, but this comes with a bit of an overhead. So now flat queries (without expands) are now done as a `SELECT` directly.

Additionally the `_json_` column is returned as an `NVARCHAR` as the HANA network protocol for `NCLOB` requires additional `FETCH` calls which cause a dramatic overal request response time.


johannes-vogel (approved):

johannes-vogel:
@BobdenOs Is there something you want to do within this PR or should we merge?

PR #325, opened 2023-10-27
feat(`UPDATE`/`DELETE`): Enable path expressions for improved data manipulation

### LINKED ISSUE #159: @cap-js/sqlite: nested properties in where clause of delete crash

patricebender:
this change enables the usage of path expressions in `where` clauses of `UPDATE` and `DELETE` queries.

If there is a path expressions found in a `where` of an `UPDATE` or `DELETE`, the `where` clause is cut out of the query and is put into a `where` clause  in a dummy subselect. This subquery, selectes all primary keys. It is then transformed by `cqn4sql`. The result of the transformation is then put into the main queries `where` again, as `<list of primary key> in <subquery>`.

---

fixes #159

### DIFF
--- a/db-service/test/cqn4sql/DELETE.test.js
+++ b/db-service/test/cqn4sql/DELETE.test.js
@@ -77,7 +77,7 @@ describe('DELETE', () => {
 
     // this is the final exists subquery
     const subquery = CQL`
-     SELECT 1 from bookshop.Authors as author
+     SELECT author.ID from bookshop.Authors as author
       left join bookshop.Books as books on books.author_ID = author.ID
      where exists (
       SELECT 1 from bookshop.Books as Books2 where Books2.author_ID = author.ID
@@ -89,11 +89,21 @@ describe('DELETE', () => {
             "ref": [
               "bookshop.Authors"
             ],
-            "as": "author"
+            "as": "author2"
           }
         }
       }`)
-    expected.DELETE.where = ['exists', subquery]
+    expected.DELETE.where = [
+      {
+        list: [
+          {
+            ref: ['author2', 'ID'],
+          },
+        ],
+      },
+      'in',
+      subquery,
+    ]
     expect(query.DELETE).to.deep.equal(expected.DELETE)
   })
 
--- a/db-service/test/cqn4sql/DELETE.test.js
+++ b/db-service/test/cqn4sql/DELETE.test.js
@@ -69,6 +69,33 @@ describe('DELETE', () => {
       }`)
     expect(query.DELETE).to.deep.equal(expected.DELETE)
   })
+  it('DELETE with where exists expansion and path expression', () => {
+    cds.model = cds.compile.for.nodejs(cds.model)
+    const { DELETE } = cds.ql
+    let d = DELETE.from('bookshop.Books:author').where(`books.title = 'Harry Potter'`)
+    const query = cqn4sql(d)
+
+    // this is the final exists subquery
+    const subquery = CQL`
+     SELECT 1 from bookshop.Authors as author
+      left join bookshop.Books as books on books.author_ID = author.ID
+     where exists (
+      SELECT 1 from bookshop.Books as Books2 where Books2.author_ID = author.ID
+     ) and books.title = 'Harry Potter'
+    `
+    const expected = JSON.parse(`{
+      "DELETE": {
+          "from": {
+            "ref": [
+              "bookshop.Authors"
+            ],
+            "as": "author"
+          }
+        }
+      }`)
+    expected.DELETE.where = ['exists', subquery]
+    expect(query.DELETE).to.deep.equal(expected.DELETE)
+  })
 
   it('DELETE with assoc filter and where exists expansion', () => {
     const { DELETE } = cds.ql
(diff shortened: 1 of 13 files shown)


stewsk (approved):

johannes-vogel (approved):

db-service/test/cqn4sql/UPDATE.test.js
```diff
@@ -70,17 +70,52 @@ describe('UPDATE', () => {
...
+      exists (SELECT 1 from bookshop.Books as Books
+              where author.name LIKE '%Bron%' or ( author.name LIKE '%King' and title = 'The Dark Tower') and stock >= 15
+            )
+    `)
+    let expected = UPDATE.entity({ ref: ['bookshop.Books'] }).where(`
+    exists (SELECT 1 from bookshop.Books as Books
+              left join bookshop.Authors as author on author.ID = Books.author_ID
+              where author.name LIKE '%Bron%' or ( author.name LIKE '%King' and Books.title = 'The Dark Tower') and Books.stock >= 15
+            )
+  `)
```
patricebender, line 89: I did some manual, end-to-end testing with delete statements. It seems that an important peace is missing. Without the correlation to the outer query (`UPDATE` in this case), all entries will be matched and hence updated :(

BobdenOs: My expectation would be something like this
```suggestion
    let expected = UPDATE.entity({ ref: ['bookshop.Books'] }).where(`
    ID IN (SELECT ID from bookshop.Books as Books
              left join bookshop.Authors as author on author.ID = Books.author_ID
              where author.name LIKE '%Bron%' or ( author.name LIKE '%King' and Books.title = 'The Dark Tower') and Books.stock >= 15
            )
  `)
```

patricebender: cool idea, locally, I solved it by injecting the correlation into the `where` of the subquery

patricebender: > Please also add a test for updating a projection. It's important that the SELECT is on the projection while the UPDATE's projection must be resolved to the table below.

what exactly does it mean? Does `cqn4sql` then learn what `resolveView` currently does, also w.r.t. renamings?

johannes-vogel: No, that would simply cover the end-to-end scenario to work. Should be a test outside cqn4sql.

test/compliance/DELETE.test.js
```diff
@@ -1,13 +1,28 @@
...
+    test.skip('path expressions', async () => {
+      const deleteEmilysBooks = DELETE.from('AdminService.RenameKeys').where(`author.name = 'Emily Brontë'`)
+      const selectEmilysBooks = CQL`SELECT * FROM sap.capire.bookshop.Books where author.name = 'Emily Brontë'`
+
+      const beforeDelete = await cds.run(selectEmilysBooks)
+      await cds.run(deleteEmilysBooks)
+      const afterDelete = await cds.run(selectEmilysBooks)
+
+      expect(beforeDelete).toHaveLength(1)
+      expect(afterDelete).toHaveLength(0)
```
patricebender, line 25: added end2end test which works locally with the `resolveView` fix in the dev setup → skipped in this test suite for now as we need to wait for the next `cds` release.

johannes-vogel: Can you please already prepare a PR to enable them? Just that we do not forget.

db-service/test/cqn4sql/DELETE.test.js
```diff
@@ -69,80 +69,101 @@ describe('DELETE', () => {
...
                 {
-                  "val": 99
-                }
-              ]
-            }
-          }
-        ]
-      }
+                  val: 99,
+                },
```
johannes-vogel, line 161: Why are there so many trailing commas?

patricebender: Because of this:
https://github.com/cap-js/cds-dbs/blob/7896289ec97f28bdc2c1d7700af1953608c96ef2/.prettierrc.js#L16

johannes-vogel (approved):

test/compliance/DELETE.test.js
patricebender: #329

patricebender:
please let me merge my PRs, I had some more cosmetic changes to do :(

johannes-vogel:
> please let me merge my PRs, I had some more cosmetic changes to do :(

sry, I thought that's ready to merge.
Usually, I use the _draft PRs_ until they are ready to merge. Maybe we can do that as a general rule?

PR #327, opened 2023-10-30
fix: Disconnect db service on shutdown

BobdenOs:
Currently with `HANA` and `Postgres` the service does not disconnect with the default `cds.test`. So this registers the `disconnect` function to the `shutdown` event.

### DIFF
--- a/db-service/lib/common/DatabaseService.js
+++ b/db-service/lib/common/DatabaseService.js
@@ -118,7 +118,11 @@ class DatabaseService extends cds.Service {
    */
   async disconnect(tenant) {
     const pool = this.pools[tenant]
-    if (!pool) return
+    if (!pool) {
+      if (tenant == null)
+        return Promise.all(Object.keys(this.pools).map(t => this.disconnect(t)))
+      return
+    }
     await pool.drain()
     await pool.clear()
     delete this.pools[tenant]
--- a/db-service/lib/common/DatabaseService.js
+++ b/db-service/lib/common/DatabaseService.js
@@ -15,7 +15,7 @@ class DatabaseService extends cds.Service {
   /**
    * Dictionary of connection pools per tenant
    */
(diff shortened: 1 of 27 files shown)


patricebender (approved):

db-service/lib/common/DatabaseService.js
```diff
@@ -6,6 +6,12 @@ const cds = require('@sap/cds/lib')
 /** @typedef {unknown} DatabaseDriver */
 
 class DatabaseService extends cds.Service {
+
+  init() {
+    cds.on('shutdown', () => this.disconnect())
```
johannes-vogel, line 11: We might need to implement something that disconnects all tenant pools in multitenancy mode. Something like `disconnect('*')`

BobdenOs: `disconnect(undefined)` should match all tenants. Will update the `disconnect` function to just go over all pools.

db-service/lib/common/DatabaseService.js
```diff
@@ -111,11 +117,18 @@ class DatabaseService extends cds.Service {
...
+      if (!pool) {
+        return
+      }
+      await pool.drain()
+      await pool.clear()
+      delete this.pools[tenant]
+    }
+    if (tenant == null)
+      return Promise.all(Object.keys(this.pools).map(_disconnect))
+    return _disconnect(tenant)
```
johannes-vogel, line 131:
```suggestion
      if (tenant == null)
        return Promise.all(Object.keys(this.pools).map(this.disconnect))
      const pool = this.pools[tenant]
      if (!pool) return
      await pool.drain()
      await pool.clear()
      delete this.pools[tenant]
    }
```

BobdenOs: It might be more understandable, but is also an infinite loop.

PR #331, opened 2023-11-02
chore: reorganize compliance tests to always use same models

johannes-vogel:
@BobdenOs Do you see the problem with the hana/pg tests? I feel that views are not supported in CREATE.test.js? Should we simply skip to unblock us internally?

### DIFF
--- a/test/compliance/DELETE.test.js
+++ b/test/compliance/DELETE.test.js
@@ -12,16 +12,8 @@ describe('DELETE', () => {
   })
 
   describe('where', () => {
-    test.skip('path expressions', async () => {
-      const deleteEmilysBooks = DELETE.from('complex.RenameKeys').where(`author.name = 'Emily'`)
-      const selectEmilysBooks = CQL`SELECT * FROM complex.Books where author.name = 'Emily'`
-
-      const beforeDelete = await cds.run(selectEmilysBooks)
-      await cds.run(deleteEmilysBooks)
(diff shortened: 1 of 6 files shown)


test/compliance/DELETE.test.js
```diff
@@ -3,7 +3,7 @@
 const cds = require('../cds.js')
 
 describe('DELETE', () => {
-  cds.test(__dirname + '/../bookshop')
+  cds.test(__dirname + '/resources')
```
patricebender, line 6: there is a similar test in `UPDATE.test.js`

test/compliance/resources/db/complex/index.cds
```diff
@@ -11,3 +11,10 @@ entity Authors {
   name   : String(111);
   books  : Association to many Books on books.author = $self;
 }
+
```
patricebender, line 14: we could clean up the `RenameKeys` projection in the bookshop then

test/compliance/DELETE.test.js
johannes-vogel: I don't see any :)

patricebender: oops, thought https://github.com/cap-js/cds-dbs/pull/329/files was already merged

patricebender: i will change it in the PR once this is in main :)

BobdenOs (approved):

PR #335, opened 2023-11-09
chore: enable skipped tests

patricebender:
if a virtual column is the only column in a given clause, we strip the clause from the query as the virtual column doesnt materialize.


db-service/test/cds-infer/elements.test.js
```diff
@@ -12,7 +12,7 @@ describe('infer elements', () => {
   })
 
   describe('path expressions', () => {
-    it('along simple association', () => {
+    it.only('along simple association', () => {
```
patricebender, line 15:
```suggestion
    it('along simple association', () => {
```

PR #341, opened 2023-11-10
test: correlated subquery vs. inner joins

patricebender:


patricebender:
works as designed, the inner join which is generated in the cds-compiler could be aligned with the runtime behavior

PR #343, opened 2023-11-15
feat: Support Readable Streams inside INSERT.entries

BobdenOs:
It should always be possible to `INSERT` the results of a `SELECT` statement. Which was no longer possible with the new approach of streaming `cds.LargeBinary`. So this implementation allows for the streams to be merged back into a single `JSON` stream.

```js
await (INSERT || UPSERT)(await SELECT.from('Books.drafts')).into('Books')
```


vkozyura (approved):
:star::star::star:

johannes-vogel (approved):

PR #345, opened 2023-11-17
chore: remove obsolete comment

patricebender:
as requested by Oliver K.

### DIFF
--- a/db-service/lib/deep-queries.js
+++ b/db-service/lib/deep-queries.js
@@ -35,7 +35,7 @@ async function onDeep(req, next) {
     if (query.UPDATE) return this.onUPDATE({ query })
     if (query.DELETE) return this.onSIMPLE({ query })
   }))
-  return res[0] ?? 0 // TODO what todo with multiple result responses?
+  return res[0] ?? 0
 }
 
 const hasDeep = (q, target) => {


BobdenOs:
Honestly I would just leave it for what it is right now. As it would be removed once the `deep-queries.js` logic is re worked.

Also it seems that this might actually be a bug with the current implementation if the root entity was not being modified.
As the answer to the question is that we should always return the number of root entities being modified.
Where if the current implementation does not find any changes on the root entity it would not add an `UPDATE` query for the root entity and it would return the number of modification in the child queries rather then the root entries. My expectation is that currently it will mostly return `1` or `0` so it doesn't show that it is not very obvious that it is returning the wrong value.

patricebender:
> Honestly I would just leave it for what it is right now. As it would be removed once the `deep-queries.js` logic is re worked.
> 
> Also it seems that this might actually be a bug with the current implementation if the root entity was not being modified. As the answer to the question is that we should always return the number of root entities being modified. Where if the current implementation does not find any changes on the root entity it would not add an `UPDATE` query for the root entity and it would return the number of modification in the child queries rather then the root entries. My expectation is that currently it will mostly return `1` or `0` so it doesn't show that it is not very obvious that it is returning the wrong value.

makes sense. In this case, the comment is not really obsolete at this point and should be left as is. Thanks for explaining.

PR #348, opened 2023-11-20
fix: apply schema to hdb connection

BobdenOs:


johannes-vogel (approved):

johannes-vogel:
@sap/hana-client sets the schema by default already?

BobdenOs:
`@sap/hana-client` uses `currentSchema` as part of the credentials [here](https://github.com/cap-js/cds-dbs/blob/main/hana/lib/drivers/hana-client.js#L16)

PR #352, opened 2023-11-21
fix: wrong odata count in filter with groupby

larslutz96:
Encase the groupby within an additional SELECT statement to ensure accurate row counting.


db-service/lib/SQLService.js
```diff
@@ -236,6 +236,18 @@ class SQLService extends DatabaseService {
       expand: false,
       limit: 0,
       orderBy: 0,
+      groupBy: 0,
+      ...(query.SELECT.groupBy
+        ? {
+            from: {
+              SELECT: {
+                columns: [{ ref: [query.SELECT.groupBy[0].ref[0]] }],
```
BobdenOs, line 244: This does not look dynamic enough. A GroupBy can have more then 1 column. And the ref can have more then 1 path entry.

johannes-vogel: do we need to specify any column at all?
isn't a select 1 sufficient?

I'd guess that it's sufficient if we ignore any limit clause in the sub select as well.

test/scenarios/sflight/read.test.js
```diff
@@ -44,6 +44,7 @@ describe('SFlight - Read', () => {
     // REVISIT: works in sflight not in tests
     // 'Bookings?$apply=concat(groupby((BookingID,ConnectionID,CurrencyCode_code,FlightDate,ID,TravelID,airline,status))/aggregate($count%20as%20UI5__leaves),aggregate(FlightPrice,CurrencyCode_code),groupby((airline,airlineName),aggregate(FlightPrice,CurrencyCode_code))/concat(aggregate($count%20as%20UI5__count),top(53)))',
     'Bookings?$apply=groupby((airline,airlineName),aggregate(FlightPrice with average as avgPrice,FlightPrice with max as maxPrice,FlightPrice with min as minPrice))&$skip=0&$top=1',
+    `Bookings?$apply=filter(airline eq 'EA' and status eq 'B')/groupby((ID),aggregate(FlightPrice,CurrencyCode_code))&$count=true&$skip=0&$top=1`
```
johannes-vogel, line 47: we don't really test that the count is correct.
The request went throught before as well but included a wrong count

db-service/lib/SQLService.js
BobdenOs: Wouldn't it make more sense to just wrap the original query (minus limiting factors) with a count.
```cds
SELECT count() FROM (query)
```

test/scenarios/sflight/read.test.js
BobdenOs: Also having a more simplified query that reproduces this issue would make it easier to understand the problem later when someone breaks the test.

db-service/lib/SQLService.js
johannes-vogel: Yes, I think that would be best.

test/scenarios/bookshop/read.test.js
```diff
@@ -30,6 +30,15 @@ describe('Bookshop - Read', () => {
...
   })
 
+  test('Books $count with $top=2 and filter', async () => {
+    const res = await GET(
+      `/browse/ListOfBooks?$apply=filter(currency_code eq 'GBP')/groupby((ID),aggregate(ID with countdistinct as countBookings))&$count=true&$skip=0&$top=2`,
+    )
+    expect(res.status).to.be.eq(200)
+    expect(res.data.value.length).to.be.eq(2)
+    expect(res.data['@odata.count']).to.be.eq(2)
+  })
```
BobdenOs, line 40:
```suggestion
  test('Books $count with $top=0 and groupby', async () => {
    const res = await GET(
      `/browse/ListOfBooks?$apply=groupby((ID),aggregate(ID with countdistinct as countBookings))&$count=true&$top=2`,
    )
    expect(res.status).to.be.eq(200)
    expect(res.data.value.length).to.be.eq(0)
    expect(res.data['@odata.count']).to.be.eq(5)
  })
```

This change is trying to make the most minimal reproducable scenario. The problem was not the filter it is the `groupby` and the `top` being less then the count.

larslutz96: test updated as suggested

stewsk:
Hi @johannes-vogel , would this fix the issue with the Analytical List Page in SFlight?

johannes-vogel:
@stewsk Yes, that's the PR for it.

BobdenOs (approved):

johannes-vogel (approved):

PR #355, opened 2023-11-23
fix: large amounts of expands in HANA Service

BobdenOs:


johannes-vogel (approved):
I suppose we merge this as well to have all implementation in main

PR #358, opened 2023-11-24
ci: prefer SHA over tag for github action

patricebender:
Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. 

see [github docs](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

---

[this commit](https://github.com/amannn/action-semantic-pull-request/commit/e9fabac35e210fea40ca5b14c0da95a099eff26f) is now used instead of the tag

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -15,6 +15,7 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
+      # v5.4.0
       - uses: amannn/action-semantic-pull-request@e9fabac35e210fea40ca5b14c0da95a099eff26f
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/conventional-commits.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5
+      - uses: amannn/action-semantic-pull-request@e9fabac35e210fea40ca5b14c0da95a099eff26f
```
bugwelle, line 18:
```suggestion
      # v5.4.0
      - uses: amannn/action-semantic-pull-request@e9fabac35e210fea40ca5b14c0da95a099eff26f
```

bugwelle (approved):

PR #360, opened 2023-11-25
ci: Update actions/checkout and actions/setup-node to v4

bugwelle:
See

 - https://github.com/actions/checkout/releases/tag/v4.0.0
 - https://github.com/actions/setup-node/releases/tag/v4.0.0

Major version change due to internal upgrade from Node 16 to 20.


patricebender (approved):
Thanks :)

PR #363, opened 2023-11-29
ci: include @cap-js/hana in release pipeline

johannes-vogel:
### TODO
- [ ] include hana job in release pipeline
- [ ] is there something else missing?
- [x] to be done before release
  - [x] pooling
  - [x] mt support
- [ ] compile list of missing features
  - [ ] for update/share lock
  - [ ] only plain mode supported
  - [ ] experimental hana-client/hdb config
  - [ ] timestamp normalization - to double check
  - [ ] tx protection for parallel before handlers

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -3,7 +3,8 @@
   "packages": {
     "db-service": {},
     "sqlite": {},
-    "postgres": {}
+    "postgres": {},
+    "hana": {}
   },
   "changelog-sections": [
     {
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -4,7 +4,9 @@
     "db-service": {},
     "sqlite": {},
     "postgres": {},
-    "hana": {}
+    "hana": {
+      "release-as": "0.0.1"
+    }
   },
   "changelog-sections": [
     {
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -50,3 +50,9 @@ jobs:
         run: npm publish --workspace postgres --access public
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
+
+      - name: Publish SAP HANA
+        if: ${{ steps.release.outputs.releases_created && steps.release.outputs.hana--release_created }}
+        run: npm publish --workspace hana --access public
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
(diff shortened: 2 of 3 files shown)


patricebender (commented):
we need to configure the [release-please-manifest.json](https://github.com/cap-js/cds-dbs/blob/main/.release-please-manifest.json) as well. Also, if we do not want to include every commit we made to the `hana` scope in the initial changelog, we could change the [bootstrap-sha](https://github.com/cap-js/cds-dbs/blob/70690a1a13e72bfbb66f03bf315d3f2d48672bf6/release-please-config.json#L2C1-L2C1) in the release-please-config.json.

patricebender:
> > we need to configure the [release-please-manifest.json](https://github.com/cap-js/cds-dbs/blob/main/.release-please-manifest.json) as well.
> 
> Which version do we need to choose here? 0.0.1?

It depends, with which version we want to start. Should the first version be `1.0.0`?

we should configure the initial release of hana in a dedicated section in the release-please-config.json:
```
  "packages": {
    // path segment should be relative to repository root
    "path/to/hana": {
      "release-as": "1.0.0"
      "bootstrap-sha": "XXXX",
    },
  } 
```
With the `release-as` property, we can force release please to use the given version for the next release in the respective scope. We should clean the whole section up once the first release is out.

This is documented in the [release-please docs](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md)

> > Also, if we do not want to include every commit we made to the `hana` scope in the initial changelog, we could change the [bootstrap-sha](https://github.com/cap-js/cds-dbs/blob/70690a1a13e72bfbb66f03bf315d3f2d48672bf6/release-please-config.json#L2C1-L2C1) in the release-please-config.json.
> 
> I think we should do so as the initial release does not include any changelog.

alright, then as described above we can use `release-as` and `bootstrap-sha` to tweak our inital release accordingly. However, the easiest way would be to have one dummy commit to the `hana` scope which can be picked up by `release-please`. We can then manually delete the auto-generated section in the changelog for the inital version. :)

release-please-config.json
```diff
@@ -3,7 +3,8 @@
   "packages": {
     "db-service": {},
     "sqlite": {},
-    "postgres": {}
+    "postgres": {},
+    "hana": {}
```
johannes-vogel, line 7:
```suggestion
    "hana": {
      "release-as": "0.0.1"
    }
```

johannes-vogel:
https://github.com/cap-js/cds-dbs/blob/f70ddd54278cf3e959ea17cfdca50646fba9cf04/.github/workflows/release-please.yml#L31-L35

@BobdenOs @patricebender Can we simply add `-w hana` to run hana tests as well or do we need additional things?

patricebender (approved):

PR #365, opened 2023-12-01
chore: Allow testing with a shared HANA Cloud instance

BobdenOs:
## Changes

Currently it is not possible to connect with the tests to a HANA Cloud instance. This is related to the reduced privileges that `DB_ADMIN` has. The `SYSTEM` user had enough privileges to run the previous script, but with HANA Cloud instances the `SYSTEM` user is owned by the HANA Cloud Dev Ops and therefor the end user will only have access to the `DB_ADMIN` user.

Which means that it is required to use the `USERGROUP` API to create isolated `USER ADMIN` users that can only create and modify users in their own `USERGROUP`.

Additionally the `test/cds.js` script now checks the current `username` before falling back to `test_db` as this value will be hashed and used by the `database` function to create a new `HDI CONTAINER GROUP` isolation. Which requires it to be different for each user or the isolation layers will have name clashes defeating the purpose of having isolation. 

## HANA 2.0 vs HANA Cloud 4.0

Depending on the HANA version used the `OPERATOR` role has a different name. HANA Cloud uses `OPERATOR` ([docs](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-sql-reference-guide/grant-statement-access-control)) while HANA 2.0 uses `USERGROUP OPERATOR` ([docs](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20f674e1751910148a8b990d33efbdc5.html))


danjoa (commented):
Trust it's ok :)

johannes-vogel (approved):

PR #369, opened 2023-12-04
chore: align hana plugin

johannes-vogel:


hana/cds-plugin.js
```diff
@@ -1,6 +1,9 @@
 const cds = require('@sap/cds/lib')
 
+if (!cds.env.fiori.lean_draft) {
+  throw new Error('"@cap-js/hana" only works if cds.fiori.lean_draft is enabled. Please adapt your configuration.')
+}
+
```
johannes-vogel, line 6: Why do we need to set `cds.env.sql.dialect = 'hana'`?

hana/package.json
```diff
@@ -36,9 +36,6 @@
     "requires": {
       "kinds": {
         "sql": {
-          "[development]": {
-            "kind": "sqlite"
```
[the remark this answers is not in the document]
johannes-vogel: I've added it now explicitly.

hana/cds-plugin.js
johannes-vogel: Then this is basically "only" required for our tests which generate the plain "CREATE" statements as of now. In production, the deployment is anyway done by HDI?

johannes-vogel: Yes, but quoted mode is not yet supported, is it?

BobdenOs (approved):

PR #374, opened 2023-12-06
fix: typo in error message

schwma:


patricebender (approved):
Thanks :) 
→ Changed from `fix` to `chore`, no need to mention in changelog.

PR #383, opened 2023-12-13
feat: Compress HANA expand queries by reducing duplicated statements

BobdenOs:
## Concept

When constructing an `expand` query it is required to reference the parent query in the child query. This was originally prevented by storing the results into a table value, but when switching away from the `DO BEGIN` syntax the queries where being duplicated into each child query. This change attempts to remove the query duplication by defining them inside the `WITH` clause which allows the child queries to directly reference the parent query. This _SHOULD_ allow for the SQL parser to parse less and _SHOULD_ allow for the execution plan optimizer to easier recognize that the query is doing the same thing. The big danger is that the execution plan recognizes that the query is doing the same thing and results in the queries executing in sequence. Which might result in reverting the behavior back to before https://github.com/cap-js/cds-dbs/pull/342. So the hope is that this makes everything much better and faster. 🤞

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -24,6 +24,7 @@ db-service/test/tsc/**.js @BobdenOs @danjoa @johannes-vogel @stewsk
 db-service/test/index.js @BobdenOs @danjoa @johannes-vogel @stewsk
 
 sqlite/ @BobdenOs @patricebender @danjoa @johannes-vogel @stewsk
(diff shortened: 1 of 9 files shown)


johannes-vogel:
@BobdenOs I've tried to resolve the conflicts but I don't know whether the failing tests are a result of a bad conflict resolution. Could you double check? At least the commit history tells me that this was anyway not yet working?!

johannes-vogel (approved):

PR #384, opened 2023-12-13
feat: strict mode to validate input for `INSERT`, `UPDATE` and `UPSERT`

larslutz96:
Add a strict mode to validate input for INSERT, UPDATE and UPSERT.
This "strict" mode is a configuration that can be turned on, e. g. cds.env.features.db_strict.


db-service/lib/SQLService.js
```diff
@@ -20,6 +20,33 @@ class SQLService extends DatabaseService {
...
     this.on(['INSERT', 'UPSERT', 'UPDATE'], require('./deep-queries').onDeep)
+    if (cds.env.features.db_strict) {
+      this.on(['INSERT', 'UPSERT', 'UPDATE'], req => {
+        const query = req.query
+        const entries = (query.INSERT || query.UPSERT)?.entries
+        const missingColumns = entries
+          ? entries.flatMap(entry => Object.keys(entry).filter(el => !Object.keys(query.target.elements).includes(el)))
+          : (query.INSERT?.columns || Object.keys(query.UPDATE?.data)).filter(
+              el => !Object.keys(query.target.elements).includes(el),
+            )
```
BobdenOs, line 31:
```suggestion
        const { query } = req
        const kind = q.kind || Object.keys(q)[0]
        const operation = query[kind]
        const columns = operation.colums || Object.keys(
          operation.entries?.reduce(Object.assign, {}) || 
          operation.data
        )
        const InvalidColumns = columns.filter(c => !query.elements[c])
```

db-service/lib/SQLService.js
```diff
@@ -20,6 +20,33 @@ class SQLService extends DatabaseService {
     this.on(['UPDATE'], this.transformStreamIntoCQN)
     this.on(['INSERT', 'UPSERT', 'UPDATE'], require('./fill-in-keys')) // REVISIT: should be replaced by correct input processing eventually
     this.on(['INSERT', 'UPSERT', 'UPDATE'], require('./deep-queries').onDeep)
+    if (cds.env.features.db_strict) {
+      this.on(['INSERT', 'UPSERT', 'UPDATE'], req => {
```
BobdenOs, line 24: My understanding is that the `on` handlers would have to call the `next` function to ensure that the flow keeps going. Currently it might work because of the implicit `next`, but my understanding is that this behavior is deprecated and should no longer be relied upon.

BobdenOs (changes requested):

test/scenarios/bookshop/strictMode.test.js
```diff
@@ -0,0 +1,124 @@
+process.env.cds_features_db__strict = 'true'
```
BobdenOs, line 1: This test file is not actually running

larslutz96: required the test

db-service/lib/SQLService.js
larslutz96: changed code snippet accordingly

sqlite/test/general/strictMode.test.js
```diff
@@ -0,0 +1,105 @@
+const cds = require('../../../test/cds.js')
```
BobdenOs, line 1: This test should probably be inside `test/compliance`. That way it would also be included inside the other database services their tests.

db-service/lib/SQLService.js
```diff
@@ -20,6 +20,39 @@ class SQLService extends DatabaseService {
...
+          const err =
+            invalidColumns.length === 1
+              ? createColumnError(query.target.name, invalidColumns[0])
+              : new Error('MULTIPLE_ERRORS')
+
+          if (invalidColumns.length > 1) {
+            err.details = invalidColumns.map(el => createColumnError(query.target.name, el))
+          }
+
+          throw Object.assign(err, { statusCode: 400 })
```
BobdenOs, line 51: Some how this comment was not included in the last review.

The database service is not allowed to use HTTP codes in the errors. As it is not possible to know what is the source of the erroneous query it cannot be determined that the code should be `400`. Arguably it is least likely that this would happen with queries coming from the protocol adapters as they tent to check their payload before passing it along to the database service. So most likely this error would result in a `500`.

Also the multiple errors occurred error message format used is `OData` specific and would be very confusing when the error happens multiple times in a OData request, because it would say multiple times that multiple errors occurred. 

```suggestion
        if (invalidColumns.length > 0) {
          cds.error(`STRICT MODE: Trying to ${kind} non existent columns (${invalidColumns})`)
```

test/scenarios/bookshop/strictMode.test.js
```diff
@@ -0,0 +1,128 @@
+const cds = require('../../cds.js')
```
BobdenOs, line 1: This test file is still not being executed. As it is not included inside the `test/scenarios/bookshop/index.js` file.

BobdenOs (changes requested):

test/scenarios/bookshop/strictMode.test.js
larslutz96: yes i moved it to sqlite/general so we can test this feature flag separately

db-service/lib/SQLService.js
larslutz96: thanks for the explanation. removed it according to your suggestion

db-service/lib/SQLService.js
```diff
@@ -20,6 +20,27 @@ class SQLService extends DatabaseService {
...
+            operation.entries?.reduce((acc, obj) => {
+              return Object.assign(acc, obj)
+            }, {}) || operation.data,
+          )
+        const invalidColumns = columns.filter(c => !query.target?.elements?.[c])
+
+        if (invalidColumns.length > 0) {
+          cds.error(`STRICT MODE: Trying to ${kind} non existent columns (${invalidColumns})`)
+        }
+        next()
```
BobdenOs, line 41:
```suggestion
        return next()
```

BobdenOs (commented):
What about https://github.com/cap-js/cds-dbs/pull/384/files#r1427978987 ?

larslutz96:
> What about https://github.com/cap-js/cds-dbs/pull/384/files#r1427978987 ?

when i set the feature flag `cds.env.features.db_strict` to true its also enabled for the other test required inside the compliance folder. Therefore i moved the test to sqlite/general and required it for the other db services.

sqlite/test/general/strictMode.test.js
```diff
@@ -0,0 +1,117 @@
...
+process.env.cds_features_db__strict = 'true'
+
+describe('strict mode', () => {
+  cds.test(__dirname, 'model.cds')
+
+  afterAll(() => {
+    process.env.cds_features_db__strict = undefined
+  })
```
BobdenOs, line 9: This should fix the issue that it would impact other tests. If tests impact other tests it is usually a flaw in the test definition.

```suggestion
describe('strict mode', () => {
  cds.test(__dirname, 'model.cds')

  beforeAll(() => {
    process.env.cds_features_db__strict = 'true'
  })

  afterAll(() => {
    process.env.cds_features_db__strict = undefined
  })
```

larslutz96: i already tried this solution. But still the env variable was true in the other tests

test/compliance/resources/srv/index.cds
```diff
@@ -0,0 +1,5 @@
+using {complex} from '../db/index.cds';
+
+service ComplianceService {
```
BobdenOs, line 3: is this really required ?

BobdenOs (approved):

db-service/lib/SQLService.js
```diff
@@ -23,6 +23,26 @@ class SQLService extends DatabaseService {
...
+      this.on(['INSERT', 'UPSERT', 'UPDATE'], ({ query }, next) => {
+        const kind = query.kind || Object.keys(query)[0]
+        const operation = query[kind]
+        if (!operation.columns && !operation.entries && !operation.data) return next()
+        const columns =
+          operation.columns ||
+          Object.keys(
+            operation.entries?.reduce((acc, obj) => {
+              return Object.assign(acc, obj)
+            }, {}) || operation.data,
```
danjoa, line 36:
```suggestion
            operation.data || operation.entries?.reduce((acc, obj) => {
              return Object.assign(acc, obj)
            }, {}),
```

db-service/lib/SQLService.js
```diff
@@ -23,6 +23,26 @@ class SQLService extends DatabaseService {
   init() {
     this.on(['INSERT', 'UPSERT', 'UPDATE'], require('./fill-in-keys')) // REVISIT should be replaced by correct input processing eventually
     this.on(['INSERT', 'UPSERT', 'UPDATE'], require('./deep-queries').onDeep)
+    if (cds.env.features.db_strict) {
+      this.on(['INSERT', 'UPSERT', 'UPDATE'], ({ query }, next) => {
+        const kind = query.kind || Object.keys(query)[0]
```
danjoa, line 28:
```suggestion
        const elements = query.target?.elements; if (!elements) return
        const kind = query.kind || Object.keys(query)[0]
```

db-service/lib/SQLService.js
```diff
@@ -23,6 +23,26 @@ class SQLService extends DatabaseService {
...
+        const operation = query[kind]
+        if (!operation.columns && !operation.entries && !operation.data) return next()
+        const columns =
+          operation.columns ||
+          Object.keys(
+            operation.entries?.reduce((acc, obj) => {
+              return Object.assign(acc, obj)
+            }, {}) || operation.data,
+          )
+        const invalidColumns = columns.filter(c => !query.target?.elements?.[c])
```
danjoa, line 38:
```suggestion
        const invalidColumns = columns.filter(c => !(c in elements)) // IMPORTANT: also support falsy values in data
```

db-service/lib/SQLService.js
```diff
@@ -23,6 +23,26 @@ class SQLService extends DatabaseService {
...
+
+        if (invalidColumns.length > 0) {
+          cds.error(`STRICT MODE: Trying to ${kind} non existent columns (${invalidColumns})`)
+        }
+        return next()
```
danjoa, line 43:
```suggestion
```

danjoa (commented):
Instead of upfront extra loops, I always prefer to put such checks into the code which has to loop anyways

larslutz96:
> Instead of upfront extra loops, I always prefer to put such checks into the code which has to loop anyways

updated the code according to your suggestions

db-service/lib/SQLService.js
```diff
@@ -23,6 +23,26 @@ class SQLService extends DatabaseService {
...
+        const operation = query[kind]
+        if (!operation.columns && !operation.entries && !operation.data) return
+        const columns =
+          operation.columns ||
+          Object.keys(
+            operation.data || operation.entries?.reduce((acc, obj) => {
+              return Object.assign(acc, obj)
+            }, {}),
+          )
+          const invalidColumns = columns.filter(c => !(c in elements)) // also support falsy values in data
```
BobdenOs, line 39: The comment is incorrect as `elements` is the entity definition. There for is not related to the values. The `in` operator is good.

```suggestion
          const invalidColumns = columns.filter(c => !(c in elements))
```

danjoa (approved):

PR #386, opened 2023-12-15
test: remove accidental (?) .only

patricebender:
in #32 a test has been added, I assume it should not be the `.only` one run as part of this test suite. @johannes-vogel


johannes-vogel:
😵 shame on me. Yes, that was accidental.

Thanks for fixing it.

PR #392, opened 2023-12-20
fix: ignore empty order by

BobdenOs:
In the base `cqn2sql` implementation the `orderBy` is checked for its length. This check was missing in the HANA Service.


johannes-vogel (approved):

johannes-vogel:
Did you observe that in any test project?
Or only by comparing the implementation?

BobdenOs:
This issue was hit when trying to reproduce another error with the `@cap-js/hana@0.0.3` release.

PR #398, opened 2023-12-27
fix: Do not generate UUIDs for association keys

mariayord:
Fixes : https://github.tools.sap/cap/cdsnode/issues/1699


sqlite/test/general/uuid.test.js
```diff
@@ -25,4 +25,16 @@ describe('UUID Generation', () => {
       await DELETE('test.bar')
     })
   })
+
+  test('INSERT entity with missing key as association throws error', async () => {
+    const db = await cds.connect.to('db')
+    return db.tx(async () => {
+      try {
+       await INSERT.into('test.BooksWithAssocAsKey').entries([{}])
```
BobdenOs, line 33: Please use [this](https://jestjs.io/docs/tutorial-async#rejects)

db-service/lib/fill-in-keys.js
```diff
@@ -8,7 +8,8 @@ const generateUUIDandPropagateKeys = (target, data, event) => {
   if (!data) return
   const keys = target.keys
   for (const key in keys) {
-    if (keys[key].type === 'cds.UUID' && !data[key] && event === 'CREATE') {
+    const keyValue = keys[key]
```
BobdenOs, line 11: it is not a value. It is the type.

BobdenOs (approved):
Looks good. Once the test is passing.

PR #401, opened 2024-01-02
test(deep-update): add regression test for huge payloads

johannes-vogel:
We've received a ticket for old db implementation that the deep update does not accept payloads with lots of children.
This test ensures that our new implementation is (and stays) capable of it.

The max number of placeholders for sqlite is 32766 which is documented here: https://www.sqlite.org/limits.html
The max number of placeholders for hana is ? → from experience somewhat above 3000
The max number of placeholders for postgres is 65536 → https://github.com/brianc/node-postgres/issues/2579

### DIFF
--- a/test/compliance/UPDATE.test.js
+++ b/test/compliance/UPDATE.test.js
@@ -1,4 +1,9 @@
+const cds = require('../cds.js')
+
 describe('UPDATE', () => {
+  const { expect, data } = cds.test(__dirname + '/resources')
+  data.autoIsolation(true)
+
   describe('entity', () => {
     test.skip('missing', () => {
       throw new Error('not supported')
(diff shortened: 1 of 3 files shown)


test/compliance/UPDATE.test.js
```diff
@@ -16,4 +21,18 @@ describe('UPDATE', () => {
       throw new Error('not supported')
     })
   })
+
+  describe('deep update', () => {
+    test('huge deep update does not cause sql exception', async () => {
+      // sqlite limit of placeholders is 32766 -> 9. of https://www.sqlite.org/limits.html
+      await INSERT.into('compositions.Travel').entries({
+        ID: 4711,
+        bookings: (new Array(33000).fill(0).map((_, i) => ({ ID: i})))
```
BobdenOs, line 30: Would it make sense to have a scaling number ? To identify the limits for all database implementations.

BobdenOs:
Seems that the github runner does not agree with this test with HANA.

```
TimeoutError: ResourceRequest timed out
```

PR #405, opened 2024-01-09
Skip virtual fields on UPSERTs

danjoa:


db-service/lib/cqn2sql.js
```diff
@@ -620,15 +620,16 @@ class CQN2SQLRenderer {
    * @returns {string} SQL
    */
   UPSERT(q) {
-    let { UPSERT } = q,
-      sql = this.INSERT({ __proto__: q, INSERT: UPSERT })
+    const { UPSERT } = q
+    const elements = q.target?.elements || {}
+    let sql = this.INSERT({ __proto__: q, INSERT: UPSERT })
     let keys = q.target?.keys
     if (!keys) return (this.sql = sql) // REVISIT: We should converge q.target and q._target
```
BobdenOs, line 627: The `q.target` topic is already being addressed so this comment could be removed.
```suggestion
    if (!keys) return (this.sql = sql)
```

BobdenOs (approved):

PR #406, opened 2024-01-09
fix: Restore former deep upsert behavior / error

danjoa:
Excluding associations from UPSERT broke a test in @sap/cds, so this PR restores the former behavior for the time being which let updates to associations slip through and fail with a database error. 

We need to follow up on what the correct behavior should be...


danjoa:
Force-merged to unlock put internal test pipelines

PR #408, opened 2024-01-11
chore: unify physical elements check

BobdenOs:
Add an `exists(<element>)` function into `cqn2sql` to unify checking whether an element exists in the database.

```js
element.virtual // virtual does not exist on the database
element.value // calculated column cannot be written into
element.isAssociation // associations are meta columns that cannot be addressed
```


BobdenOs:
this change still breaks `@sap/cds` tests.

danjoa (commented):
I think in most cases we allow `!(c in elements)` to read or feed native db columns, not in the model → would only reject in _strict_ mode. 

In addition we need to follow up on the `isAssociation` case marked as revisit, which breaks a test in `@sap/cds` if we always skip these.

BobdenOs:
We do allow writing into elements that don't exists inside the model. The distinction that this PR is trying to address is the difference between the model and the actual tables on the database. For us to be able to allow users to provide `associations` and `virtual` fields it is required that we don't include these values inside the `SQL` statements or the database would throw.

Additionally for the `UPSERT` case with the `isAssociation` check it should be allowed to provide data inside an `association` column, but currently it is not allowed to do **deep** `UPSERT` queries. Which is the test case that is failing. If we define that with the new databases it is allowed to do deep `UPSERT` it would mean that the current expectation is wrong. As currently the `UPSERT` implementation is wrong when providing data inside an `association` column.

danjoa:
> currently it is not allowed to do deep UPSERT queries. Which is the test case that is failing

Yes, and that test is strangely written in a way that it expects associations to pass through and crash in the database → if we are sure that this test is written badly and fix it, I think we can filter associations always.

Issue #409, opened 2024-01-12
SQLite: Update can lead to Uncaught RangeError: Invalid time value


nlaenger:
### Description of erroneous behaviour 

When updating a string field with a text containing a JS timestamp it will result in an ungaught excption.
example
```TS
  LOG._info && LOG.info("This will break:");
  await UPDATE.entity(Belege)
    .with({
      descr: 'Timestamp in JS looks like "2025-04-01T15:00:00Z"',
    })
    .where({ ID: "0f3966ef-142c-4d72-9694-c0f06e898d98" });
```

Will result in:
```sh
[cds] - ❗️Uncaught RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at CQN2SQLite.val (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\cqn2sql.js:770:49)
    at CQN2SQLite.val (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\sqlite\lib\SQLiteService.js:177:20)
    at C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\cqn2sql.js:553:38
    at _add (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\cqn2sql.js:558:40)
    at CQN2SQLite.UPDATE (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\cqn2sql.js:553:15)
    at CQN2SQLite.render (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\cqn2sql.js:72:15)
    at SQLiteService.cqn2sql (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\SQLService.js:292:20)
    at SQLiteService.onSIMPLE (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\SQLService.js:150:34)
    at SQLiteService.onUPDATE (C:\issue-cds-sqlite-text-timestamp\node_modules\@cap-js\db-service\lib\SQLService.js:126:17)
```

it seams the regex [in @cap-js/sqlite/lib/SQLiteService.js](https://github.com/cap-js/cds-dbs/blob/251575b8497e305eea575391a875475a7d083b9a/sqlite/lib/SQLiteService.js#L176) sees a correct date within the string and from there it gets bad.

[...]

PR #410, opened 2024-01-12
fix: sqlite date string compatibility parsing only for valid dates

BobdenOs:
Enhances the `DateTime` / `Timestamp` regex to match only for exact `DateTime` / `Timestamp` strings. Additionally check whether the parsed `Date` is valid. If not keep the original string to reduce the chance of things breaking.

fixes: #409


patricebender (approved):

db-service/lib/cqn2sql.js
```diff
@@ -832,9 +832,9 @@ class CQN2SQLRenderer {
       case 'number': return `${val}` // REVISIT for HANA
       case 'object':
         if (val === null) return 'NULL'
-        if (val instanceof Date) return `'${val.toISOString()}'`
-        if (val instanceof Readable); // go on with default below
-        else if (Buffer.isBuffer(val)); // go on with default below
+        if (val instanceof Date) val = val.toJSON() // returns null if invalid
```
danjoa, line 835: When would an instance of Date be invalid?

db-service/lib/cqn2sql.js
```diff
@@ -832,9 +832,9 @@ class CQN2SQLRenderer {
...
+        else if (val instanceof Readable); // go on with default below
+        else if (Buffer.isBuffer(val)) val = val.toString('base64')
```
danjoa, line 837: Why did that work before and not any longer?

sqlite/lib/SQLiteService.js
```diff
@@ -167,7 +167,12 @@ class SQLiteService extends SQLService {
     val(v) {
       if (Buffer.isBuffer(v.val)) v.val = v.val.toString('base64')
       // intercept DateTime values and convert to Date objects to compare ISO Strings
-      else if (/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}[Z+-]/.test(v.val)) v.val = new Date(v.val)
+      else if (/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d{1,9})?(Z|[+-]\d{2}(:?\d{2})?)$/.test(v.val)) {
```
danjoa, line 170: Why do we need to intercept that at all?

sqlite/lib/SQLiteService.js
```diff
@@ -167,7 +167,12 @@ class SQLiteService extends SQLService {
...
+        const date = new Date(v.val)
+        if (!Number.isNaN(date.getTime())) {
```
danjoa, line 172: When could that happen, given the regex check above?

sqlite/lib/SQLiteService.js
```diff
@@ -167,7 +167,12 @@ class SQLiteService extends SQLService {
...
+          v.val = new Date(v.val)
```
danjoa, line 173:
```suggestion
          v.val = date
```

db-service/lib/cqn2sql.js
BobdenOs: Good catch was a side effect from a merge conflict.

sqlite/lib/SQLiteService.js
BobdenOs: This has to be unified for when selecting data from a `DateTime` columns and then using the result in a `where` clause. As the `DateTime` value will be stored on `SQLite` in `Timestamp` formatting, but will be returned as `DateTime` format when selected.

danjoa (commented):
→ see inline questions...

sqlite/lib/SQLiteService.js
BobdenOs: Also this logic is only for `SQLite` as all other database have binary `Date` / `Time` formats.

db-service/lib/cqn2sql.js
BobdenOs: When someone would provide an invalid `Date` in their query.

```js
SELECT.from('Books').where(`createdAt <`, new Date('0000-00-00'))
```

This would throw an error when parsing it into an `ISO` string. Without any indication where the `Date` instance was created. Making it look like an database service error.

sqlite/lib/SQLiteService.js
danjoa: Do we have a test that shows that use case and expected behavior? 
Respectively: we should have one, I think :)

db-service/lib/cqn2sql.js
danjoa: Ok. But then, is it correct to silently have that running into another SQL error like 'comparing with NULL osing < operator is not supported'?

BobdenOs: Having the database throw the error would be a lot more clear in this instance. Another option would be to check whether the `Date` is valid and throw our own error. As currently the error thrown is really confusing.

```
Uncaught RangeError: Invalid time value
<database service stack traces>
```

Where a `SQL` error would at least include the original `SQL` statement that caused the error. With often additional information showing which operator / column was being compared with.

sqlite/lib/SQLiteService.js
BobdenOs: Added a test which shows that both `Timestamp` and `DateTime` values can be compared against `DateTime` columns (https://github.com/cap-js/cds-dbs/pull/410/commits/46805e6846562f3272c647909dc255119a81f92b).

danjoa: 👍 thanks. 
But then: these tests are also green when I revert that change here

BobdenOs: Yes the test that check the current change is [this](https://github.com/cap-js/cds-dbs/pull/410/files#diff-a51794365405e10091720b99518753899c1f3d2dd6395edf483b465560c51e9a). Where the string contains a valid `DateTime` / `Timestamp`, but should not be converted into a `Date`.

danjoa (approved):

PR #416, opened 2024-01-17
feat: SELECT returns binaries as Buffers

vkozyura:


patricebender (commented):
please add a corresponding test

test/compliance/CREATE.test.js
patricebender, line 1: if this feature is only relevant for sqlite, it should be added as an sqlite specific test, not as part of the compliance tests which are run for all dbs, shouldnt it?

BobdenOs (approved):

patricebender (approved):

PR #417, opened 2024-01-17
chore: expose native hana lob stream on structured selects

BobdenOs:


danjoa (approved):
Looked through it for the changes I could judge, and trust you do the right things for the rest :)

PR #419, opened 2024-01-18
fix: show clear error message when unable to load project package.json

sjvans:


BobdenOs (approved):

hana/lib/drivers/index.js
```diff
@@ -5,18 +5,17 @@ Object.defineProperties(module.exports, {
...
+        const projectPackage = require(cds.root + '/package.json')
+        const dependencies = {
+          ...projectPackage.dependencies,
+          ...(process.env.NODE_ENV !== 'production' && projectPackage.devDependencies),
+        }
+        // Have a bias to hdb as the default driver
+        if (dependencies.hdb) return module.exports.hdb
+        if (dependencies['@sap/hana-client']) return module.exports['hana-client']
+      } catch (e) {
+        console.trace(`WARNING! Unable to require the project's package.json at "${cds.root + '/package.json'}". Please check your project setup.`)
```
johannes-vogel, line 18: rethrow error?

BobdenOs: It should not rethrow it should just act like no driver was defined.

PR #420, opened 2024-01-18
chore: use better-sqlite3@9.2.2

sjvans:
9.3 has breaking changes re JSON functions

### DIFF
--- a/hana/package.json
+++ b/hana/package.json
@@ -20,7 +20,7 @@
     "npm": ">=8"
   },
   "scripts": {
-    "test": "jest --silent",
+    "test": "npm start && jest --silent",
     "start": "npm run start:hce || npm run start:hxe",
     "start:hce": "cd ./tools/docker/hce/ && ./start.sh",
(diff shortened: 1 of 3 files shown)


patricebender (commented):
should we really pin a specific version? Could the json functionality be adapated? And why are there breaking changes in minor releases 😵‍💫

BobdenOs:
> should we really pin a specific version? Could the json functionality be adapated? And why are there breaking changes in minor releases 😵‍💫

The `better-sqlite3@9.3.0` bumps the minor version of `sqlite3.45.0` which has the following [changelog](https://sqlite.org/releaselog/3_45_0.html)

Which states:

> All JSON functions are rewritten to use a new internal parse tree format called [JSONB](https://sqlite.org/json1.html#jsonbx)

Of which it seems to not have achieved full compatibility with the original implementation. Or potentially with the `JSON` specification.

BobdenOs:
@sjvans this change will no longer be required. When the fix from https://github.com/cap-js/cds-dbs/pull/422 is in place.

PR #422, opened 2024-01-22
fix: make @cap-js/sqlite work with better-sqlite3@9.3.0

BobdenOs:
With the release of `better-sqlite3@9.3.0` it bumped the sqlite version to `3.45.0`. Which comes with a complete rework of the `json` functions. With the big difference being that it now supports an internal `jsonb` format. Which is supposed to be only used by `sqlite` itself, but with this new change all the `json` functions accept `binary` data as the `jsonb` format. Therefor as the json that we produced was being converted into `Buffer`s it resulted in `better-sqlite3` providing them as `binaries` to the `json_each` function which made it interoperate it as an malformed `jsonb`. So this change forced the `json` to be bound as a `strings` which makes everything in its own behavior.

Additionally now that the `json_insert` function has a `jsonb_insert` equavelant it fixes a memory issue `sqlite` had introduced in the past and we can remove the `json_merge` custom function again. Which removes the requirement to call back to `js` when selecting more then `50` columns. Which should make reading many columns faster again.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -257,9 +257,9 @@ class CQN2SQLRenderer {
     // Prevent SQLite from hitting function argument limit of 100
     let obj = "'{}'"
     for (let i = 0; i < cols.length; i += 48) {
-      obj = `json${isRoot ? '' : 'b'}_insert(${obj},${cols.slice(i, i + 48)})`
+      obj = `jsonb_insert(${obj},${cols.slice(i, i + 48)})`
     }
-    return `SELECT ${isRoot || SELECT.one ? obj : `jsonb_group_array(${obj})`} as _json_ FROM (${sql})`
+    return `SELECT ${isRoot || SELECT.one ? obj.replace('jsonb', 'json') : `jsonb_group_array(${obj})`} as _json_ FROM (${sql})`
   }
 
   /**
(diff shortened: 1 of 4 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -252,21 +252,14 @@ class CQN2SQLRenderer {
...
-      obj = `json_merge(${chunks})`
+    // Prevent SQLite from hitting function argument limit of 100
+    let obj = "'{}'"
+    for (let i = 0; i < cols.length; i += 48) {
+      obj = `json${isRoot ? '' : 'b'}_insert(${obj},${cols.slice(i, i + 48)})`
     }
-
-
-    return `SELECT ${SELECT.one || SELECT.expand === 'root' ? obj : `json_group_array(${obj.includes('json_merge') ? `json_insert(${obj})` : obj})`} as _json_ FROM (${sql})`
+    return `SELECT ${isRoot || SELECT.one ? obj : `jsonb_group_array(${obj})`} as _json_ FROM (${sql})`
```
patricebender, line 262: just wondering about the sqlite specific stuff in the generic `db-service`, could you elaborate why we need this here?

danjoa (approved):
Looks good to me

db-service/lib/cqn2sql.js
BobdenOs: It has always been in the generic `db-service`. The original reason for having it here is that there is no `SQL` specified uniform `API` to create `json` values. Therefor taking `SQLite` makes it easy for other database services to find their equivalent functions.

patricebender: Got it. Thanks!

PR #424, opened 2024-01-24
chore: improved debugging information

BobdenOs:


db-service/lib/cqn2sql.js
```diff
@@ -77,11 +77,20 @@ class CQN2SQLRenderer {
...
-    )
+    if (DEBUG) { // Only process debuggin
+      const values = this.values.length
+        ? this.values
+        : q[kind].entries ?? q[kind].rows ?? q[kind].data
+      const sanitize_values = process.env.NODE_ENV === 'production' && cds.env.log.sanitize_values !== false
+      DEBUG(
+        this.sql,
+        sanitize_values
+          ? (Array.isArray(values) && values.length) || typeof values === 'object'
```
danjoa, line 88:
```suggestion
? typeof values === 'object'
```

This condition will also be true for empty arrays, so the first part of that is useless

PR #425, opened 2024-01-24
fix: properly support `default`, `cds.on.insert` and `cds.on.update` for `UPSERT` queries

SamuelBrucksch:
This PR unifies the `UPSERT` implementation of all the database services. resulting in 2 helper functions in each database service that allow for defining the database specific JSON extraction logic and the database specific `undefined` check. With these special cases covered it is possible to generate `INSERT` and `UPSERT` statements that correctly apply `default`, `@cds.on.insert` and `@cds.on.update` values. Which currently only apply to `@cap-js/hana` and is a big blocker for most use cases of `UPSERT` and the `@sap/cds/common.cds` use the `@cds.on.insert` annotations. Which with the current `UPSERT` implementation could possibly be `null` which should be impossible as every entry in the database has to have been inserted once.


danjoa (changes requested):
Let's review that in one of our next syncs please ...
These are a lot of code changes all over the place, just to add support for defaults?
Particularly dislike, that we now seem to do quoting everywhere?

P.S. After the big cleanup we need extra discipline to keep it clean ;)

db-service/lib/cqn2sql.js
```diff
@@ -492,12 +494,25 @@ class CQN2SQLRenderer {
...
+          const merges = extractions.filter(c => elements?.[c.name]?.['@cds.extension'])
+          if (merges.length) {
+            c.sql = `json_set(ifnull(${c.sql},'{}'),${merges.map(
+              c => this.string('$."' + c.name + '"') + ',' + c.sql,
+            )})`
+          }
+        }
+        return c
+      })
+      .filter(a => a)
```
danjoa, line 514: All this is for the UIFlex thing which we removed in cds8, isn't it? → why do we still need that?

db-service/lib/cqn2sql.js
```diff
@@ -492,12 +494,25 @@ class CQN2SQLRenderer {
...
+      .map(c => c.insert)
```
danjoa, line 515: Before this PR `c.insert` was `c.sql`?

db-service/lib/cqn2sql.js
```diff
@@ -719,14 +734,36 @@ class CQN2SQLRenderer {
    */
   UPSERT(q) {
     const { UPSERT } = q
+    const entity = this.name(q.target?.name || UPSERT.into.ref[0])
```
danjoa, line 737: Why `q.target?.name` ?

db-service/lib/cqn2sql.js
```diff
@@ -719,14 +734,36 @@ class CQN2SQLRenderer {
...
+    const alias = UPSERT.into.as
```
danjoa, line 738: Move that line down to where it is required → principle of locality

db-service/lib/cqn2sql.js
```diff
@@ -719,14 +734,36 @@ class CQN2SQLRenderer {
...
     const elements = q.target?.elements || {}
-    let sql = this.INSERT({ __proto__: q, INSERT: UPSERT })
+    const insert = this.INSERT({ __proto__: q, INSERT: UPSERT })
```
danjoa, line 740: Should we adhere to using variable name `sql` for generated SQL?

db-service/lib/cqn2sql.js
```diff
@@ -719,14 +734,36 @@ class CQN2SQLRenderer {
...
+
     let keys = q.target?.keys
-    if (!keys) return this.sql = sql
+    if (!keys) return insert
     keys = Object.keys(keys).filter(k => !keys[k].isAssociation && !keys[k].virtual)
 
-    let updateColumns = q.UPSERT.entries ? Object.keys(q.UPSERT.entries[0]) : this.columns
-    updateColumns = updateColumns.filter(c => {
+    // temporal data
+    keys.push(...ObjectKeys(q.target.elements).filter(e => q.target.elements[e]['@cds.valid.from']))
```
danjoa, line 747: Use `for..in` loop instead

db-service/lib/cqn2sql.js
```diff
@@ -1030,50 +1071,92 @@ class CQN2SQLRenderer {
...
       }
+      // if (sql[0] !== '$') sql = converter(sql, element)
+
+      let onInsert = this.managed_session_context(element[cdsOnInsert]?.['='])
+        || this.managed_session_context(element.default?.ref?.[0])
+        || (element.default?.val !== undefined && { val: element.default.val, param: false })
+      let onUpdate = this.managed_session_context(element[cdsOnUpdate]?.['='])
+
+      onInsert = onInsert && this.expr(onInsert)
+      onUpdate = onUpdate && this.expr(onUpdate)
```
danjoa, line 1115:
```suggestion
      if (onInsert) onInsert = this.expr(onInsert)
      if (onUpdate) onUpdate = this.expr(onUpdate)
```

danjoa:
Some more observations below, using this generated query out of `INSERT.into(Books,{title:'t'})`:

```sql
INSERT INTO sap_capire_bookshop_Books (
  createdAt,
  createdBy,
  modifiedAt,
  modifiedBy,
  ID,
  ...
) SELECT 
  (
    CASE
      WHEN json_type(value, '$."createdAt"') IS NULL THEN session_context('$now')
      ELSE value ->> '$."createdAt"'
    END
  ), (
    CASE
      WHEN json_type(value, '$."createdBy"') IS NULL THEN session_context('$user.id')
      ELSE value ->> '$."createdBy"'
    END
  ), (
    CASE
      WHEN json_type(value, '$."modifiedAt"') IS NULL THEN session_context('$now')
      ELSE value ->> '$."modifiedAt"'
    END
  ), (
    CASE
      WHEN json_type(value, '$."modifiedBy"') IS NULL THEN session_context('$user.id')
      ELSE value ->> '$."modifiedBy"'
    END
  ),
  value ->> '$."ID"',
  ...
FROM json_each(?)
```
...

danjoa:
Why are we quoting names in JSON path arguments? → i.e. 
```sql
SELECT
      value ->> '$."ID"',
      value ->> '$."title"',
      value ->> '$."descr"',
      ...
```
could be: 
```sql
```sql
SELECT
      value ->> '$.ID',
      value ->> '$.title',
      value ->> '$.descr',
      ...
```

danjoa:
Superfluous `(...)`: 

```sql
SELECT 
  (
    CASE
      WHEN json_type(value, '$."createdAt"') IS NULL THEN session_context('$now')
      ELSE value ->> '$."createdAt"'
    END
  ), ...
```
→ 
```sql
SELECT 
    CASE
      WHEN json_type(value, '$."createdAt"') IS NULL THEN session_context('$now')
      ELSE value ->> '$."createdAt"'
    END, 
    ...
```

danjoa:
Before this PR the following expressions where used for `createdAt` 
```sql
    COALESCE ( ISO(value ->> '$.createdAt'), session_context('$now') )
```
Now it's this: 
```sql
    CASE
      WHEN json_type(value, '$.createdAt') IS NULL THEN session_context('$now')
      ELSE value ->> '$.createdAt'
    END
```

1. Why can't we keep using `COALESCE`?
2. Don't we need to use the `ISO` function anymore?

danjoa:
I did a quick performance comparison, running 1e4 inserts each: 

- JSON Insert w/ this PR → 915 ms 
- JSON Insert before → 687 ms
- Plain Data Insert → 179 ms

danjoa:
Btw: `cds w cap/sample/bookshop` crashes with this PR

db-service/test/cqn2sql/__snapshots__/upsert.test.js.snap
```diff
@@ -29,6 +29,6 @@ exports[`upsert test with rows (quoted) 1`] = `
       "[[1,null,2]]",
     ],
   ],
-  "sql": "INSERT INTO """Foo2Quoted""" ("""ID""","""name""","""a""") SELECT value->>'$[0]',value->>'$[1]',value->>'$[2]' FROM json_each(?) WHERE true ON CONFLICT("""ID""") DO UPDATE SET """name""" = excluded."""name""","""a""" = excluded."""a"""",
+  "sql": "INSERT INTO """Foo2Quoted""" ("""ID""","""name""","""a""") SELECT (CASE WHEN OLD."""ID""" IS NULL THEN value->>'$."""ID"""' ELSE (CASE WHEN json_type(value,'$."""ID"""') IS NULL THEN OLD."""ID""" ELSE value->>'$."""ID"""' END) END) as """ID""",(CASE WHEN OLD."""ID""" IS NULL THEN value->>'$."""name"""' ELSE (CASE WHEN json_type(value,'$."""name"""') IS NULL THEN OLD."""name""" ELSE value->>'$."""name"""' END) END) as """name""",(CASE WHEN OLD."""ID""" IS NULL THEN value->>'$."""a"""' ELSE (CASE WHEN json_type(value,'$."""a"""') IS NULL THEN OLD."""a""" ELSE value->>'$."""a"""' END) END) as """a""" FROM (SELECT value, value->>'$."""ID"""' as """ID""" from json_each(?)) as NEW LEFT JOIN """Foo2Quoted""" AS OLD ON NEW."""ID"""=OLD."""ID""" WHERE TRUE ON CONFLICT("ID") DO UPDATE SET """name""" = excluded."""name""","""a""" = excluded."""a"""",
```
David-Kunz, line 32: For keys, I'm a bit confused by

```sql
(CASE 
  WHEN OLD."ID" IS NULL THEN value->>'$."ID"' 
  ELSE 
  (CASE 
    WHEN json_type(value, '$."ID"') IS NULL THEN OLD."ID" 
    ELSE value->>'$."ID"' 
    END) 
  END) AS "ID"
```

We demand that all keys are provided, otherwise `UPSERT` doesn't make sense. Hence, for key fields, we can always take the value from the JSON input, no?

```sql
value->>'$."ID"'
```

db-service/lib/cqn2sql.js
```diff
@@ -1030,50 +1071,92 @@ class CQN2SQLRenderer {
    * @param {Boolean} isUpdate
    * @returns {string[]} Array of SQL expressions for processing input JSON data
    */
-  managed(columns, elements, isUpdate = false) {
-    const annotation = isUpdate ? '@cds.on.update' : '@cds.on.insert'
+  managed(columns, elements) {
```
David-Kunz, line 1074: Typo in JSDoc, `Convers` -> `Converts`

db-service/lib/cqn2sql.js
```diff
@@ -1030,50 +1071,92 @@ class CQN2SQLRenderer {
...
-          e =>
-            (elements[e]?.[annotation] || (!isUpdate && elements[e]?.default && !elements[e].virtual && !elements[e].isAssociation)) &&
-            !columns.find(c => c.name === e),
-        )
+      : ObjectKeys(elements)
+        .filter(e => {
+          const element = elements[e]
+          // Physical column check
+          if (!element || element.virtual || element.isAssociation) return false
+          // Existance check
```
David-Kunz, line 1087:
```suggestion
          // Existence check
```

db-service/lib/cqn2sql.js
```diff
@@ -1030,50 +1071,92 @@ class CQN2SQLRenderer {
...
+          if (columns.find(c => c.name === e)) return false
+          // Actual mandatory check
+          if (element.default || element[cdsOnInsert] || element[cdsOnUpdate]) return true
```
David-Kunz, line 1090: Switch the order of those two checks, property lookup is faster

postgres/lib/PostgresService.js
```diff
@@ -494,8 +518,11 @@ GROUP BY k
       ...super.InputConverters,
       // UUID:      (e) => `CAST(${e} as UUID)`, // UUID is strict in formatting sflight does not comply
       boolean: e => `CASE ${e} WHEN 'true' THEN true WHEN 'false' THEN false END`,
-      Float: (e, t) => `CAST(${e} as decimal${t.precision && t.scale ? `(${t.precision},${t.scale})` : ''})`,
-      Decimal: (e, t) => `CAST(${e} as decimal${t.precision && t.scale ? `(${t.precision},${t.scale})` : ''})`,
+      // REVISIT: Postgres and HANA round Decimal numbers differently therefor precision and scale are removed
```
David-Kunz, line 521:
```suggestion
      // REVISIT: Postgres and HANA round Decimal numbers differently therefore precision and scale are removed
```

db-service/lib/cqn2sql.js
```diff
@@ -1030,50 +1071,92 @@ class CQN2SQLRenderer {
...
+      let onUpdate = this.managed_session_context(element[cdsOnUpdate]?.['='])
+
+      onInsert = onInsert && this.expr(onInsert)
+      onUpdate = onUpdate && this.expr(onUpdate)
+
+      const insert = onInsert ? this.managed_default(name, onInsert, sql) : sql
+      const update = onUpdate ? this.managed_default(name, onUpdate, sql) : sql
+      const upsert = keyZero && (
+        // If both insert and update have the same managed definition exclude the old value check
+        (onInsert && onUpdate && insert === update)
```
David-Kunz, line 1121:
```suggestion
        (element.key || onInsert && onUpdate && insert === update)
```

David-Kunz: reason: https://github.com/cap-js/cds-dbs/pull/425/files#r1743310464

db-service/lib/cqn2sql.js
BobdenOs: This code was put in here to make `mtxs` pass its tests. My understanding is that in the mean time someone looked into the root cause of this arriving at the database service. Therefor it should be possible to remove this logic from the database now.

After applying all review comments the internal tests will be run with this PR to verify this.

db-service/lib/cqn2sql.js
BobdenOs: This PR changes the re-use function `managed` to better serve `INSERT`, `UPSERT` and `UPDATE`. As `UPSERT` requires both the `INSERT` and `UPDATE` behavior it now computes both inside `managed` and exposes all scenarios with their respective names.

This makes it also so that `managed` always generates the same results for any entity. Which makes it cacheable, but is not currently done.

db-service/lib/cqn2sql.js
BobdenOs: My question would rather be the other way around. Why it still has the fallback to the `into` as now `infer` should always put the `target` onto the query with the correct name even when the `target` is flagged as unresolved. As `q.target` should always be properly inferred allow `UPSERT` to be called on path expressions.

```suggestion
    const entity = this.name(q.target.name)
```

db-service/lib/cqn2sql.js
BobdenOs: This has been bothering me for a while now. Is it really necessary to respect the `alias` for `INSERT`, `UPSERT` and `UPDATE` queries ? My understanding is that these aliases don't do anything except for making some  unit test pass.

These two queries would do the exact same. It is not even clear to me where this alias could possibly be consumed.
```
INSERT INTO sap_capire_bookshop_Books as Books (<columns>) <data>

INSERT INTO sap_capire_bookshop_Books (<columns>) <data>
```

If you agree I will take this opportunity to just ignore these root modifying query aliases.

db-service/lib/cqn2sql.js
BobdenOs: Side effect of trying to make the code readable. Will be adjusted 👍

BobdenOs:
> Btw: `cds w cap/sample/bookshop` crashes with this PR

Thanks for the heads up. I have identified the root cause as the `UPSERT` inside `init` [here](https://github.com/SAP-samples/cloud-cap-samples/blob/0ba1f2e294fc815f46fed7c30089ffdce59f8af5/bookshop/db/init.js#L12). This PR currently assumed that the only valid `UPSERT` format is as documented [here](https://cap.cloud.sap/docs/node.js/cds-ql#upsert). Which only mentions `entries` not `rows`. Depending on what we want to support we can update this PR.

This gap was not caught by the tests as the bookshop used in the test is located inside the repository [here](
https://github.com/cap-js/cds-dbs/blob/main/test/bookshop/db/init.js#L15). This should probably be aligned with the way the `sflight`, but with the way the [cloud-cap-samples](https://github.com/SAP-samples/cloud-cap-samples) repository is setup it is not possible to install it like is done with `sflight` as `npm` does not support installing sub directories of GitHub repositories.

Magoli1:
Hey @BobdenOs , is there any ETA on this PR? 💯

johannes-vogel (approved):

BobdenOs (approved):

PR #426, opened 2024-01-25
fix: improve `!=` and `==` implementation for @cap-js/hana

BobdenOs:


hana/lib/HANAService.js
```diff
@@ -720,27 +720,61 @@ class HANAService extends SQLService {
...
 
+    get is_distinct_from_() { return '!=' }
+    get is_not_distinct_from_() { return '==' }
+
+    /**
+     * Checks if the xpr is a comparison or a value
+     * @param {} xpr 
+     * @returns 
+     */
+    comparerator({ xpr }, start) {
```
sjvans, line 755:
```suggestion
    comparator({ xpr }, start) {
```
or maybe `is_comparator` based on the docs comment?

sjvans (approved):
lgtm

hana/lib/HANAService.js
```diff
@@ -636,19 +636,22 @@ class HANAService extends SQLService {
...
+      return this.where(xpr)
     }
 
-    xpr({ xpr, _internal }, caseSuffix = '') {
+    xpr(_xpr, caseSuffix = '') {
+      const { xpr, _internal } = _xpr
       // Maps the compare operators to what to return when both sides are null
       const compareOperators = {
-        '=': true,
+        '==': true,
```
danjoa, line 650: This would mean to change what we decided and specified?

hana/lib/HANAService.js
```diff
@@ -636,19 +636,22 @@ class HANAService extends SQLService {
     }
 
     where(xpr) {
-      return this.xpr({ xpr }, ' = TRUE')
+      return this.xpr({ xpr: [...xpr, 'THEN'] }).slice(0, -4)
```
danjoa, line 639: Hard to understand → What is that doing?

hana/lib/HANAService.js
BobdenOs: this change makes it as followed:

| xpr | sql |
| --- | --- |
| A = NULL | A IS NULL |
| A = B | A = B |
| A == NULL | A IS NULL |
| A == B | (A = B \|\| A IS NULL AND B IS NULL) |

As part of the reported issue it was determined that `A = NULL` was not being properly converted to `A IS NULL`. That is now resolved by calling `super.operator`. Which will return `IS` or `IS NOT` when applicable. Just like it is done for `SQLite`.

Hope that makes this a bit more clear.

danjoa (approved):

PR #431, opened 2024-01-31
fix(cqn2sql): $user.locale refs

danjoa:
The implementation was simply plain wrong, e.g. for query`SELECT localized.title from Books` the `$user.locale` in the infix filter of the `localized` association was translated to `session_context('$user.id')` instead of `session_context('$user.locale')` 

→ see the added tests

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -805,9 +805,8 @@ class CQN2SQLRenderer {
    */
   ref({ ref }) {
     switch (ref[0]) {
-      case '$now': return this.func({ func: 'session_context', args: [{ val: '$now', param: false }] })
-      case '$user':
-      case '$user.id': return this.func({ func: 'session_context', args: [{ val: '$user.id', param: false }] })
+      case '$now':  return this.func({ func: 'session_context', args: [{ val: '$now', param: false }] }) // REVISIT: why do we need param: false here?
+      case '$user': return this.func({ func: 'session_context', args: [{ val: '$user.'+ref[1]||'id', param: false }] }) // REVISIT: same here?
       default: return ref.map(r => this.quote(r)).join('.')
     }
   }
--- a/test/scenarios/bookshop/localization.test.js
+++ b/test/scenarios/bookshop/localization.test.js
@@ -1,8 +1,51 @@
 const cds = require('../../cds.js')
 const bookshop = cds.utils.path.resolve(__dirname, '../../bookshop')
+const de = { locale: 'de' }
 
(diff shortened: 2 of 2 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -805,9 +805,8 @@ class CQN2SQLRenderer {
    */
   ref({ ref }) {
     switch (ref[0]) {
-      case '$now': return this.func({ func: 'session_context', args: [{ val: '$now', param: false }] })
-      case '$user':
-      case '$user.id': return this.func({ func: 'session_context', args: [{ val: '$user.id', param: false }] })
+      case '$now':  return this.func({ func: 'session_context', args: [{ val: '$now', param: false }] }) // REVISIT: why do we need param: false here?
+      case '$user': return this.func({ func: 'session_context', args: [{ val: '$user.'+ref[1]||'id', param: false }] }) // REVISIT: same here?
```
danjoa, line 809: @BobdenOs can you answer the REVISIT questions? → Why do we need `param: false` in here?
If there is a reason, we should know/document it, like in a comment somewhere.
If not, we should remove it.

BobdenOs: It is required to reduce the amount of `values` used inside queries. As they cannot change it is best to have them embedded into the SQL. That way the optimizer doesn't need to consider all possible values. Additionally it is not required to send the value every execution.

```suggestion
      // param: false is used to ensure the value is embedded into the SQL as it will not change
      case '$now':  return this.func({ func: 'session_context', args: [{ val: '$now', param: false }] })
      case '$user': return this.func({ func: 'session_context', args: [{ val: '$user.'+ref[1]||'id', param: false }] })
```

test/scenarios/bookshop/localization.test.js
```diff
@@ -13,7 +56,7 @@ describe('Bookshop - localization', () => {
     const default_ = await SELECT.localized.from('AdminService.Books', {ID: 201}).columns('ID', 'title')
     expect(default_.title).to.be.eq('Wuthering Heights')
 
-    await cds.tx({ locale: 'de' }, async (tx) => {
+    await cds.tx(de, async (tx) => {
```
BobdenOs, line 59: How is it possible this test worked ?

danjoa: That worked as it used the compiler-generated `localized.AdminService.Books` table, which had the correct session_context usage in.

Note: `de` = `{ locale: 'de' }`

BobdenOs (approved):

PR #432, opened 2024-01-31
wip: Universal CSN

danjoa:
... continued :)


danjoa:
## Streamlining Localized Views

With commit https://github.com/cap-js/cds-dbs/pull/432/commits/e04ee292ccf1fecad03585f86bfb7a38896f64b7 I reduced the proxies to localized views added to the CSN as follows by ignoring all transient ones, like Author → Books.

db-service/lib/cqn4sql.js
```diff
@@ -991,7 +991,7 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
    */
   function getElementForRef(ref, def) {
     return ref.reduce((prev, res) => {
-      return prev?.elements?.[res] || prev?._target?.elements[res]
+      return (prev?.elements || prev?.foreignKeys)?.[res] || prev?._target?.elements[res] // PLEASE REVIEW: should we add the .foreignKey check here for the non-ucsn case?
```
patricebender, line 994: We don't need it, this function resolve also non-fk steps with `prev?._target?.elements[res]`, as far as I can tell, it shouldn't hurt to have it. We could then shortcut to the foreign key and not have to look into the assocs target.

db-service/lib/infer/index.js
```diff
@@ -1134,15 +1134,10 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
  * Returns true if e is a foreign key of assoc.
  * this function is also compatible with unfolded csn (UCSN),
  * where association do not have foreign keys anymore.
```
patricebender, line 1136: comment is outdated

patricebender (approved):
changes in my area of expertise LGTM 💯

PR #434, opened 2024-01-31
fix: deep delete

larslutz96:
Support deep delete when navigation is not included in the view.

### DIFF
--- a/db-service/lib/SQLService.js
+++ b/db-service/lib/SQLService.js
@@ -1,7 +1,7 @@
 const cds = require('@sap/cds/lib'),
   DEBUG = cds.debug('sql|db')
 const { Readable } = require('stream')
-const { resolveView, getDBTable } = require('@sap/cds/libx/_runtime/common/utils/resolveView')
+const { resolveView, getDBTable, getTransition } = require('@sap/cds/libx/_runtime/common/utils/resolveView')
 const DatabaseService = require('./common/DatabaseService')
 const cqn4sql = require('./cqn4sql')
 const { getEnabledCategories } = require('trace_events')
@@ -173,9 +173,15 @@ class SQLService extends DatabaseService {
     // REVISIT: It's not yet 100 % clear under which circumstances we can rely on db constraints
     return (super.onDELETE = /* cds.env.features.assert_integrity === 'db' ? this.onSIMPLE : */ deep_delete)
     async function deep_delete(/** @type {Request} */ req) {
-      if(true){ // first
-        const cqn = this.cqn4sql(req.query, req.data)
-        req.query.DELETE.from.ref[0] = { id: getDBTable(req.target).name, where: cqn.DELETE.where }
+      const transitions = getTransition(req.target, this)
+      if (transitions.target !== transitions.queryTarget) {
+        const query = DELETE.from({ ref: [
+          {
+            id: transitions.target.name,
+            where: [{list:Object.keys(transitions.target.keys || {}).map(k => ({ref:[k]}))},'in',SELECT.from(DELETE.from).where(DELETE.where)]
+          }
+        ] })
+        return this.onDELETE({query, target: query.target})
       }
       const table = getDBTable(req.target)
       const {compositions} = table
--- a/db-service/lib/SQLService.js
+++ b/db-service/lib/SQLService.js
@@ -1,9 +1,10 @@
 const cds = require('@sap/cds/lib'),
   DEBUG = cds.debug('sql|db')
 const { Readable } = require('stream')
-const { resolveView } = require('@sap/cds/libx/_runtime/common/utils/resolveView')
+const { resolveView, getDBTable } = require('@sap/cds/libx/_runtime/common/utils/resolveView')
 const DatabaseService = require('./common/DatabaseService')
 const cqn4sql = require('./cqn4sql')
+const { getEnabledCategories } = require('trace_events')
 
 const BINARY_TYPES = {
   'cds.Binary': 1,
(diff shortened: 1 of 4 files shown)


test/compliance/resources/srv/index.cds
```diff
@@ -0,0 +1,17 @@
+using {complex as my} from '../db/complex';
```
BobdenOs, line 1: You don't need a service. Please just add these projections into the data model.

BobdenOs (changes requested):

db-service/lib/SQLService.js
```diff
@@ -192,7 +192,26 @@ class SQLService extends DatabaseService {
...
+              id: transitions.target.name,
+              where: [
+                { list: Object.keys(transitions.target.keys || {}).map(k => ({ ref: [k] })) },
+                'in',
+                SELECT.from(req.query.DELETE.from).columns(matchedKeys).where(req.query.DELETE.where),
+              ],
+            },
+          ],
+        })
+        return this.onDELETE({ query })
```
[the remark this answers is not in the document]
larslutz96: yes you are right, but we need the `transitions.target.keys` in the `list` and the `transitions.queryTarget.keys` in the columns due to renaming keys (see `bookshop/delete.test.js test: 'Delete with path expressions'`).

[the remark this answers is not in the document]
BobdenOs: Using the mapping provided from the transitions makes that an easy fix.

The issue with the reduced / no keys entities is still not resolved.

db-service/lib/SQLService.js
```diff
@@ -192,7 +192,27 @@ class SQLService extends DatabaseService {
...
+      if (transitions.target !== transitions.queryTarget) {
+        const matchedKeys = Object.keys(transitions.queryTarget.keys)
+          .filter(key => transitions.mapping.has(key))
+          .map(k => ({ ref: [k] }))
+        const query = DELETE.from({
+          ref: [
+            {
+              id: transitions.target.name,
+              where: [
+                { list: matchedKeys.map(k => transitions.mapping.get(k.ref[0])) },
```
BobdenOs, line 205: the map result still has to be `{ref:[...]}`.

larslutz96: it returned `{ref:[...]}` correctly but in the case of `IsActiveEntity` we got an error because it could not be found in the select statement

db-service/lib/SQLService.js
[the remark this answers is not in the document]
larslutz96: @BobdenOs i fixed the issue mentioned above for `sqlite`. But for ´hana´ there seems to be a problem with such modelling where the keys are missing from the projection like:
```cds
entity Root {
  key ID       : Integer;
      fooRoot  : String;
      children : Composition of many Child
                   on children.parent = $self;
}

entity RootPNoKeys     as
  projection on Root {
    fooRoot,
    children
  }
```
i get `invalid identifier: "TB42EC190DD05D308DC68898C43E1E549812E3FB8"."(DO statement)": line 104 col 1 (at pos 2806): association join condition: ID is invalid`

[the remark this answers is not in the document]
larslutz96: deleted the test due to compliler error in cds file: `Foreign key “ID” has not been found in target “MyService.RootPNoKeys”`

db-service/lib/SQLService.js
```diff
@@ -199,7 +199,32 @@ class SQLService extends DatabaseService {
...
-      let { compositions } = req.target
+      const transitions = getTransition(req.query.target, this)
+      if (transitions.target !== transitions.queryTarget) {
+        const elements = transitions.queryTarget.keys
+          ? Object.keys(transitions.queryTarget.keys)
+          : Object.keys(transitions.queryTarget.elements).filter(
+              key => !transitions.queryTarget.elements[key].isAssociation,
+            )
+        const matchedKeys = elements
+          .filter(key => key !== 'IsActiveEntity' && transitions.mapping.has(key))
```
BobdenOs, line 210: `key !== 'IsActiveEntity'` doesn't look right. It should probable be a check that the element is `virtual`.

db-service/lib/SQLService.js
BobdenOs: Does this line produce `[{ref:[<col>]},...]` or does this produce `[<col>,...]` ?

BobdenOs (commented):
Looks good. Still very compact solution.

larslutz96:
@BobdenOs 
Johannes mentioned a scenario:
```cds
service S {
  entity RootP as projection on Root {
        key ID,
        children
    };

   entity ChildP as projection on Child where x = 1
}
```
`DELETE.from(RootP(5))` deletes root with id 5 and all children. Johannes said:
I think that this is not what we want. The compiler redirects the composition of RootP to the filtered projection ChildP. Hence, I believe we must respect the modeled where clause and must not delete children which are excluded by the projection.

It also contradicts the behavior of deep update where we compare the data on projection level which respects the where clause. See https://github.tools.sap/cap/cdsnode/issues/1711#issuecomment-5457862

patricebender (commented):
👍 before merging, would you please update the PR title and explain your approach in the PR description in more detail? Thanks!

db-service/lib/SQLService.js
larslutz96: @BobdenOs 
`matchedKeys.map(k => transitions.mapping.get(k.ref[0])) ` returns `[ { ref: [ 'ID' ] } ]`

db-service/lib/SQLService.js
```diff
@@ -201,12 +201,20 @@ class SQLService extends DatabaseService {
...
-          : Object.keys(transitions.queryTarget.elements).filter(
-              key => !transitions.queryTarget.elements[key].isAssociation,
-            )
-        const matchedKeys = elements.filter(key => transitions.mapping.has(key)).map(k => ({ ref: [k] }))
+        const keys = []
+        const transitionsTarget = transitions.queryTarget.keys || transitions.queryTarget.elements
+        for (const key in transitionsTarget) {
+          let isKeyValid
+          if (transitions.queryTarget.keys) {
+            isKeyValid = !transitionsTarget[key].virtual
```
BobdenOs, line 209: It is good to see that this code was re written as it exposed the flaw of the original implementation.

It is not possible to select `virtual` and `association` fields as they don't exist on the database. So no matter whether it is the `keys` or the `elements` being processed they will have to be removed from the `keys`.

To prevent these issues from causing bugs a PR was created to unify this check, but it currently cannot be merged as it breaks the `UPSERT` behavior. So please take over the check as defined [here](https://github.com/cap-js/cds-dbs/pull/408/files#diff-d5bbedeb9d69ec4e9a2a2cb63e32f88b5098749d0e2505bd8fefc6b6d33cf3eaR959)

BobdenOs (changes requested):

db-service/lib/SQLService.js
larslutz96: Change this code according to your suggestion

BobdenOs (approved):

danjoa:
Had to revert this commit as it broke our internal pipelines du to the incompatible override of `db.exists()`. 
I'd also like to have a review of the implementation: I'm not in favour of reusing more internal `@sap/cds` functions

PR #435, opened 2024-01-31
fix: always generate unique subquery aliases

patricebender:
if we have a query hierarchy, make sure that subqueries always have aliases which are unique to the outer scope. This does not yield more functionality for a user, since if we would want to address elements of an outer query  (i.e. make use of correlated subqueries), one has to define an alias beforehand.

### DIFF
--- a/db-service/test/cqn2sql/__snapshots__/expression.test.js.snap
+++ b/db-service/test/cqn2sql/__snapshots__/expression.test.js.snap
@@ -55,7 +55,7 @@ exports[`expressions with complex xpr 1`] = `
 
 exports[`expressions with exists 1`] = `
 {
-  "sql": "SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE exists (SELECT Foo2.name FROM Foo2 as Foo2) or not exists (SELECT Foo2.name FROM Foo2 as Foo2)",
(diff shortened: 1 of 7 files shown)


stewsk (approved):

db-service/test/cqn2sql/__snapshots__/expression.test.js.snap
```diff
@@ -55,7 +55,7 @@ exports[`expressions with complex xpr 1`] = `
 
 exports[`expressions with exists 1`] = `
 {
-  "sql": "SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE exists (SELECT Foo2.name FROM Foo2 as Foo2) or not exists (SELECT Foo2.name FROM Foo2 as Foo2)",
+  "sql": "SELECT Foo.ID,Foo.a,Foo.b,Foo.c,Foo.x FROM Foo as Foo WHERE exists (SELECT Foo2.name FROM Foo2 as Foo2) or not exists (SELECT Foo22.name FROM Foo2 as Foo22)",
```
patricebender, line 58: this `Foo22` looked wrong at a first glimpse. However, as we `SELECT FROM Foo2` the 2nd `Foo2` alias becomes `Foo22`, so all good :)

PR #438, opened 2024-02-02
fix(`cqn4sql`): only transform list if necessary

patricebender:
because of odata special magic, a tokenstream with `length === 1` which only has a value, is transformed into a full key tuple comparison, e.g. to make this possible: `select from bookshop.Books[42]`. If we now have a list, with exactly one entry, the same logic kicked in and transformed the single value in the token stream to such a key comparison.

to fix this, we now skip processing of lists which only consist of values, which circumvents the described issue.

### DIFF
--- a/db-service/test/cqn4sql/DELETE.test.js
+++ b/db-service/test/cqn4sql/DELETE.test.js
@@ -107,6 +107,82 @@ describe('DELETE', () => {
     expect(query.DELETE).to.deep.equal(expected.DELETE)
   })
 
+  it('dont mess up everything', () => {
+    const query = {
+      DELETE: {
+        from: {
+          ref: [
+            {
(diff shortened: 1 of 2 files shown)


db-service/test/cqn4sql/DELETE.test.js
```diff
@@ -107,6 +107,82 @@ describe('DELETE', () => {
     expect(query.DELETE).to.deep.equal(expected.DELETE)
   })
 
+  it('dont mess up everything', () => {
```
patricebender, line 110:
```suggestion
  it('in a list with exactly one val, dont transform to key comparison', () => {
```

BobdenOs (approved):

sjvans (approved):
lgtm

PR #439, opened 2024-02-02
fix: Never use entity._isDraftEnabled or alike

danjoa:
... as all these are fragile internal APIs!


hana/lib/HANAService.js
```diff
@@ -393,7 +393,9 @@ class HANAService extends SQLService {
                 expands[this.column_name(x)] = x.SELECT.one ? null : []
 
                 const parent = src
-                x.element._foreignKeys.forEach(k => {
+                let fkeys = x.element._foreignKeys()
```
danjoa, line 396:
```suggestion
                let fkeys = x.element._foreignKeys
```

PR #446, opened 2024-02-07
fix: Retain Error object for unique constraint violation

BobdenOs:


patricebender (dismissed):
please add a test :)

johannes-vogel:
@BobdenOs I think we need to apply the same logic in hana service as well.

johannes-vogel (approved):

PR #453, opened 2024-02-13
fix: INSERT entries containing undefined values

etimr:
`undefined` values in `INSERT.entries ` were previously replaced with `null` leading to the problem with fields with `default` values.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -479,10 +479,11 @@ class CQN2SQLRenderer {
 
           buffer += '"'
         } else {
+          if (val === undefined) continue
           if (elements[key]?.type in BINARY_TYPES) {
             val = transformBase64(val)
           }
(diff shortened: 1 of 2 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -479,10 +479,11 @@ class CQN2SQLRenderer {
 
           buffer += '"'
         } else {
+          if (val === undefined) continue
           if (elements[key]?.type in BINARY_TYPES) {
             val = transformBase64(val)
           }
-          buffer += `${keyJSON}${val === undefined ? 'null' : JSON.stringify(val)}`
+           buffer += `${keyJSON}${JSON.stringify(val)}`
```
BobdenOs, line 486:
```suggestion
          buffer += `${keyJSON}${JSON.stringify(val)}`
```

BobdenOs (approved):

johannes-vogel (approved):

PR #455, opened 2024-02-14
fix: fill in keys from where also for CREATE

patricebender:
for a request like 
```
.post(`/Orders(ID=${rootID},IsActiveEntity=true)/items`)
```
we need to propagate `rootID` to the `items` insert as foreign key.

fixes `'cds/tests/_runtime/odata/__tests__/integration/draft-create.test.js',` for `ucsn2`


patricebender:
The `req.data` should already include the foreign key... This change does not fix the root cause

patricebender:
real fix done in cds, problem was `@core.computed` propagation

PR #463, opened 2024-02-14
fix: ignore virtual keys in UPSERT conflict detection

johannes-vogel:
UPSERTs for draft enabled entities do have IsActiveEntity as virtual key. The virtuality was ignored here and led to invalid SQL statements.


johannes-vogel:
@BobdenOs could you please double check the problem with hana?

BobdenOs (approved):

PR #464, opened 2024-02-14
feat: support hybrid access for universal csn

patricebender:
universal csn  makes it possible to use structured **and** flat access to elements of the underlying entity.

the non-enumerable elements of an entity are marked as "already unfolded". This marker can be used to skip those elements from e.g. wildcard expansion but makes it possible to directly access them in the query.


patricebender:
looks like the topic is - for now- not further pursued, so this PR is on hold

PR #465, opened 2024-02-14
fix( `infer`): unique alias for scoped subqueries

patricebender:
If a unique alias is calculated for a scoped subquery (`from.ref.length > 1`), we must also assign the alias to the last `$refLink` so that it is considered during `cqn4sql` rewriting.


db-service/lib/infer/index.js
```diff
@@ -114,6 +114,8 @@ function infer(originalQuery, model = cds.context?.model || cds.model) {
         (ref.length === 1 ? first.match(/[^.]+$/)[0] : ref[ref.length - 1].id || ref[ref.length - 1])
       if (alias in querySources) throw new Error(`Duplicate alias "${alias}"`)
       querySources[alias] = target
+      const last = from.$refLinks.at(-1)
+      last.alias = alias
```
patricebender, line 118: this is the change, rest is formatting

stewsk (approved):

PR #466, opened 2024-02-16
fix: select without columns from unknown entity

BobdenOs:


hana/lib/HANAService.js
```diff
@@ -249,7 +254,7 @@ class HANAService extends SQLService {
       this.temporary = this.temporary || []
       this.temporaryValues = this.temporaryValues || []
 
-      const { limit, one, orderBy, expand, columns, localized, count, parent } = q.SELECT
+      const { limit, one, orderBy, expand, columns = ['*'], localized, count, parent } = q.SELECT
```
johannes-vogel, line 257: Is there a reason we do not need this for sqlite?

PR #468, opened 2024-02-18
fix(ucsn): Support $self.<element> in on conditions

danjoa:
In contrast to 4odata, on conditions are not rewritten in universal csn. 
In particular `on foo.id = $self.foo_id` is not rewritten to `on foo.id = foo_id` as in 4odata. 

@patricebender : **please double-check** whether this fix is correct. 
Find info about fixed tests in jest.config-usn.js.


patricebender (approved):
this fix is fine for now, I identified minor issues with this kind of on-condition in special cases and will follow up to this with some tests and improvements 👍

PR #469, opened 2024-02-19
fix: Allow SELECT.join queries again with full infer call

BobdenOs:
Currently the `@sap/cds` infer does not support `SELECT.join` queries. The `Query.target` for `join` queries always returns `undefined`. So when trying to identify `unresolved` queries the output converters where no longer being applied to `join` queries.

There have been discussions about adding `join` support to the `@sap/cds` infer function. If this will be available soon it might already solve this issue. If this still takes longer it might be required to add this quick fix.

### DIFF
--- a/db-service/lib/SQLService.js
+++ b/db-service/lib/SQLService.js
@@ -114,6 +114,9 @@ class SQLService extends DatabaseService {
    * @type {Handler}
    */
   async onSELECT({ query, data }) {
+    if (!query.target) {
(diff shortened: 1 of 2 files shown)


patricebender (approved):
fine by me, can be changed once `cds.infer` supports this 👍

johannes-vogel:
We can merge now to unblock stakeholders, but eventually we should have a better solution.
@BobdenOs Is it possible to create a test case out of the issue?

PR #471, opened 2024-02-19
fix(`cqn4sql`): be robust against `$self.<element>` references

patricebender:
if a `ref` of length > 1, has `ref[0] === 'self'`, the `$self` can just be ignored. `$self` comparisons have only special semantics if compared to an association. With this change, we become more robust in that regard and do not prepend wrong table aliases for the case `$self.<assoc>.<element>` - in that case, we must ensure that the alias of the association is used and not the alias of the entity. Follow up #468

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -918,7 +918,7 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
       if (q.SELECT.from.uniqueSubqueryAlias) return
       const last = q.SELECT.from.ref.at(-1)
       const uniqueSubqueryAlias = inferred.joinTree.addNextAvailableTableAlias(
-        getLastStringSegment(last.id||last),
+        getLastStringSegment(last.id || last),
         originalQuery.outerQueries,
       )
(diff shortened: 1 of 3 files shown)


db-service/lib/cqn4sql.js
```diff
@@ -1813,8 +1814,12 @@ function cqn4sql(originalQuery, model = cds.context?.model || cds.model) {
...
           i += wrapInXpr ? 1 : backlinkOnCondition.length // skip inserted tokens
         } else if (lhs.ref) {
-          if (lhs.ref[0] === '$self') result[i].ref.splice(0, 1, targetSideRefLink.alias)
-          else if (lhs.ref.length > 1) {
+          if (lhs.ref[0] === '$self') { // $self in ref of length > 1
+            // if $self is followed by association, the alias of the association must be used
+            if (lhs.$refLinks[1].definition.isAssociation) result[i].ref.splice(0, 1)
+            // otherwise $self points to the entity itself
+            else result[i].ref.splice(0, 1, targetSideRefLink.alias)
+          } else if (lhs.ref.length > 1) {
```
patricebender, line 1822: relevant changes, other stuff is formatting

stewsk (approved):

PR #472, opened 2024-02-20
fix: Throw 'new Error' instead of string on $search with multiple words

nils:
This patch improves the error handling for #127 by showing a slightly better understandable error message to the user.

Before:
image

After:
image

Seems like CAP expects an Error instance these days, because it accesses error.message.
`Parameter 'error.message' must be type of 'string'`

# CDS setup
@sap/cds: 7.3.1
@sap/cds-compiler: 4.3.2
@sap/cds-dk (global): 7.3.2
@sap/cds-fiori: 1.1.0
@sap/cds-foss: 4.0.2
@sap/cds-mtxs: 1.12.1
@sap/eslint-plugin-cds: 2.6.3
Node.js: v18.19.1
cds-dbs: 1.3.1

### DIFF
--- a/db-service/lib/cql-functions.js
+++ b/db-service/lib/cql-functions.js
@@ -21,7 +21,7 @@ const StandardFunctions = {
    * @returns {string}
    */
   search: function (ref, arg) {
-    if (!('val' in arg)) throw `Only single value arguments are allowed for $search`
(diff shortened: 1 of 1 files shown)


johannes-vogel (approved):

johannes-vogel:
@nils Thanks for the PR. It should definitely be errors that we throw.

PR #484, opened 2024-02-28
fix: `INSERT` with first `undefined` value

etimr:
`INSERT({ stock: undefined, ID: 223, title: 'Harry Potter' }).into(Books)` had led to `SqliteError: malformed JSON`

### DIFF
--- a/test/scenarios/bookshop/insert.test.js
+++ b/test/scenarios/bookshop/insert.test.js
@@ -16,8 +16,8 @@ describe('Bookshop - Insert', () => {
 
   test('insert with undefined value works', async () => {
     const { Books } = cds.entities('sap.capire.bookshop')
-    const insert = INSERT({ stock: undefined, ID: 223, title: 'Harry Potter' }).into(Books)
-    const resp = await cds.run(insert)
-    expect(resp.results[0].changes).to.be.eq(1)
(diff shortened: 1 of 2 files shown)


BobdenOs (approved):

test/scenarios/bookshop/insert.test.js
```diff
@@ -13,4 +13,11 @@ describe('Bookshop - Insert', () => {
     // expect(err instanceof Error).to.be.true
     expect(err.message).to.be.eq('ENTITY_ALREADY_EXISTS')
   })
+
+  test('insert with undefined value works', async () => {
+    const { Books } = cds.entities('sap.capire.bookshop')
+    const insert = INSERT({ stock: undefined, ID: 223, title: 'Harry Potter' }).into(Books)
+    const resp = await cds.run(insert)
+    expect(resp.results[0].changes).to.be.eq(1)
```
johannes-vogel, line 21: can we use the insert result here?

BobdenOs: Should work something like this:

```suggestion
    const { Books } = cds.entities('sap.capire.bookshop')
    const resp = await INSERT({ stock: undefined, ID: 223, title: 'Harry Potter' }).into(Books)
    expect(resp | 0).to.be.eq(1)
```

johannes-vogel (approved):

PR #488, opened 2024-02-29
feat(hana): Add views with parameters support for HANA

BobdenOs:
Added support for selecting from views with parameters.

Examples:
```js
SELECT.from({ref:[{
  id: view,
  args: {
    arg1: {val: 0},
    arg2: CXL`1 + 1`,
    arg3: SELECT`min(col)`.from(view),
    ...
  }
}]})
```

Changes from the `@sap/hana` service.

1. Added support for all valid `expr` for arguments as defined in the `CQN` type
1. Added support for association / compositions (does not compile)
    - Expanding associations / compositions as columns 
    - Path expressions inside the `from` clause

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -1542,7 +1542,7 @@ function cqn4sql(originalQuery, model) {
         const nextStep = refReverse[i + 1] // only because we want the filter condition
 
         if (stepLink.definition.target && nextStepLink) {
-          const { where } = nextStep
+          const { where, args } = nextStep
           if (isStructured(nextStepLink.definition)) {
             // find next association / entity in the ref because this is actually our real nextStep
             const nextStepIndex =
@@ -1563,7 +1563,7 @@ function cqn4sql(originalQuery, model) {
             as = getNextAvailableTableAlias(as)
           }
           nextStepLink.alias = as
-          whereExistsSubSelects.push(getWhereExistsSubquery(stepLink, nextStepLink, where))
+          whereExistsSubSelects.push(getWhereExistsSubquery(stepLink, nextStepLink, where, false, args))
         }
       }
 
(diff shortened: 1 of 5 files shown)


db-service/lib/cqn4sql.js
```diff
@@ -1563,7 +1561,7 @@ function cqn4sql(originalQuery, model) {
             as = getNextAvailableTableAlias(as)
           }
           nextStepLink.alias = as
-          whereExistsSubSelects.push(getWhereExistsSubquery(stepLink, nextStepLink, where))
+          whereExistsSubSelects.push(getWhereExistsSubquery(stepLink, nextStepLink, where, false, args))
```
patricebender, line 1564: I'm not entirely sure but I can't think of a way to define a scoped query which ends in a parameterized view, the compiler does not allow associations targeting entities with parameters:
```
Error[def-unexpected-paramview-assoc]: Expected association target to have no parameters
    |
  e.cds:23:9, at entity:“Scoped”/element:“pbooks”
    |
 23 |         pbooks
    |         ^
```

patricebender (approved):

hana/test/param-views.test.js
```diff
@@ -0,0 +1,53 @@
...
+      SELECT.from({ ref: [root] }),
+      SELECT.from({ ref: [root, 'author'] }),
+      SELECT`ID,stock,author{ID}`.from({ ref: [root] }),
+    ])
+
+    expect(booksRes).to.have.property('length').to.be.eq(books)
+    const authorKeys = expandRes.map(r => r.author.ID)
+    expect(authorsRes.filter(r => authorKeys.includes(r.ID))).to.have.property('length').to.be.eq(authorsRes.length)
+  })
+})
```
patricebender, line 53:
```suggestion
})

```

PR #489, opened 2024-03-01
feat: support path expressions for parameterized entities

patricebender:
consider `args` in a `ref` for path expression. If no arguments are provided, but we navigate along a parameterized entity, we provide an empty argument list (same for `@cds.persistence.udf`)
 
→This is in line with the compiler behavior

### DIFF
--- a/db-service/test/cqn4sql/with-parameters.test.js
+++ b/db-service/test/cqn4sql/with-parameters.test.js
@@ -74,7 +74,7 @@ describe('entities and views with parameters', () => {
       const query = cqn4sql(cqn, model)
       const expected = CQL`
       SELECT FROM PBooks(P1: 42, P2: 45) as PBooks
-      left join Authors(P1: 1) as author on author.ID = PBooks.author_ID
+      left join Authors(P1: dummy) as author on author.ID = PBooks.author_ID
         {
           author.name as author
         }
--- a/db-service/test/cqn4sql/with-parameters.test.js
+++ b/db-service/test/cqn4sql/with-parameters.test.js
@@ -8,39 +8,40 @@ const { SELECT } = require('@sap/cds/lib/ql/cds-ql')
 const cqn4sql = require('../../lib/cqn4sql')
 const cds = require('@sap/cds/lib')
 const { expect } = cds.test
-describe('Repetitive calls to cqn4sql must work', () => {
+describe('entities and views with parameters', () => {
   let model
   beforeAll(async () => {
     model = await cds.load(__dirname + '/model/withParameters').then(cds.linked)
   })
 
-  it('select from view with param', () => {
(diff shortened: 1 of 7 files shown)


db-service/test/cqn4sql/with-parameters.test.js
```diff
@@ -8,32 +8,191 @@ const { SELECT } = require('@sap/cds/lib/ql/cds-ql')
...
+      expect(query).to.deep.equal(expected)
+    })
+    it('empty argument list if no params provided for association', () => {
+      const cqn = CQL`SELECT from PBooks(P1: 42, P2: 45) {
+            author.name as author,
+    }`
+      const query = cqn4sql(cqn, model)
+      const expected = CQL`
+      SELECT FROM PBooks(P1: 42, P2: 45) as PBooks
+      left join Authors(P1: 1) as author on author.ID = PBooks.author_ID
```
stewsk, line 77: Maybe replace `P1 : 1` by `P1 : dummy` like below to indicate that this is only there to trick the compiler (parser).

db-service/test/cqn4sql/with-parameters.test.js
```diff
@@ -8,32 +8,191 @@ const { SELECT } = require('@sap/cds/lib/ql/cds-ql')
...
-    // currently only possible with cds-compiler beta-mode,
-    // as the view with params does not yet support associations
-    const query = cqn4sql(SELECT.from('PBooks(P1: 1, P2: 2)').columns('author.name as author'), model)
-    const expected = CQL`SELECT FROM PBooks(P1: 1, P2: 2) as PBooks
-                         left join Authors as author on author.ID = PBooks.author_ID {
-                          author.name as author
-                         }`
-    expect(query).to.deep.equal(expected)
+
+  describe('subqueries', () => {
```
stewsk, line 188: Does HANA allow a subquery at this location?

stewsk (approved):

db-service/test/cqn4sql/with-parameters.test.js
patricebender: It doesnt work on my HANA Cloud:

```sql
-- generated by cds-compiler version 4.7.7
CREATE TABLE Books (
  ID INTEGER NOT NULL,
  PRIMARY KEY(ID)
);

-- generated by cds-compiler version 4.7.7
CREATE VIEW P(IN P1 INTEGER) AS SELECT
  Books_0.ID
FROM Books AS Books_0
WHERE :P1 = Books_0.ID;

INSERT INTO Books (ID) VALUES (1);

SELECT * FROM P((SELECT 1 FROM DUMMY));
```

```
SQL Error [7] [HY000]: SAP DBTech JDBC: [7] (at 16): feature not supported: cannot support scalar subquery as input of parameterized sql view: line 1 col 17 (at pos 16)
```

PR #492, opened 2024-03-04
feat(orderby): allow to disable collations with @cds.collate: false

David-Kunz:
Disable order-by collations of string columns with `@cds.collate: false`


ThePlenkov:
@David-Kunz hi! Sorry for tagging in the closed ticket- but I just found that after switching to cap-js/hana my queries from a flat table like BKPF turned out into something like this:
```
SELECT *,
('$[' || lpad("$$RN$$", 6, '0')) as "_path_"
FROM (
        SELECT *,
            ROW_NUMBER() OVER (
                ORDER BY "MANDT" COLLATE ENGLISH ASC,
                    "BUKRS" COLLATE ENGLISH ASC,
                    "BELNR" COLLATE ENGLISH ASC,
                    "GJAHR" COLLATE ENGLISH ASC
            ) as "$$RN$$"
        FROM (
                SELECT "BKPF"."MANDT",
                    "BKPF"."BUKRS",
                    "BKPF"."BELNR",
                    "BKPF"."GJAHR",
                    "BKPF"."BLART",
                    "BKPF"."BLDAT",
                    "BKPF"."BUDAT",
                    "BKPF"."MONAT",
                    "BKPF"."CPUDT",
                    "BKPF"."CPUTM",
 ...
                FROM "api.odata.BKPF" as "BKPF"
                ORDER BY "BKPF"."MANDT" COLLATE ENGLISH ASC,
                    "BKPF"."BUKRS" COLLATE ENGLISH ASC,
                    "BKPF"."BELNR" COLLATE ENGLISH ASC,
                    "BKPF"."GJAHR" COLLATE ENGLISH ASC
                LIMIT 10
            ) as "BKPF"
    ) as "BKPF"
```

To be honest I am little bit confused and concerned about this functionality.

Isn't it supposed to be something opposite - like collate is switched off global wise and enabled only when is really needed with cds.collate: true?

However even like this - it's not clear how to use it.  Should we mention every single field to collate/not collate?

Thanks!

danjoa (commented):
Why is that required? Is that a public feature? Also supported equally in Java?

ThePlenkov:
Well from my side what is required - is to be able to control collation logic globally ( disable ) using @cap-js/hana module. It's described here https://github.com/cap-js/cds-dbs/issues/965

ThePlenkov:
Yes it's a public documented feature, however provided only in Java not JS

ThePlenkov:
Yes it's a public documented feature, however provided only in Java not JS

PR #496, opened 2024-03-05
fix: deep delete for views

BobdenOs:


larslutz96 (approved):

johannes-vogel:
I double checked with our internal stakeholder tests and observe some failures that are related.
Before merging, we should double check these tests.

I can help in reproducing.

danjoa:
Our code explodes ... wonder if we could do that better. Maybe it's due to how getTransitions() works. 

Btw: using these private helpers from @sap/cds is no good.

larslutz96:
> I double checked with our internal stakeholder tests and observe some failures that are related. Before merging, we should double check these tests.
> 
> I can help in reproducing.

@johannes-vogel i executed the stakeholder tests again after our fix. Only two tests are failing but they also fail with main.

johannes-vogel (approved):

PR #497, opened 2024-03-05
Tests for unresolved entities (fails on Postgres)

nils:
This is only a PR for providing **tests**.
I do not know enough about the internals of cds-dbs to fix the issue (at least without an expert providing help).

The tests fail **only** for Postgres (cd postgres && npm run test)


johannes-vogel:
Hi Nils,

thanks for the PR.
We are already aware and working on a fix for it.

nils:
@johannes-vogel we still have the issue... (see last pipeline run)
1. Is there any timeline/progress you can share on it? 
2. Maybe if the team capacity does not allow to fix it on short notice, are there any hints/preferred solutions you can share for somebody else to fix it as a PR?

johannes-vogel:
> @johannes-vogel we still have the issue... (see last pipeline run)
> 
> 1. Is there any timeline/progress you can share on it?
> 2. Maybe if the team capacity does not allow to fix it on short notice, are there any hints/preferred solutions you can share for somebody else to fix it as a PR?

It should work already with cds 7.8. I'm about to update the deps with #573 and will update this PR afterwards for proving that it works.

test/compliance/CREATE.test.js
```diff
@@ -25,7 +25,9 @@ describe('CREATE', () => {
   // Load model before test suite to generate test suite from model definition
   const model = cds.load(__dirname + '/resources/db', { sync: true })
 
-  const literals = Object.keys(model.definitions).filter(n => model.definitions[n].kind === 'entity')
+  const literals = Object.keys(model.definitions)
+    .filter(n => model.definitions[n].kind === 'entity')
+    .filter(n => n !== "complex.JoinBooksAndAuthors" )
```
BobdenOs, line 30: It seems like you are trying to exclude the more complex view. Which this test suite does not (yet) support. It would be good to make the check less specific and also catch any entities added in the future which would cause trouble for this test.

You would have to try it out, but I guess it should look something like:

```suggestion
    .filter(n => !model.definitions[n].query) // Exclude views
    .filter(n => !model.definitions[n].query?.from?.args) // Exclude views with join source
```

nils:
I don't remember what this was about and we currently don't have any issues with it, so let's hope that the cause is fixed

PR #505, opened 2024-03-05
fix: issue with reused select cqns

johannes-vogel:
We've observed a problem when a cqn is executed and reused as subselect.
Most probably this is due to our manipulation `SELECT.expand='root'` which is then also considered in the subsequent subselect and produces an incorrect subselect:

```sql
SELECT json_insert('{}', '$."name"', name, '$."title"', title) as _json_
FROM (
    SELECT a.name,
      b.title
    FROM sap_capire_bookshop_Books as b
      inner JOIN sap_capire_bookshop_Authors as a ON a.ID = b.author_ID
    WHERE b.ID in (
        SELECT json_insert('{}', '$."ID"', ID) as _json_  # this is wrong
        FROM (
            SELECT ID
            FROM sap_capire_bookshop_Books
          )
      )
    ORDER BY b.ID ASC
    LIMIT 1
  )
```


johannes-vogel:
@BobdenOs I think you need to approve yourself :)

patricebender (approved):

PR #514, opened 2024-03-05
fix: Getting rid of quirks mode

danjoa:
Optimisation for quirks mode off with cds8


BobdenOs (approved):
Looks like a good solution. As it retains the logic for skipping `cqn4sql` for unknown entities. Which is my main concern.

PR #515, opened 2024-03-06
chore: remove unused dev dependency

patricebender:


patricebender:
seems to be needed for `cds.test`...

PR #519, opened 2024-03-07
fix(cqn2sql): Smart quoting of columns inside UPSERT rows

BobdenOs:
Defer the column quoting until rendering them into the sql. To prevent them from being quoted twice.


db-service/test/cqn2sql/testModel.cds
```diff
@@ -13,6 +13,13 @@ entity Foo2 {
   virtual something : String(11);
 }
 
+entity !["Foo2Quoted"] {
+  key !["ID"]: Integer;
+  !["name"]: String;
+  !["a"]: Integer;
```
johannes-vogel, line 19: can you rename/add one of the columns to a reserved keyword as in the bug report?

BobdenOs: The root cause was not the reserved key word. As this was correctly being detected. It was just the problem `quote` was being called twice.

patricebender (approved):

db-service/test/cqn2sql/__snapshots__/upsert.test.js.snap
```diff
@@ -21,3 +21,14 @@ exports[`upsert test with keys only 1`] = `
...
 `;
+
+exports[`upsert test with rows (quoted) 1`] = `
+{
+  "entries": [
+    [
+      "[[1,null,2]]",
+    ],
+  ],
+  "sql": "INSERT INTO """Foo2Quoted""" ("""ID""","""name""","""a""") SELECT value->>'$[0]',value->>'$[1]',value->>'$[2]' FROM json_each(?) WHERE true ON CONFLICT("""ID""") DO UPDATE SET """name""" = excluded."""name""","""a""" = excluded."""a"""",
```
BobdenOs, line 32: the `db-service` tests don't have reserved words configured as these are database specific. Therefor the `order` is not quoted in this test.

db-service/test/cqn2sql/__snapshots__/upsert.test.js.snap
```diff
@@ -21,3 +21,14 @@ exports[`upsert test with keys only 1`] = `
...
 `;
+
+exports[`upsert test with rows (quoted) 1`] = `
+{
+  "entries": [
+    [
+      "[[1,null,2,3]]",
+    ],
+  ],
+  "sql": "INSERT INTO """Foo2Quoted""" ("""ID""","""name""","""a""",order) SELECT value->>'$[0]',value->>'$[1]',value->>'$[2]',value->>'$[3]' FROM json_each(?) WHERE true ON CONFLICT("""ID""") DO UPDATE SET """name""" = excluded."""name""","""a""" = excluded."""a""",order = excluded.order",
```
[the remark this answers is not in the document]
johannes-vogel: hm, then it does not have much value to add it

PR #520, opened 2024-03-07
fix: affectedRows

larslutz96:
fix: affectedRows, filtered aggregation, and recursive composition timeout.
Note: 
filtered aggregations are already fixed in the meantime by [pr](https://github.com/cap-js/cds-dbs/pull/527)
recursive composition timeout will have to be solved by changing deep update logic.


db-service/lib/deep-queries.js
```diff
@@ -35,7 +36,42 @@ async function onDeep(req, next) {
...
+            resultsArray.push(result.results[0])
+          }
+          return resultsArray
+        },
+        summedResult.results.length ? summedResult.results : [summedResult.results],
+      )
+    }
+  }
+  
+  return summedResult ? summedResult : (res[0] ?? 0) // TODO what todo with multiple result responses?
```
BobdenOs, line 74: I have a feeling this could be done a bit easier.

```suggestion
  return beforeData.length || Array.isArray(req.data) ? req.data.length : 1
```

db-service/lib/deep-queries.js
```diff
@@ -114,6 +150,24 @@ const _calculateExpandColumns = (target, data, expandColumns = [], elementMap =
       expandColumns.push(expandColumn)
     }
 
+    // for deep update check if is recursiv and has looping association --> would result in cqn4sql loop
```
BobdenOs, line 153: The check does not actually do what the comment states it does. As recursive `compositions` can have any number of `compositions` in between. I don't see a `while(true)` or recursive function call in side this check which would be required to find the cycle.

db-service/lib/deep-queries.js
```diff
@@ -114,6 +150,24 @@ const _calculateExpandColumns = (target, data, expandColumns = [], elementMap =
...
+    if(target.name === compositions[compName].target){
+      for(const compName2 in compositions){
+        if(compName2 === compName) continue
+        const innerCompDefinition = compositions[compName2]
+        const innerCompositions = innerCompDefinition?._target?.compositions
+        if(!innerCompositions || target.name === innerCompDefinition.target) continue
+        for(const innerCompName in innerCompositions){          
+          if(innerCompositions[innerCompName].target === target.name){
+            throw new Error(
```
BobdenOs, line 162: Rather then throwing an `Error`. This functionality is supposed to be supported. As we have the Bookshop genre test to make sure recursive `compositions` work. The only difference is that it uses a self recursive `composition`.

db-service/lib/deep-queries.js
larslutz96: i dont think so. The tests expects the result structured like this:
```json
{
"affectedRows": 1,
"results": [{
  "changes": 1,
  "lastInsertRowid": 8319
}],
"query": ...
}
```
Therefore we would have to sum up res.results and res.affectedRows

PR #521, opened 2024-03-07
test: Validate missing output converter for aggregate functions

BobdenOs:
Currently `cqn.elements` does not provide the correct type for functions. Therefor it is not possible to apply the correct output converter.

As example the current `DateTime` type requires an output converter to remove the additional sub second precision.
```cql
SELECT dateTime from dateTime -- returns in format 1970-01-01T00:00:00Z
SELECT min(dateTime from dateTime -- returns in format 1970-01-01T00:00:00.000Z
```

### DIFF
--- a/test/compliance/SELECT.test.js
+++ b/test/compliance/SELECT.test.js
@@ -284,8 +284,22 @@ describe('SELECT', () => {
   })
 
   describe('groupby', () => {
-    test.skip('missing', () => {
-      throw new Error('not supported')
+    test('aggregate functions', async () => {
+      const [{ min, max }] = await cds.run(CQL`
+        SELECT
+          min(dateTime),
+          max(dateTime)
+        FROM basic.projection.dateTime
+        WHERE dateTime != null
+      `)
+
+      const [{ dateTime: minMatch }, { dateTime: maxMatch }] = await Promise.all([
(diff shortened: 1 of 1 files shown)


test/compliance/SELECT.test.js
```diff
@@ -284,8 +284,22 @@ describe('SELECT', () => {
...
+        SELECT
+          min(dateTime),
+          max(dateTime)
+        FROM basic.projection.dateTime
+        WHERE dateTime != null
+      `)
+
+      const [{ dateTime: minMatch }, { dateTime: maxMatch }] = await Promise.all([
+        SELECT.one.from('basic.projection.dateTime').where(`dateTime = `, new Date(min)),
+        SELECT.one.from('basic.projection.dateTime').where(`dateTime = `, new Date(max)),
```
danjoa, line 298: Do we / everybody need to wrap these arguments into new Date()?

danjoa:
I don't get the conclusion. Is it that you can't use result values you got of min/max(x) as values for subsequent queries with where x=?

If yes: yes, that's true in general. I'd even not claim/document that wrapping them into new Date would fix that. 

The only recommendation we can give, imho, is to never do where clauses with some datetime = some value. Always only with >, >=, <, <=, because time is floating. Always.

PR #522, opened 2024-03-08
chore: fix 78fe10b

johannes-vogel:
without that fix all elements not specified in entries would be nulled.
bug introduced with 78fe10b


patricebender (approved):

johannes-vogel:
@BobdenOs Was it intentional to be removed?

patricebender:
fixes https://github.com/cap-js/cds-dbs/pull/519

BobdenOs:
Please consider merging https://github.com/cap-js/cds-dbs/pull/425 as it should have fixed all mentioned issues already. It will only break some cds tests. As it makes upset work properly.

PR #524, opened 2024-03-11
fix(search): search on aggregated results in HAVING clause 

patricebender:
Although this use case is probably not really common or makes much sense, there were some failing `cds` tests which should be fixed by this.

---

If the query contains a `group by`, the search is performed on the queries
columns. If one of those columns is an aggregate function,
we must put the search term in the `HAVING` clause instead of `WHERE`.
Since the logical processing order of SQL queries is `FROM -> WHERE -> GROUP BY -> HAVING -> SELECT -> ORDER BY` the aggregated results are not accessible in the `WHERE` clause.

This led to an error:

```sql
> Uncaught SqliteError: misuse of aggregate: COUNT() in:
SELECT
 COUNT(Books.stock) AS foo
FROM
  sap_capire_bookshop_Books AS Books
WHERE
  (IFNULL(INSTR(LOWER(COUNT(stock)), LOWER(?)), 0))
GROUP BY Books.title
```

With this change, the searchable columns of the query are
checked against the  aggregate functions of the ANSI SQL standard.
If one is found and the query contains a `group by` (which should then
always be the case), the search expression is put into the `HAVING` clause.

### DIFF
--- a/db-service/test/cqn4sql/search.test.js
+++ b/db-service/test/cqn4sql/search.test.js
@@ -120,15 +120,17 @@ describe('Replace attribute search by search predicate', () => {
       search((books.createdBy, books.modifiedBy, books.anotherText, books.title, books.descr, books.currency_code, books.dedication_text, books.dedication_sub_foo, books.dedication_dedication), ('x' OR 'y')) `,
     )
   })
-  it('Search with aggregated column and groupby', () => {
-    let query = SELECT.from('bookshop.Books')
-      .columns({ args: [{ xpr: [{ ref: ['stock'] }] }], as: 'foo', func: 'count' })
+  it('Search with aggregated column and groupby must be put into having', () => {
+    // if we search on aggregated results, the search must be put into the having clause
+    const { Books } = cds.entities
+    let query = SELECT.from(Books)
+      .columns({ args: [{ ref: ['title'] }], as: 'firstInAlphabet', func: 'MIN' })
       .groupBy('title')
-      .search('foo')
+      .search('Cat')
 
-    let res = cqn4sql(query, model)
-    expect(JSON.parse(JSON.stringify(res.SELECT.where))).to.not.deep.equal([
-      { func: 'search', args: [{ args: [{ xpr: [{ ref: ['stock'] }] }], as: 'foo', func: 'count' }, { val: 'foo' }] },
-    ])
+    expect(JSON.parse(JSON.stringify(cqn4sql(query, model)))).to.deep.equal(CQL`
+      SELECT from bookshop.Books as Books {
+        MIN(Books.title) as firstInAlphabet
+      } group by Books.title having search(MIN(Books.title), 'Cat')`)
   })
 })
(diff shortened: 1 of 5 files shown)


BobdenOs:
@patricebender this change does not make any sense to me. A `$search` is looking for a `string`. So why would we try to search inside the result of a `sum` ?

Also if the `$search` is applied after the `group by` it might create the wrong result.

patricebender:
> @patricebender this change does not make any sense to me. A `$search` is looking for a `string`. So why would we try to search inside the result of a `sum` ?
> 
> Also if the `$search` is applied after the `group by` it might create the wrong result.

This was requested by @larslutz96 because some CDS tests are failing → In the legacy stack there is such a mechanism in place. Maybe Lars can elaborate a little further.

I aggree that there are probably no real use cases for this, but at least there is no sql error anymore. And it should not break what has worked before, does it?

db-service/test/cqn4sql/search.test.js
```diff
@@ -120,4 +120,17 @@ describe('Replace attribute search by search predicate', () => {
       search((books.createdBy, books.modifiedBy, books.anotherText, books.title, books.descr, books.currency_code, books.dedication_text, books.dedication_sub_foo, books.dedication_dedication), ('x' OR 'y')) `,
     )
   })
+  it('Search with aggregated column and groupby must be put into having', () => {
+    // if we search on aggregated results, the search must be put into the having clause
+    const { Books } = cds.entities
+    let query = SELECT.from(Books)
```
BobdenOs, line 126: Please also add this test to the bookshop scenario test. To make sure this also works on HANA. As the type cast from sum to string might cause errors.

test/scenarios/bookshop/search.test.js
```diff
@@ -0,0 +1,19 @@
+const cds = require('../../cds.js')
+const bookshop = require('path').resolve(__dirname, '../../bookshop')
```
BobdenOs, line 2: Needs to be included in the bookshop/index.js file. So didn't actually run in the GitHub action.

db-service/test/cqn4sql/search.test.js
```diff
@@ -120,4 +120,17 @@ describe('Replace attribute search by search predicate', () => {
...
+      .columns({ args: [{ ref: ['title'] }], as: 'firstInAlphabet', func: 'MIN' })
```
johannes-vogel, line 127: Can you add some other aggregation to ensure that they are not searched?

patricebender: all `.funcs` are put into the search if the query result is grouped, `search.js`:
```js
if (cqn._aggregated || /* new parser */ cqn.SELECT.groupBy || cqn.SELECT?.from?.SELECT?.groupBy) {
    cqn.SELECT.columns &&
      cqn.SELECT.columns.forEach(column => {
        if (column.func) {
          // exclude $count by SELECT of number of Items in a Collection
          if (
            cqn.SELECT.columns.length === 1 &&
            column.func === 'count' &&
            (column.as === '_counted_' || column.as === '$count')
          ) {
            return
          }

          toBeSearched.push(column)
          return
        }
…
```

Only the ANSI SQL aggregations are put into the `having` clause though. Do you want me to add a test for aggregations which are _not_ put into the `having` clause?

johannes-vogel: No, I meant to add an aggregation in the test that is not string like which should be excluded, e. g. sum(price) as price

patricebender: okay sure, then I need to add a list of aggregations which must be included from being searched first.

johannes-vogel (approved):

PR #525, opened 2024-03-14
fix(`hana`): use keyword list from compiler

patricebender:
today we updated the keywords in the compiler because new hana version came with some new keywords. Instead of copying the keywords to this repo, we should rather import the list from the compiler to get rid of this double maintenance.

I already requested that also keywords for other sql dialects will be available in a similar fashion.


BobdenOs (approved):
👍 would be great for the others as well.

patricebender:
> 👍 would be great for the others as well.

added this for sqlite in #525.
With the next compiler version there will be a similar API for the postgres keywords 💯

hana/lib/HANAService.js
```diff
@@ -6,6 +6,12 @@ const { SQLService } = require('@cap-js/db-service')
 const drivers = require('./drivers')
 const cds = require('@sap/cds')
 const collations = require('./collations.json')
+const keywords = require('@sap/cds-compiler').to.hdi.keywords
```
johannes-vogel, line 9: with that we have introduced a dependency to compiler at runtime. So far we did not have that. We either need to declare it or keep the local list. If this is all we need, I'd vote for a local list.

patricebender: dont we have the dependency in any case due to the dependency to `sap/cds`?

(I think `cds` also exports the keywords, we could use the cds api instead)

johannes-vogel: If cds exports it, that's fine.
I was just thinking that we introduce a dependency to the compiler only for keywords which seems like an overkill to me.

patricebender: → #530

PR #528, opened 2024-03-14
feat: also support lowercase matchespattern function

johannes-vogel:


db-service/lib/cql-functions.js
```diff
@@ -99,6 +99,13 @@ const StandardFunctions = {
    * @returns {string}
    */
   matchesPattern: (x, y) => `(${x} regexp ${y})`,
+  /**
+   * Generates SQL statement that matches the given string against a regular expression
+   * @param {string} x
+   * @param {string} y
+   * @returns {string}
+   */
+  matchespattern: (x, y) => `(${x} regexp ${y})`,
```
BobdenOs, line 108: This function is not defined in the other databases ?

johannes-vogel: Isn't this the generic cql functions that we introduce?

BobdenOs: Yes, but these are the database specific translations into SQL. I checked and it seems that this currently only works for sqlite.

So this PR should add a test using the function and the posters and HANA versions.

johannes-vogel: Got it, will do.

BobdenOs (approved):

PR #529, opened 2024-03-14
chore: return normal results for lock queries

BobdenOs:
When using `forUpdate` or `forShareLock` now `sqlite` and `postgres` will both use normal `root` queries. While for HANA it will first execute the lock SQL and then execute the query with the lock removed. As HANA does not allow locks when using sub queries. 

Example:
```SQL
SELECT ID FROM (SELECT * FROM Books) FOR UDPATE ID NO WAIT -- cannot identify Books as the lock target table
```

When using expands in lock queries only the lock on the root table is guaranteed.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -110,26 +110,34 @@ class HANAService extends SQLService {
     if (!query.target) {
       try { this.infer(query) } catch (e) { /**/ }
     }
-    if (
-      !query.target
-      || query.target._unresolved
-      || query.SELECT.forUpdate
-      || query.SELECT.forShareLock
-    ) {
+    if (!query.target || query.target._unresolved) {
       return super.onSELECT(req)
     }
 
-    // REVISIT: disable this for queries like (SELECT 1)
-    // Will return multiple rows with objects inside
-    query.SELECT.expand = 'root'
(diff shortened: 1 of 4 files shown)


hana/lib/HANAService.js
```diff
@@ -110,26 +110,34 @@ class HANAService extends SQLService {
...
     delete query.SELECT.expand
 
     // REVISIT: add prepare options when param:true is used
-    const sqlScript = this.wrapTemporary(temporary, withclause, blobs)
+    const sqlScript = isLockQuery ? sql : this.wrapTemporary(temporary, withclause, blobs)
     let rows = (values?.length || blobs.length > 0)
       ? await (await this.prepare(sqlScript, blobs.length)).all(values || [])
       : await this.exec(sqlScript)
+
+    if (isLockQuery) {
```
johannes-vogel, line 133: should we add a comment why we need to do this?

johannes-vogel (approved):

johannes-vogel:
Should we rename the PR title as it's a fix for an unreleased feature.

hana/lib/HANAService.js
David-Kunz: @johannes-vogel , That comment would've been helpful for me ;)

PR #531, opened 2024-03-15
test: simplify hana testing

johannes-vogel:
this picks up automatically the same env variables as before.


hana/package.json
```diff
@@ -20,7 +20,8 @@
     "npm": ">=8"
   },
   "scripts": {
-    "test": "npm start && jest --silent",
+    "test": "npm start && npm run jest",
+    "jest": "jest --silent",
```
BobdenOs, line 24: This desperation is causing trouble in the pipeline as it is not able to forward the `--maxWorkers=1` argument which prevents the database isolation race condition.

BobdenOs (approved):

hana/package.json
johannes-vogel: I was only trying to find a solution to omit the start script for local runs :)

PR #533, opened 2024-03-18
fix(hana): reduce service manager calls for failing tenants

BobdenOs:
When the tenant does not exist or the service manager is rate limited. The `create` function no longer retries and waits for the `acquireTimeoutMillis` to be reached.


hana/lib/HANAService.js
```diff
@@ -60,7 +60,12 @@ class HANAService extends SQLService {
           HANAVERSION = dbc.server.major
           return dbc
         } catch (err) {
-          if (!isMultitenant || err.code !== 10) throw err
+          if (isMultitenant) {
+            // REVISIT: throw the error and break retry loop
+            // Stop trying when the tenant does not exist or is rate limited
+            if (err.status == 404 || err.status == 429)
+              return new Promise(() => { })
```
johannes-vogel, line 67: I think we need to have a timeout here as otherwise we have unresolving promises?!

hana/lib/HANAService.js
```diff
@@ -60,7 +60,12 @@ class HANAService extends SQLService {
           HANAVERSION = dbc.server.major
           return dbc
         } catch (err) {
-          if (!isMultitenant || err.code !== 10) throw err
+          if (isMultitenant) {
```
johannes-vogel, line 63: Should we drain the pool in addition as otherwise already opened connections would remain non-functional?

BobdenOs: Opened connections keep working for HANA

hana/lib/HANAService.js
johannes-vogel:
```suggestion
              return new Promise(function (_, reject) { setTimeout(() => reject(e)), this._config.acquireTimeoutMillis})
```

PR #535, opened 2024-03-18
fix: joins without columns are rejected

johannes-vogel:
sqlite service allows join CQN without specifying columns, hana service does not.
As we already documented that join CQNs must be well-formed, we have to reject these CQNs.


db-service/lib/SQLService.js
```diff
@@ -114,6 +114,11 @@ class SQLService extends DatabaseService {
    * @type {Handler}
    */
   async onSELECT({ query, data }) {
+    if (query.SELECT.from?.join && !query.SELECT.columns) {
```
BobdenOs, line 117: What is the `join` is not on root level ?

```cql
SELECT ID FROM (
  SELECT * FROM
  Books
  LEFT JOIN
  Authors
  ON
  Books.author_ID = Authors.ID
)
```

Maybe it makes sense to move it into `cqn2sql` ([code](https://github.com/cap-js/cds-dbs/blob/main/db-service/lib/cqn2sql.js#L208)). Which will reject it on all levels.

johannes-vogel: I'll double check

test/scenarios/bookshop/read.test.js
```diff
@@ -255,6 +255,40 @@ describe('Bookshop - Read', () => {
...
+          args: [
+            { ref: ['sap.capire.bookshop.Books'], as: 'b' },
+            { ref: ['sap.capire.bookshop.Authors'], as: 'a' },
+          ],
+          on: [{ ref: ['a', 'ID'] }, '=', { ref: ['b', 'author_ID'] }],
+        },
+      },
+    }
+
+    expect(cds.db.run(query)).to.be.rejectedWith(/Ambiguous wildcard elements/)
```
BobdenOs, line 272:
```suggestion
    return expect(cds.db.run(query)).to.be.rejectedWith(/Ambiguous wildcard elements/)
```

PR #537, opened 2024-03-19
feat(hana): drop prepared statements after end of transaction

SamuelBrucksch:


hana/lib/drivers/base.js
```diff
@@ -37,20 +37,58 @@ class HANADriver {
...
+          Promise.all(
+            streams.map(
+              stream =>
+                new Promise((resolve, reject) => {
+                  stream.on('end', () => {
+                    resolve()
+                  })
+                  stream.on('error', () => {
+                    // REVISIT: does it actually matter if it fails or is successful, so should we resolve here?
+                    reject()
```
BobdenOs, line 51: As the comment states this should resolve. The main reason is that it is a dangling promise. Which means if you `reject` it will go to the root of the process will crash the application.

hana/lib/drivers/base.js
```diff
@@ -37,20 +37,58 @@ class HANADriver {
...
         return { changes }
       },
       runBatch: async params => {
         const stmt = await prep
-        const changes = await prom(stmt, 'exec')(params)
+        const changes = await prom(
+          stmt,
+          'exec',
+        )(params).then(changes => {
+          stmt.drop()
```
BobdenOs, line 68: Calling drop inside these functions is not good. Rather define a new drop function on the returned value and calling it where the `prepare` was originally called. This pattern can also be put into `sqlite` even when it is just an empty function.

BobdenOs (changes requested):

BobdenOs:
@johannes-vogel Currently it is possible to return `Readable` streams from `SELECT` queries out of the transaction. With this change the `Readable` stream have to finish reading inside the transaction otherwise their prepared statement are invalidated.

This is currently inconsistent with the `SQLite` service as it doesn't have the same requirement. It will probably be required to adjust the `SQLite` implementation to validate that the original transaction is still valid.

BobdenOs (approved):

PR #545, opened 2024-03-22
chore: increase db-service version for pg,hana,sqlite

johannes-vogel:
@patricebender Do you think that works if we merge this PR just before the release?

### DIFF
--- a/hana/package.json
+++ b/hana/package.json
@@ -28,7 +28,7 @@
   },
   "dependencies": {
     "hdb": "^0.19.5",
-    "@cap-js/db-service": "^1.6.0"
+    "@cap-js/db-service": "^1.7.0"
   },
   "peerDependencies": {
     "@sap/hana-client": ">=2",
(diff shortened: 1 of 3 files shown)


patricebender:
this should work in the release PR. There the db-service@1.7.0 is already available.

johannes-vogel:
> this should work in the release PR. There the db-service@1.7.0 is already available.

👍  thought so as well.

patricebender (approved):

BobdenOs (approved):

patricebender:
Next time we should use semantic commit type `deps` here, but as this was merged to the release branch, I had to manually add this to the changelog regardless.

johannes-vogel:
> Next time we should use semantic commit type `deps` here, but as this was merged to the release branch, I had to manually add this to the changelog regardless.

the problem is that you can only force merge it to main, as the versions do not yet exist.

Issue #543, opened 2024-03-21
@cap-js/postgres: column reference is ambiguous


Floghost:
### Description of erroneous behaviour 

Hello!
In special cases the error "column reference is ambiguous" is occured.

Here is the error from debugging-console:
```sh
[cds] - error: column reference "number" is ambiguous
    at c:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\pg\lib\client.js:526:17
    at process.processTicksAndRejections (C:\Users\flori\Documents\nodejs\Test\cap-pg\lib\internal\process\task_queues.js:95:5)
    at async Object.all (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@cap-js\postgres\lib\PostgresService.js:163:26)
    at async PostgresService.onSELECT (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@cap-js\db-service\lib\SQLService.js:127:16)
    at async next (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@sap\cds\lib\srv\srv-dispatch.js:79:17)
    at async PostgresService.handle (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@sap\cds\lib\srv\srv-dispatch.js:77:10)
    at async ApplicationService.<anonymous> (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@sap\cds\libx\_runtime\common\generic\crud.js:73:16)
    at async next (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@sap\cds\lib\srv\srv-dispatch.js:79:17)
    at async ApplicationService.handle (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@sap\cds\lib\srv\srv-dispatch.js:77:10)
    at async _readCollection (C:\Users\flori\Documents\nodejs\Test\cap-pg\node_modules\@sap\cds\libx\_runtime\cds-services\adapter\odata-v4\handlers\read.js:250:19) {length: 112, name: 'error', severity: 'ERROR', code: '42702', detail: undefined, …}
```

The problem seems to be the association of two entities with the same attribute name in combination with an SELECT-Statement with expand and orderby.

### Detailed steps to reproduce

Database entities:
```cds
using {
    cuid,
    managed
} from '@sap/cds/common';

namespace cap.pg.example;


entity Entity1 : cuid, managed {
    number  : String;
[...]

PR #546, opened 2024-03-25
fix(`order by`): for localized sorting, prepend table alias

patricebender:
Usually, column references in _simple_ names in order by clauses are left as is. That means **no table alias** is prepended. In expressions however, **only source elements** can be reffered to.

That is why in case of localized sorting, we must prepend the table alias nevertheless, as in the DB specific drivers, a `COLLATE` expression is wrapped around the string reference. Hence we must replace the column reference with the underlying source reference.

as reported in #543, postgres is not able to resolve a reference to a column unambigously if the reference is part of a `COLLATE` expression.


patricebender:
@stewsk @BobdenOs please have a look

test/scenarios/bookshop/orderBy.test.js
```diff
@@ -0,0 +1,46 @@
...
+    // the original query is sorted by the **column** "createdBy"
+    // the resulting query has two query sources in the end --> Authors and Books
+    // even though both Tables have the element "createdBy", the DB should be able to resolve the
+    // order by reference to the column unambiguously
+    const query = CQL(`SELECT from sap.capire.bookshop.Books {
+      createdBy,
+      author.name as author
+    } order by createdBy`)
+    const res = await cds.run(query)
+    expect(res.length).to.be.eq(5)
```
BobdenOs, line 17: I would recommend putting a limit on the query and expecting `1`. As it is order relevant it might make sense to check for the returned value. As currently it just checks that it does not throw.

test/scenarios/bookshop/orderBy.test.js
```diff
@@ -0,0 +1,46 @@
...
+  })
+  test('collations with val', async () => {
+    if(cds.env.requires.db.impl === '@cap-js/hana')
+      return // FIXME: the `val` is put into window function, which ends in an error on HANA
```
BobdenOs, line 21: Will have to look what can be done about that. Luckily sorting by static values is not the most common. Might even make sense to simply filter them out for `HANA`.

.gitignore
```diff
@@ -122,6 +122,9 @@ dist
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
+# VSCode configs
+/.vscode
```
BobdenOs, line 126: This seems a bit unrelated to this change. Are there some vscode configs we want to include in this repository ?

patricebender: it might make sense to enforce prettier as formatter

.gitignore
```diff
@@ -122,6 +122,7 @@ dist
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
+
```
patricebender, line 125:
```suggestion
```

BobdenOs (approved):

PR #551, opened 2024-03-26
fix(`inline`, `expand`): Only accept on structures, assocs or table aliases

patricebender:
an `expand` or `inline` on a `many` type has no defined behavior, yet.
→ as of now this is an error because the path cant be resolved (expand / inline without `.elements`)

we should reject it with a proper error, just as the compiler does for static views.

### DIFF
--- a/db-service/test/bookshop/db/schema.cds
+++ b/db-service/test/bookshop/db/schema.cds
@@ -300,6 +300,9 @@ entity SoccerPlayers {
   key jerseyNumber: Integer;
   name: String;
   team: Association to SoccerTeams;
+  emails: many {
+    address: String;
+  }
 }
 
 entity TestPublisher {
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -727,6 +727,9 @@ function infer(originalQuery, model) {
       function resolveInline(col, namePrefix = col.as || col.flatName) {
         const { inline, $refLinks } = col
         const $leafLink = $refLinks[$refLinks.length - 1]
+        if(!$leafLink.definition.target && !$leafLink.definition.elements) {
+          throw new Error(`Unexpected “inline” on “${col.ref.map(idOnly)}”; can only be used after a reference to a structure, association or table alias`)
+        }
         let elements = {}
         inline.forEach(inlineCol => {
           inferQueryElement(inlineCol, false, $leafLink, { inExpr: true, inNestedProjection: true, baseColumn: col })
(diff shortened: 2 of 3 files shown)


db-service/test/bookshop/db/schema.cds
```diff
@@ -300,6 +300,9 @@ entity SoccerPlayers {
   key jerseyNumber: Integer;
   name: String;
   team: Association to SoccerTeams;
+  emails: many {
```
johannes-vogel, line 303: afaik, collection valued properties are not queried with expand but as a regular column

patricebender: Okay, this PR aligns the runtime with the compiler behavior. I would say, this change is fine as-is because it definetly improves the current error situation. If we come up with some meaningful semantics I'd be happy to implement it 👍

patricebender: How about `Expecting a reference to a structure, association, or table alias before using '{ ‹expand› }'.`

---

Navigating into an arrayed structure is not possible, yet. At least its not supported by the compiler SQL Backend. 

```cds
entity Authors {
  key ID         : Int16;
      name       : String(111);
      manyStruct : many {
        leaf   : Integer;
        sub    : {
          leaf : Integer;
        }
      };
}
```

will in the end only be:

```sql
-- generated by cds-compiler version 4.8.1
CREATE TABLE Authors (
  ID SMALLINT NOT NULL,
  name NVARCHAR(111),
  manyStruct NCLOB, -- leafs are all gone
  PRIMARY KEY(ID)
);
```

if you try to navigate into the struct:
```log
Error[ref-unexpected-many-navigation]: Unexpected navigation into arrayed structure
    |
  e.cds:23:5, at entity:“A”/select/column:“leaf”
    |
 23 |     manyStruct.leaf
    |     ^
```

---


→ The error message which comes with this change tries to tell the user that the expand on the arrayed element is _unexpected_. In the second part of the message, I try to give guidance on which columns the `expand` could be used.

In the compiler the error message is very similar:

```log
Error[def-unexpected-nested-proj]: Unexpected ‘{ ‹expand› }’; can only be used after a reference to a structure, association or table alias
    |
  e.cds:23.16-25.6, at entity:“A”/column:“manyStruct”
    |
 23 |     manyStruct {
 24 |       leaf
 25 |     }
```

patricebender:
@stewsk please have a look

stewsk (approved):

PR #552, opened 2024-03-27
fix: Only check first row if no changes required

BobdenOs:
When post processing blobs in SQLite stop walking the data as early as possible. If the first row shows that there are not columns to be changed return false and skip all following rows.


db-service/lib/SQLService.js
```diff
@@ -65,28 +65,30 @@ class SQLService extends DatabaseService {
...
+    let changes = false
     for (let col of columns) {
       const name = col.as || col.ref?.[col.ref.length - 1] || (typeof col === 'string' && col)
       if (col.element?.isAssociation) {
         if (one) this._changeToStreams(col.SELECT.columns, rows[0][name], false, compat)
         else
-          rows.forEach(row => {
-            this._changeToStreams(col.SELECT.columns, row[name], false, compat)
-          })
+          changes = rows.find(row => !this._changeToStreams(col.SELECT.columns, row[name], false, compat))
```
danjoa, line 74: rows.some instead of rows.find?

johannes-vogel:
```suggestion
          changes = rows.some(row => !this._changeToStreams(col.SELECT.columns, row[name], false, compat))
```

johannes-vogel (approved):

PR #558, opened 2024-03-27
test(odata): added tests for all built-in query functions

etimr:


test/scenarios/bookshop/funcs.test.js
```diff
@@ -234,7 +260,8 @@ describe('Bookshop - Functions', () => {
     })
 
     test.skip('fractionalseconds', async () => {
-      // REVISIT: ERROR: Feature is not supported: Method "fractionalseconds" in $filter or $orderby query options
+      // okra error: Feature is not supported: Method "fractionalseconds" in $filter or $orderby query options
+      // new adapter: no error, wrong result
```
BobdenOs, line 264: The function does not seem to produce the wrong result, but it is a miss match of precision. The function returns with 7 precision, but the query is checking for 3 precision.

[fractionalseconds](https://github.com/cap-js/cds-dbs/blob/main/db-service/lib/cql-functions.js#L197C44-L197C50)

test/scenarios/bookshop/funcs.test.js
```diff
@@ -249,14 +276,16 @@ describe('Bookshop - Functions', () => {
       expect(res.data.value.length).to.be.eq(1)
     })
     test.skip('maxdatetime', async () => {
-      // REVISIT: ERROR: Feature is not supported: Method "maxdatetime" in $filter or $orderby query options
+      // okra error: Feature is not supported: Method "maxdatetime" in $filter or $orderby query options
+      // new adapter error: 500 - sql error
```
BobdenOs, line 280: The `maxdatetime` and `mindatetime` are missing the SQL string quotes.

[maxdatetime](https://github.com/cap-js/cds-dbs/blob/main/db-service/lib/cql-functions.js#L203)

test/scenarios/bookshop/funcs.test.js
```diff
@@ -249,14 +276,16 @@ describe('Bookshop - Functions', () => {
...
       const res = await GET(`/browse/Books?$select=ID&$filter=maxdatetime() eq 9999-12-31T23:59:59.999Z&$top=1`)
 
       expect(res.status).to.be.eq(200)
       expect(res.data.value.length).to.be.eq(1)
     })
     test.skip('mindatetime', async () => {
-      // REVISIT: ERROR: Feature is not supported: Method "mindatetime" in $filter or $orderby query options
+      // okra error: Feature is not supported: Method "mindatetime" in $filter or $orderby query options
+      // new adapter error: 500 - sql error
```
BobdenOs, line 288: see previous comment

test/scenarios/bookshop/funcs.test.js
```diff
@@ -276,6 +305,7 @@ describe('Bookshop - Functions', () => {
     })
     test('now', async () => {
       // REVISIT: this test does not really proof much
+      // new adapter error: 500 - sql error
```
BobdenOs, line 308: `now` is not implemented on the database service. Could be added as a proxy for `session_context('$now')`.

test/scenarios/bookshop/funcs.test.js
```diff
@@ -288,9 +318,17 @@ describe('Bookshop - Functions', () => {
...
+      // Works on SQLite
+      const res = await GET(`/browse/Books?$select=ID,createdAt&$filter=time(createdAt) gt 00:01:45&$top=1`)
+
+      expect(res.status).to.be.eq(200)
+      expect(res.data.value.length).to.be.eq(1)
+    })
     test.skip('totaloffsetminutes', async () => {
-      // REVISIT: ERROR: Feature is not supported: Method "totaloffsetminutes" in $filter or $orderby query options
+      // okra error: Feature is not supported: Method "totaloffsetminutes" in $filter or $orderby query options
+      // new adapter: no error, wrong result
```
BobdenOs, line 331: Should be double checked what the actual result is of the function for this value.

[totaloffsetminutes](https://github.com/cap-js/cds-dbs/blob/main/db-service/lib/cql-functions.js#L220)

BobdenOs (commented):
Would be good if we could check the failing and erroneous functions.

BobdenOs (approved):

postgres/lib/func.js
```diff
@@ -18,9 +18,14 @@ const StandardFunctions = {
...
   day: x => `date_part('day', ${castVal(x)})`,
+  time: x => `to_char(${castVal(x)}, 'HH24:MI:SS')`,
   hour: x => `date_part('hour', ${castVal(x)})`,
   minute: x => `date_part('minute', ${castVal(x)})`,
-  second: x => `date_part('second', ${castVal(x)})`,
+  second: x => `floor(date_part('second', ${castVal(x)}))`,
+  fractionalseconds: x => `CAST(date_part('second', ${castVal(x)}) - floor(date_part('second', ${castVal(x)})) AS DECIMAL)`,
+  now: function() {
+    return this.session_context({val: '$now'})
+  }
```
patricebender, line 28: As new functions are added, I would see this change as a _feature_, but the commit titel says "fix", am I wrong with my assumption?

etimr: It's a little bit of both, some functions like `second` and `maxdatetime` on sqlite are fixed, some are added, especially in postgres and hana. I can change the pr to feature if you think it's more appropriate.

PR #559, opened 2024-03-28
Reduce localized views

danjoa:
Adds support for new option `cds.env.sql.transitive_localized_views` to skip generating localized views for entities which don't have own localized elements, but only associations to such.
  - `undefined`: such views are not generated for new db services
  - `false`: they will never be generated
  - `true` they will always be generated

### DIFF
--- a/db-service/test/cqn4sql/where-exists.test.js
+++ b/db-service/test/cqn4sql/where-exists.test.js
@@ -2,6 +2,7 @@
 const cqn4sql = require('../../lib/cqn4sql')
 const cds = require('@sap/cds/lib')
 const { expect } = cds.test
+const transitive_ = !cds.unfold || 'transitive_localized_views' in cds.env.sql && cds.env.sql.transitive_localized_views !== false
 
 
 /**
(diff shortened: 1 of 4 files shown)


db-service/test/cqn4sql/where-exists.test.js
```diff
@@ -1321,28 +1322,30 @@ describe('cap issue', () => {
...
+      where exists foo.specialOwners[owner2_userID = $user.id]
+      or exists foo.activeOwners[owner_userID = $user.id]
+    `
+    let q2 = cqn4sql(q, model)
+    expect(cds.clone(q2)).to.deep.equal(CQL(`
+      SELECT from localized.Boo as boos { boos.ID } WHERE EXISTS (
+        SELECT 1 from localized.Foo as Foo3 WHERE Foo3.ID = boos.foo_ID
+      ) AND (
+        EXISTS (
+          SELECT 1 from localized.Foo as foo WHERE foo.ID = boos.foo_ID AND EXISTS (
```
patricebender, line 1335: dont we need the ternary for this `localized` as well?
```suggestion
          SELECT 1 from ${transitive_?'localized.':''}Foo as foo WHERE foo.ID = boos.foo_ID AND EXISTS (
```

db-service/test/cqn4sql/where-exists.test.js
```diff
@@ -1321,28 +1322,30 @@ describe('cap issue', () => {
...
+            SELECT 1 from ${transitive_?'localized.':''}SpecialOwner2 as specialOwners
+            WHERE specialOwners.foo_ID = foo.ID and specialOwners.owner2_userID = $user.id
           )
-          or EXISTS (
-            SELECT 1 from localized.Foo as foo2 where foo2.ID = boos.foo_ID
-              and EXISTS ( SELECT 1 from ${localized_}ActiveOwner as activeOwners where activeOwners.foo_ID = foo2.ID and activeOwners.owner_userID = $user.id )
+        )
+        OR
+        EXISTS (
+          SELECT 1 from localized.Foo as foo2 WHERE foo2.ID = boos.foo_ID AND EXISTS (
```
patricebender, line 1342: dont we need the ternary for this `localized` as well?
```suggestion
          SELECT 1 from ${transitive_?'localized.':''}Foo as foo2 WHERE foo2.ID = boos.foo_ID AND EXISTS (
```

patricebender (approved):

db-service/test/cqn4sql/where-exists.test.js
danjoa: No, we only need that for transitive localized views, i.e. on top of entities that don't have own localized elements themselves, but only references (associations) to such. `Foo` obviously is not of that category but has own localized elements.

db-service/test/cqn4sql/where-exists.test.js
danjoa: as above

PR #561, opened 2024-04-02
fix(hdb): Allow streaming blob in and out over the same connection

BobdenOs:
Currently it is possible to have draft operations hang when using the `hdb` driver. As the draft logic will stream `blob` values in/out of the database over the same connection. When using the `hdb` driver the `insert` statement is claiming the connection and all `readLob` requests are queued until the `insert` statement finishes. It will never finish as the contents of the blob have to be send as part of the `insert` statement.

This PR avoids this issue by opening a blob streaming specific connection. That is used when the query could potentially return blobs that will be read after the result has been returned. Allowing the `insert` statement to fully stream the blob contents.

This issue doesn't happen when using `hana-client` as the `insert` statement directly returns before sending the data. Allowing `hana-client` to send `readLob` and `writeLob` requests over the same connection.

### DIFF
--- a/hana/lib/drivers/hdb.js
+++ b/hana/lib/drivers/hdb.js
@@ -19,7 +19,7 @@ class HDBDriver extends driver {
    * Instantiates the HDBDriver class
    * @param {import('./base').Credentials} creds The credentials for the HDBDriver instance
    */
-  constructor(creds) {
+  constructor(creds, forBlobs) {
     creds = {
       useCesu8: false,
       fetchSize: 1 << 16, // V8 default memory page size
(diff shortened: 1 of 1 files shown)


soccermax:
@BobdenOs can we also try to fix this in HDB? Because with this solution we would have for lobs always two database connections. What do you think?

BobdenOs:
@soccermax the way `hdb` has setup its connection queue makes it a fundamental part of the library. It might be possible to have the insert return early just like `hana-client`, but I was not able to pinpoint where to do this. So rather then putting to much time into investigating that option. I wanted to have a possible solution first.

patricebender (approved):
LGTM, is this still relevant?

do we need this first?:

> This is a workaround and for it to work all the time has to be addressed inside the [node-hdb](https://github.com/sap/node-hdb) repository.

PR #562, opened 2024-04-02
fix: etag with stream_compat

vkozyura:


db-service/lib/SQLService.js
```diff
@@ -128,7 +128,14 @@ class SQLService extends DatabaseService {
     if (cds.env.features.stream_compat) {
       if (query._streaming) {
         this._changeToStreams(cqn.SELECT.columns, rows, true, true)
-        return rows.length ? { value: Object.values(rows[0])[0] } : undefined
+        if (!rows.length) return
```
BobdenOs, line 131: Why check it after calling `_changeToStreams` ?

db-service/lib/SQLService.js
```diff
@@ -128,7 +128,14 @@ class SQLService extends DatabaseService {
...
+        
+        const result = rows[0]
+        let [key, val] = Object.entries(result)[0]
```
BobdenOs, line 134: Using `[0]` is not allowed if the object has potentially multiple properties. Would be required to get the key from the `cqn.elements` to ensure that the correct value is deleted and set as `value`.

BobdenOs (approved):

PR #567, opened 2024-04-09
fix: Improved placeholders and limit clause

BobdenOs:
Added more cases that the `SQL` contains place holders. It now includes `null` and `numbers`. Now that it also includes `numbers` it is also used for the `limit` clause. With special behavior for `HANA` where the `LIMIT` is inline and the `OFFSET` is a place holder. Allowing the execution plan to create better optimizations.


db-service/test/cqn2sql/__snapshots__/delete.test.js.snap
```diff
@@ -1,8 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`delete test complex cascade delete for entity with 'not exists' 1`] = `"DELETE FROM Books as Books WHERE ( Books.author_id is not null or Books.author_version is not null ) and not exists (SELECT 1 as _exists FROM Author as Author left JOIN Books as parent ON parent.ID = Author.parent_ID WHERE Author.parent_ID = parent.code and parent.descr = Author.version)"`;
+exports[`delete test complex cascade delete for entity with 'not exists' 1`] = `"DELETE FROM Books as Books WHERE ( Books.author_id is not null or Books.author_version is not null ) and not exists (SELECT ? as _exists FROM Author as Author left JOIN Books as parent ON parent.ID = Author.parent_ID WHERE Author.parent_ID = parent.code and parent.descr = Author.version)"`;
```
johannes-vogel, line 3: Is there a way to detect SELECT 1? That's a common pattern in our implementation that we could also try to put in SQL string.

db-service/lib/cqn2sql.js
```diff
@@ -387,7 +387,7 @@ class CQN2SQLRenderer {
    */
   limit({ rows, offset }) {
     if (!rows) throw new Error('Rows parameter is missing in SELECT.limit(rows, offset)')
-    return !offset ? rows.val : `${rows.val} OFFSET ${offset.val}`
+    return !offset ? this.val(rows) : `${this.val(rows)} OFFSET ${this.val(offset)}`
```
johannes-vogel, line 390: Does this use placeholders only for LIMIT? Doesn't seem to me this way.

BobdenOs:
Blocked: DINC0119453

db-service/lib/cqn2sql.js
BobdenOs: The limit is only inlined when using HANA. Here: https://github.com/cap-js/cds-dbs/pull/567/files#diff-11bd20d3012fd985095f5a9be4a1da3e94e09eac409545ac67688c9582d1d201R711

db-service/test/cqn2sql/__snapshots__/delete.test.js.snap
BobdenOs: This is also a requirement for it to work with the `hdb` driver on HANA. As it checks the datatype of the place holder and of the data send to the placeholder. When doing `SELECT ?` the type of `?` resolves to `VARCHAR` and then sending `1` the driver rejects the request. So now when a column is a `{val:...}` it will always be inlined as otherwise all type information is lost.

johannes-vogel (approved):

db-service/lib/cqn2sql.js
```diff
@@ -290,7 +290,7 @@ class CQN2SQLRenderer {
       return `extensions__->${this.string('$."' + x.element.name + '"')} as ${x.as || x.element.name}`
     }
     ///////////////////////////////////////////////////////////////////////////////////////
-    let sql = this.expr(x)
+    let sql = this.expr({ param: false, __proto__: x })
```
danjoa, line 293: Why do we need that?

db-service/lib/cqn2sql.js
```diff
@@ -884,27 +884,35 @@ class CQN2SQLRenderer {
   }
 
   /**
-   * Renders a value into the correct SQL syntax of a placeholder for a prepared statement
+   * Renders a value into the correct SQL syntax or a placeholder for a prepared statement
    * @param {import('./infer/cqn').val} param0
    * @returns {string} SQL
    */
   val({ val, param }) {
+    const inline = !this.values || param === false
```
danjoa, line 892: Why did you pull that up here, if we still only need that down there in line 907, only if we reach that, of course?

db-service/lib/cqn2sql.js
```diff
@@ -884,27 +884,35 @@ class CQN2SQLRenderer {
...
+          else val = JSON.stringify(val)
+        }
+    }
+    if (inline) {
+      switch (typeof val) {
+        case 'string': return this.string(val)
+        case 'object': return 'NULL'
+        default:
+          return `${val}`
+      }
```
danjoa, line 913: Why did the old code work before and not anymore?

db-service/lib/cqn2sql.js
```diff
@@ -884,27 +884,35 @@ class CQN2SQLRenderer {
...
    */
   val({ val, param }) {
+    const inline = !this.values || param === false
     switch (typeof val) {
       case 'function': throw new Error('Function values not supported.')
-      case 'undefined': return 'NULL'
+      case 'undefined': val = null
+        break
       case 'boolean': return `${val}`
-      case 'number': return `${val}` // REVISIT for HANA
```
danjoa, line 896: Why don't we pass numbers as bind parameters anymore? Has that advantages?

db-service/lib/cqn2sql.js
BobdenOs: Before the only value that would arrive here where strings. Now it can also be numbers and `null`. It could also have Booleans, but currently they are still being in lined.

db-service/lib/cqn2sql.js
BobdenOs: When selecting static values it is required to always inline them. Otherwise typed databases are confused what type it is supposed to return and causes trouble with `SELECT 1` as `SELECT ?` will only accept strings as parameters as the default type of `?` is `nvarchar`. Breaking a lot of queries.

PR #570, opened 2024-04-09
fix: ignore streaming properties

SamuelBrucksch:


sjvans (changes requested):
`element.type === 'cds.LargeBinary'` should be the only check

PR #575, opened 2024-04-10
chore: rm static new adapter flag

sjvans:
accidentally introduced in https://github.com/cap-js/cds-dbs/pull/558


johannes-vogel:
solved with #574

PR #576, opened 2024-04-10
test(cqn4sql): with very deeply nested JSON, cqn4sql might not terminate

patricebender:
reason is to be investigated, see test 'PUT delete deep composition must not affect other same level composition in response' in runtime: `cds/tests/_runtime/odata/__tests__/integration/complexAssociationsAndCompositions.test.js`


BobdenOs:
The issue here seems more related to `deep-queries.js`. Where `getExpandForDeep` tries to get all data deeply nested. While in the test the data affected it is clear that it could be represented by a few simple deep deletes.

If the concern is the issue of `cqn4sql` hanging. It makes sense to continue, but if the issue is just making the test scenario work. It might be best to wait for the deep queries rework.

PR #583, opened 2024-04-11
feat: return int64 and decimals as string to not lose precision

SamuelBrucksch:
All DBs should handle int64/decimal the same.


patricebender (commented):
please add a test which runs for all dbs

BobdenOs:
Please adjust and uncomment the original tests here: https://github.com/cap-js/cds-dbs/blob/fdad6557f1aada361d48c751c5c5a290c787b520/test/compliance/resources/db/basic/literals/basic.literals.number.js#L16-L34

It will be required to add this to HANA and Postgres as well.

BobdenOs (approved):
Will impact stakeholder expectations. Lets make sure we follow up on that.

PR #586, opened 2024-04-12
chore: fix collation issue for aggregations and subselects

johannes-vogel:
this test should reproduce the issue.

I've verified that df19aa2 would fix it but I am not sure whether it's the correct fix.

### DIFF
--- a/test/scenarios/bookshop/orderBy.test.js
+++ b/test/scenarios/bookshop/orderBy.test.js
@@ -36,7 +36,7 @@ describe('Bookshop - Order By', () => {
   test('collations with val', async () => {
     if(cds.env.requires.db.impl === '@cap-js/hana')
       return // FIXME: the `val` is put into window function, which ends in an error on HANA
-    const query = CQL(`SELECT from sap.capire.bookshop.Books { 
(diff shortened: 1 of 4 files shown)


test/scenarios/bookshop/orderBy.test.js
```diff
@@ -18,10 +18,25 @@ describe('Bookshop - Order By', () => {
...
+
+    const res = await cds.run(query)
+    expect(res.length).to.be.eq(1)
+    expect(res[0]).to.eql({ title: 'Catweazle', pri: 150})
+  })
   test('collations with val', async () => {
     if(cds.env.requires.db.impl === '@cap-js/hana')
       return // FIXME: the `val` is put into window function, which ends in an error on HANA
-    const query = CQL(`SELECT from sap.capire.bookshop.Books {
+    const query = CQL(`SELECT from sap.capire.bookshop.Books {
```
patricebender, line 39:
```suggestion
    const query = CQL(`SELECT from sap.capire.bookshop.Books {
```

patricebender (approved):

PR #589, opened 2024-04-16
fix: Disconnect HANA tenant when deleted

BobdenOs:
When using `cds-mtxs` in a scale out scenario can result in an invalid connection pool cache. This change clears the tenant connection pool when a known entity is no longer deployed. Allowing the following request to create a new pool or fail when fetching the new credentials.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -134,9 +134,19 @@ class HANAService extends SQLService {
 
     // REVISIT: add prepare options when param:true is used
     const sqlScript = isLockQuery ? sql : this.wrapTemporary(temporary, withclause, blobs)
-    let rows = (values?.length || blobs.length > 0)
+    let rows
+    try {
+      rows = (values?.length || blobs.length > 0)
       ? await (await this.prepare(sqlScript, blobs.length)).all(values || [])
       : await this.exec(sqlScript)
+    } catch (err) {
+      // Ensure that the known entity still exists
+      if (cds.context.tenant && (err.code === 321 || err.code === 259)) {
+        // Clear current tenant connection pool
+        cds.disconnect(cds.context.tenant)
+      }
+      throw err
+    }
 
     if (isLockQuery) {
       // Fetch actual locked results


hana/lib/HANAService.js
```diff
@@ -30,6 +30,7 @@ class HANAService extends SQLService {
     this.on(['BEGIN'], this.onBEGIN)
     this.on(['COMMIT'], this.onCOMMIT)
     this.on(['ROLLBACK'], this.onROLLBACK)
+    this.on(['SELECT', 'INSERT', 'UPSERT', 'UPDATE', 'DELETE'], this.onNOTFOUND)
```
johannes-vogel, line 33:
```suggestion
    this.on('error', this.onNOTFOUND)
```

should we use the error event instead?

hana/lib/HANAService.js
```diff
@@ -181,6 +182,20 @@ class HANAService extends SQLService {
     }
   }
 
+  async onNOTFOUND(req, next) {
+    try {
+      return await next()
+    } catch (err) {
+      // TODO: check that the query target is defined in the model
+      // Ensure that the known entity still exists
+      if (/*this.context.tenant && */ err.code === 259) {
```
johannes-vogel, line 191: Eventually this ends in a "hick-up" and the current request is terminated with 500. That's the best we can do, isn't it?

patricebender (commented):
should we add a test for this on our side?

BobdenOs:
> should we add a test for this on our side?

If I could add a test for this I would. As the actual test for this issue can only be run with privileges I don't have.

hana/lib/HANAService.js
BobdenOs: At this point we have to assume that the database is toast. So it will be required to fail and enable the possibility to recover at a later point in time.

swaldmann (approved):
OK from MTX side as this solves Oliver's issue, i.e. our MTX HANA tests are green with both this change and the one in `@sap/cds-mtxs`.

johannes-vogel:
> OK from MTX side as this solves Oliver's issue, i.e. our MTX HANA tests are green with both this change and the one in `@sap/cds-mtxs`.

It only is solved for Oliver because they have retries implemented, isn't it?

hana/lib/HANAService.js
BobdenOs: > Also the handler should only be registered if multitenancy is enabled.

When the Database service is created it is not always configured as Multitenancy enabled. My approach would be to check whether the current context has a tenant defined or not.

swaldmann:
> It only is solved for Oliver because they have retries implemented, isn't it?

Yes, and also in our integration test we're first sending a request to instance 1 when it's offboarded, where it correctly fails. Later we send the request again and with this PR it doesn't fail any more. I assume if we don't send the first request, the first request after onboarding would also fail.

hana/lib/HANAService.js
johannes-vogel: I've created a proposal for the error handler with this draft PR #597 
What I prefer with that implementation that it only kicks in if an error occurs. With the interceptor, there's a slight overhead for all successful requests as well.

hana/lib/HANAService.js
BobdenOs, line 191:
```suggestion
      // Ensure that the known entity still exists
      if (!this.context.tenant && err.code === 259 && typeof req.query !== 'string') {
```

johannes-vogel (approved):

PR #590, opened 2024-04-16
feat(search): enable deep search with path expressions

patricebender:
support search via path expressions:

- deep search via associations
- include calculated elements if explicitly requested

---

with `cds.infer` being a part of all our query objects, each query comes with `target` and `elements` even without the `infer` function of the `db-service`being called. Hence, it was possible to calculate the search term as a very first step of `cqn4sql`. After the `contains(…)` function has been added to the queries `where` or `having` clause the usual join tree algorithm is executed, making it possible to use path expressions within the search term.

---

rework of https://github.com/cap-js/cds-dbs/pull/252


db-service/test/cqn4sql/search.test.js
```diff
@@ -222,3 +222,147 @@ describe('Replace attribute search by search predicate', () => {
...
+      {
+        func: 'search',
+        args: [{ list: [{ ref: ['author', 'lastName'] }, { ref: ['author', 'firstName'] }] }, { val: 'x' }],
+      },
+    ]
+    expected.SELECT.where = where
+    expect(JSON.parse(JSON.stringify(res))).to.deep.equal(expected)
+
+    // add another searchable field
+    BooksSearchAuthor['@cds.search.title'] = true
```
patricebender, line 361: this would not have been possible with existing caching logic

db-service/lib/infer/index.js
```diff
@@ -380,6 +381,16 @@ function infer(originalQuery, model) {
     if (_.with)
       // consider UPDATE.with
       Object.values(_.with).forEach(val => inferQueryElement(val, false))
+    if (search) {
```
johannes-vogel, line 384: Why do we need to infer the search elements? can't this be part of cqn4sql? I naively expected it there.

johannes-vogel: Can't we do it roughly like this:

1. cqn4sql is called
2. as a very first step SELECT.search is rewritten to search function and appended to where clause
3. infer does its magic
4. cqn4sql builds join tree as usual

If we have an abstract function search that does everything search related, I'd like to keep the rest independent of search. hope it's not too weird and somehow understandable what I'd like to express :) if not, we should have a call :)

patricebender: I will try it.

There a couple of obstacles to overcome if we try to build a search term without inferring the query first.

1. there is no query target, can be solved by relying on Daniels infer by setting the prototype to `SELECT()` if a query has not been initialized with the CQL api.
2. columns are not yet resolved to query elements, hence there is no type information available at the column itself -> this becomes relevant for queries with "aggregations"

patricebender: okay, I made it work with `cds.infer`. However, I found a bug in Daniels implementation. We need `cap/cds/pull/4160` (hopefully with `cds@7.9.0`) before we can merge this.

patricebender: unfortunately, after upgrading to cds 7.9, [the HANA job is failing.](https://github.com/cap-js/cds-dbs/actions/runs/9063899197/job/24900854487)

@johannes-vogel @BobdenOs Do these errors look familiar to you? If not, I will have a closer look.

db-service/lib/cqn4sql.js
```diff
@@ -44,8 +44,25 @@ const { pseudos } = require('./infer/pseudos')
...
+  const hasCustomJoins = originalQuery.SELECT?.from.args && (!originalQuery.joinTree || originalQuery.joinTree.isInitial)
+
+  if (!hasCustomJoins && inferred.SELECT?.search) {
+    // we need an instance of query because the elements of the query are needed for the calculation of the search columns
+    if (!inferred.SELECT.elements) Object.setPrototypeOf(inferred, Object.getPrototypeOf(SELECT()))
+    const searchTerm = getSearchTerm(inferred.SELECT.search, inferred)
+    if (searchTerm) {
+      // Search target can be a navigation, in that case use _target to get the correct entity
+      const { where, having } = transformSearch(searchTerm)
+      if (where) inferred.SELECT.where = where
```
johannes-vogel, line 57: don't you need to append the search terms instead of replacing the existing where/having clauses?

patricebender: This is done in line 243ff

patricebender:
blocked by `cap/issues/issues/16172`

johannes-vogel:
> blocked by `cap/issues/issues/16172` → `cds.infer` can't handle table aliases:
> 
> ```js
> > q = SELECT`from ${Books} as Foo { Foo.ID }`
> Query {
>   SELECT: {
>     from: { ref: [ 'my.bookshop.Books' ], as: 'Foo' },
>     columns: [ { ref: [ 'Foo', 'ID' ] } ]
>   }
> }
> > q.elements
> Uncaught:
> Error: Couldn't resolve element "Foo/ID" in entity my.bookshop.Books [ 'ID', 'title', 'stock' ]
>     at /Users/patricebender/SAPDevelop/foo/node_modules/@sap/cds/lib/ql/infer.js:80:22
>     at Array.forEach (<anonymous>)
>     at Function.elements4 (/Users/patricebender/SAPDevelop/foo/node_modules/@sap/cds/lib/ql/infer.js:61:32)
>     at get elements [as elements] (/Users/patricebender/SAPDevelop/foo/node_modules/@sap/cds/lib/ql/SELECT.js:157:53)
> ```

Why would cds.infer have to handle aliases? Where do they stem from?

patricebender:
> Why would cds.infer have to handle aliases?

My sample illustrates the problem, does it not? A queries column must be resolved to an element which is then attached to the query, this must work with and without table aliases in front of 
a given ref.

> Where do they stem from?
Each `cqn4sql` transformation assigns unique table aliases to the given query. All transformed columns will be absolutified → all refs are prepended with the unique table alias.
This also accounts for where exists subqueries, or correlated subqueries stemming from `expand`

### Why now?

If we want to support path expressions in a search annotation, we must attach a `search` function to the query **before it is put into the join tree algorithm** for processing. But for the calculation of the searchable columns, we need to have an _inferred query at hand_. We need the `target` of a query, to find out which elements must be searched. Hence, I rely on `cds.infer` to do the work instead of doing it by myself. If `cqn4sql` sees an `expand` column, it build a subquery, attaches a table alias and transforms the query. Here the bug I just illustrated kicks in, because `cds.infer` can't resolve table aliases whatsoever.

db-service/lib/cqn4sql.js
```diff
@@ -206,46 +216,33 @@ function cqn4sql(originalQuery, model) {
...
-    if (searchIn.length > 0) {
-      const xpr = search
-      const contains = {
-        func: 'search',
-        args: [
-          searchIn.length > 1 ? { list: searchIn } : { ...searchIn[0] },
-          xpr.length === 1 && 'val' in xpr[0] ? xpr[0] : { xpr },
-        ],
-      }
+    const contains = searchTerm
```
johannes-vogel, line 230: renaming and comment are obsolete?

johannes-vogel (approved):

PR #598, opened 2024-04-18
fix: flat update with arbitrary where clauses

larslutz96:
Flat update with or on keys in where resulted in error `UNIQUE_CONSTRAINT_VIOLATION`.
Solution: Only enrich data with keys from where for deep update.

### DIFF
--- a/db-service/lib/fill-in-keys.js
+++ b/db-service/lib/fill-in-keys.js
@@ -1,4 +1,5 @@
 const cds = require('@sap/cds')
+const { hasDeep } = require('../lib/deep-queries')
 
 // REVISIT: very deep & fragile dependencies to internal modules -> copy these into here
 const propagateForeignKeys = require('@sap/cds/libx/_runtime/common/utils/propagateForeignKeys')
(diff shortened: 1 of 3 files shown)


patricebender (commented):
please use a descriptive PR title as the title will end up in the changelog

db-service/lib/fill-in-keys.js
```diff
@@ -57,9 +57,12 @@ const generateUUIDandPropagateKeys = (entity, data, event) => {
 module.exports = async function fill_in_keys(req, next) {
   // REVISIT dummy handler until we have input processing
   if (!req.target || !this.model || req.target._unresolved) return next()
-  if (req.event === 'UPDATE') {
+  // only for deep update
+  const containsObjectOrArray = obj =>
```
BobdenOs, line 61: This check already exists inside the `deep-queries.js`. Instead make sure this function is only called when it should be called. If it is not required for any queries where it does not deep then it should be made sure this function is not called.

As this check might identify a value which is actually ignored. Like an association.

BobdenOs (changes requested):

db-service/lib/fill-in-keys.js
larslutz96: would change this to: `req.event === 'UPDATE' && hasDeep(req.query, req.target)`

johannes-vogel (approved):

PR #599, opened 2024-04-18
fix(`order by`): reject non-fk traversals of own columns in order by

patricebender:
`refs` in `order by` can address queries own columns, if such a `ref` traverses an association to a non-foreign key field as in this example: 

```sql
SELECT from bookshop.Books as foo {
        ID,
        author,
        coAuthor as co
      }
      order by
        co.name -- join with query "foo" not possible
```

we must reject the traversal, as the join partner of `Authors` is the query itself

### other changes:

fix(`order by`): do not mistake simple name with infix filter
  another fix do not mistake simple name with infix filter: If
  the query has `columns` but they solely consist of `*`, 
  `cqn4sql` wrongly assumed that an association traversal has to be resolved in the queries elements, this lead to a 
  confusing error complaining that only foreign key of assocs can be traversed in infix filters.

### DIFF
--- a/db-service/test/cqn4sql/table-alias.test.js
+++ b/db-service/test/cqn4sql/table-alias.test.js
@@ -553,7 +553,8 @@ describe('table alias access', () => {
       { SimpleBook.ID, SimpleBook.title, SimpleBook.author_ID } order by author.name`
       expect(query).to.deep.equal(expected)
     })
-    it('same as above but author is explicit column', () => {
+    // doesnt work, can't join with the query source itself
+    it.skip('same as above but author is explicit column', () => {
       let input = CQL`SELECT from bookshop.SimpleBook { *, author } order by author.name`
       let query = cqn4sql(input, model)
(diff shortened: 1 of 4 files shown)


db-service/test/cqn4sql/table-alias.test.js
```diff
@@ -545,6 +545,23 @@ describe('table alias access', () => {
       let query = cqn4sql(input, model)
       expect(query).to.deep.equal(CQL`SELECT from bookshop.Books as Books { func() as func } order by func`)
     })
+
+    it('do not try to resolve ref in columns if columns consists of star', () => {
+      let input = CQL`SELECT from bookshop.SimpleBook { * } order by author.name`
+      let query = cqn4sql(input, model)
+      const expected = CQL`SELECT from bookshop.SimpleBook as SimpleBook left join bookshop.Authors as author on author.ID = SimpleBook.author_ID
+      { SimpleBook.ID, SimpleBook.title, SimpleBook.author_ID } order by author.name`
+      expect(query).to.deep.equal(expected)
```
patricebender, line 554: @SamuelBrucksch this is the scenario you described

stewsk (approved):

PR #601, opened 2024-04-19
fix: reject unmanaged associations in infix filter

BobdenOs:
This PR adds two tests for where exists clauses that contain path expressions. This style of cqn is generated by the OData adapter when using lambda functions.

```url
/Books?$filter=author/books/any(o:o/author/name eq 'Emily')
```

When the association is managed a special error is thrown, but when the association is unmanaged the query is passed along as is. This functionality used to only work on `HANA` as it natively processed the path expression, but for the new databases this should always work.


db-service/test/cqn4sql/where-exists.test.js
```diff
@@ -218,15 +218,15 @@ describe('EXISTS predicate in where', () => {
           CQL`SELECT from bookshop.Authors { ID } WHERE EXISTS books[dedication.addressee.name = 'Hasso']`,
           model,
         ),
-      ).to.throw('Only foreign keys of "addressee" can be accessed in infix filter')
+      ).to.throw('Only foreign keys of “addressee” can be accessed in infix filter')
```
BobdenOs, line 221: `“` or `"` ?

patricebender (approved):

PR #604, opened 2024-04-19
fix(hana): Support associations with static values

BobdenOs:


hana/lib/HANAService.js
```diff
@@ -453,6 +453,7 @@ class HANAService extends SQLService {
                 let fkeys = x.element._foreignKeys
                 if (typeof fkeys === 'function') fkeys = fkeys.call(x.element)
                 fkeys.forEach(k => {
+                  if (!k?.parentElement?.name) return // not all associations have foreign key references
```
patricebender, line 456: would it make sense to add a test for this? At a first glimpse I'm not sure what this does

PR #610, opened 2024-04-23
test: Add missing types to compliance test suite

BobdenOs:
While investigating a bug report for `@cap-js/hana` it was not possible to deploy the data locally. As the insert type for `UInt8` was missing. This is a result of the `UInt8` type not being supported on the `Postgres` dialect of the compiler. This PR enables the `UInt8` data type and `cds.hana.*` data types inside our test suite. Allowing the compliance test suite to add the currently missing data types.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -165,12 +165,14 @@ class CQN2SQLRenderer {
   /** @type {Object<string,import('@sap/cds/apis/csn').Definition>} */
   static TypeMap = {
     // Utilizing cds.linked inheritance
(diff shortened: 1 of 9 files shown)


danjoa:
I'd propose to have an interactive review, as I'd like to understand many things better, which seem hard to understand to me. Some of which:

Why do we have to translate native Hana types like sap.hana.TINYINT to something different in case of Hana? 

Why should we support these native Hana types in other databases as well? Aren't they meant to be non-portable? Would we have to do the same two-way / multi-way support if we'd introduce native Postgres types?

PR #613, opened 2024-04-23
chore: fix eslint warnings

patricebender:


eslint.config.js
```diff
@@ -49,14 +49,14 @@ module.exports = [
...
-      '@typescript-eslint/no-unused-vars': ['warn'],
+      '@typescript-eslint/no-unused-vars': ['warn', { argsIgnorePattern: '^_' }],
       '@typescript-eslint/no-explicit-any': ['warn'],
     },
   },
   {
     files: ['**/hana/**/*.js'],
     rules: {
-      'no-console': 'warn',
+      'no-console': 'off',
```
patricebender, line 59: @BobdenOs do we need the console in the hana service?

BobdenOs: I just double checked the warnings on another PR and we need the `console` statements in these files. They are used inside the setup scripts.

BobdenOs (approved):

PR #617, opened 2024-04-25
refactor(infer): consolidate two main functions

patricebender:
when the `infer` function was developed, it grew quite fast and some functions were tailor made for very specific use cases. While separation of concerns is a good practice w.r.t. maintainability, the wheel must not be re-invented for each minor deviation in behavior.

That being said, the two function `inferQueryElement` and `attachRefLinksToArg` are doing essentially the same. They both walk an `arg` which is - generally spoken - an expression. That might be a simple `ref`, `func`, `val` or a (nested) `xpr`.

- `inferQueryElement` was used for the `columns` of a query. Each column was resolved to a query element. All elements together formed the set of query elements. The other main part of the function is the attachment of the infamous `$refLinks` next to each `ref` array. Moreover, there was a flag `insertIntoQueryElements` which could be set to `false`, if the `arg` must _not_ be inserted into the queries elements. This is the case for nested `xpr`, the `where`, `having`, `with`, `groupBy` and `orderBy` clause of the query.
- `attachRefLinksToArg` does exactly what the name implies, it attaches `$refLinks` to an `arg` but does not resolve the `arg` to a query element. This function was used for the `from` clause of the query.

It is obvious that both functions have significant shared logic. This change removes the usage of the `attachRefLinksToArg` function.


patricebender:
The stakeholder tests (`cap/stakeholder-tests/actions/runs/6078819`) are all green with this change, so I am confident that this can be merged.

stewsk (approved):

sjvans:
@patricebender @stewsk this pr breaks cds ci, i'll revert for now

PR #619, opened 2024-04-26
fix: Align all quote functions with @sap/cds-compiler

BobdenOs:
Add `toLowerCase()` to Postgres quoting. Align regex with `@sap/cds-compiler`.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -967,9 +967,8 @@ class CQN2SQLRenderer {
    */
   quote(s) {
     if (typeof s !== 'string') return '"' + s + '"'
-    if (s.includes('"')) return '"' + s.replace(/"/g, '""') + '"'
     // Column names like "Order" clash with "ORDER" keyword so toUpperCase is required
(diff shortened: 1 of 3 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -967,9 +967,8 @@ class CQN2SQLRenderer {
    */
   quote(s) {
     if (typeof s !== 'string') return '"' + s + '"'
-    if (s.includes('"')) return '"' + s.replace(/"/g, '""') + '"'
     // Column names like "Order" clash with "ORDER" keyword so toUpperCase is required
-    if (s in this.class.ReservedWords || /^\d|[$' ?@./\\]/.test(s)) return '"' + s + '"'
+    if (s.toUpperCase() in this.class.ReservedWords || !/^[A-Za-z_][A-Za-z_$0-9]*$/.test(s)) return '"' + s + '"'
```
danjoa, line 971: Please add 'Order', 'order', and 'ORDER' to ReservedWords, and thereby eliminate the need to always do toUpperCase() -- as we had it in a former revision. And no, we don't have to correctly handle pathological variants like 'OrDeR' ;)

danjoa (approved):

PR #620, opened 2024-04-26
fix(hana): Allow custom fuzzy search cqn

BobdenOs:
With the old database service it was possible to define a custom fuzzy search cqn.

The test implements it like:

```js
{
  func: 'CONTAINS',
  args: [
    { list: [{ ref: ['ID'] }, { ref: ['NAME'] }] },
    { val: hanaSearchTerm },
    { func: 'FUZZY', args: [{ val: 0.8 }, { val: 'similarCalculationMode=searchCompare' }] }
  ]
}
```

But according the HANA documentation this is incorrect. As `CONTIANS` is not a function it is actually a `predicate`. Which would be defined like:

```js
{
  xpr: [
    'CONTAINS',
    {
      xpr: [
        { list: [{ ref: ['ID'] }, { ref: ['NAME'] }] },
        { val: hanaSearchTerm },
        'FUZZY',
        {xpr: [[{ val: 0.8 }, { val: 'similarCalculationMode=searchCompare' }]]}
      ]
    }
  ]
}
```

Additionally the `contains` function from the OData spec clashes with the HANA internal function name. Which means that the custom code has to adjust the function name to `CONTAINS` to avoid the function from being rewritten into the OData spec compliant variant.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -866,7 +866,7 @@ class HANAService extends SQLService {
           if (up in caseOperators) break
           continue
         }
-        if (cur.func === 'CONTAINS') return true
+        if (cur.func?.toUpperCase() === 'CONTAINS' && cur.args?.length > 2) return true
         if ('_internal' in cur) return true
         if ('xpr' in cur) return this.is_comparator(cur)
       }
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -855,6 +855,7 @@ class HANAService extends SQLService {
           if (up in caseOperators) break
           continue
         }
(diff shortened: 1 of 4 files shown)


hana/lib/HANAService.js
```diff
@@ -855,6 +855,7 @@ class HANAService extends SQLService {
           if (up in caseOperators) break
           continue
         }
+        if ('func' in cur && cur.func === 'CONTAINS') return true
```
danjoa, line 858:
```suggestion
        if (cur.func === 'CONTAINS') return true
```

danjoa (approved):

PR #624, opened 2024-04-29
fix: Add multi `concat` function to `@cap-js/hana`

BobdenOs:
According to the `concat` definition it is allowed to provide more then two arguments to the function. The native HANA `concat` function only allows two arguments. Therefor this overwrite is taken from the base implementation to allow for more arguments.


patricebender (commented):
could we have a test for this?

johannes-vogel (approved):

PR #626, opened 2024-05-06
fix: improved `=` and `!=` with val `null` for HANA and Postgres

BobdenOs:
In the case of Postgres all queries doing `x = null` where being translated to `x is $1` which causes trouble as the operator name is `is null` unlike the SQLite operator which is just `is`.

The Postgres error was already resolved in the HANA Service, but the SQL Execution plan cache optimization caused trouble when comparing with `LOB` type columns. As HANA does not support compare operation on `LOB` columns. There for the only allowed operators on the `LOB` columns is `is (not) null`. Which also causes the issue that SQLite and Postgres are both allowed to execute `null = (LOB)`, but for HANA this causes a SQL error. As `null = ...` always returns null it will always filter everything out. Therefor the HANA Service translate it into `null = null` as the behavior is the same and does not throw a SQL error.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -832,17 +832,31 @@ class CQN2SQLRenderer {
   operator(x, i, xpr) {
 
     // Translate = to IS NULL for rhs operand being NULL literal
-    if (x === '=') return xpr[i + 1]?.val === null ? 'is' : '='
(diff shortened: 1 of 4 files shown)


patricebender (approved):

johannes-vogel (approved):

johannes-vogel:
makes sense to me

PR #627, opened 2024-05-06
deps: require `>= sap/cds@7.9.0`

patricebender:
this is needed for https://github.com/cap-js/cds-dbs/pull/590

deps: require `>= sap/cds-compiler@4.9`


johannes-vogel:
@patricebender we should upgrade the compiler version with this PR as well

johannes-vogel (approved):

PR #631, opened 2024-05-07
deps: `@sap/hana-client` optional peer dependency

swaldmann:
Currently noticing the huge package is still coming in `@sap/cds-dk` with `@sap/hdi-deploy` version 5 because`@cap-js/hana` defines it as a non-optional peer dependency.

### DIFF
--- a/hana/package.json
+++ b/hana/package.json
@@ -34,6 +34,11 @@
     "@sap/hana-client": ">=2",
     "@sap/cds": ">=7.6"
   },
+  "peerDependenciesMeta": {
+    "@sap/hana-client": {
+      "optional": true
+    }
+  },
   "cds": {
     "requires": {
       "kinds": {


BobdenOs:
The npm documentation is very un clear about its behavior in relation to peer dependencies. It states that it gives warnings when peer dependencies are missing. Which can be disabled by making the peer dependencies optional, but it doesn't mention installing peer dependencies. https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies

swaldmann:
It says that in the section about `peerDependencies`

> In npm versions 3 through 6, peerDependencies were not automatically installed, and would raise a warning if an invalid version of the peer dependency was found in the tree. As of npm v7, peerDependencies are installed by default.

patricebender (approved):
fine by me, @BobdenOs please confirm

PR #632, opened 2024-05-08
fix: Improve comparator check for combined and nested expressions

BobdenOs:


test/scenarios/bookshop/read.test.js
```diff
@@ -224,6 +224,12 @@ describe('Bookshop - Read', () => {
     )).to.have.nested.property('data.value.length', 0)
   })
 
+  test('Filter Books(complex filter in apply)', async () => {
+    const res = await GET(`/browse/Books?$apply=filter(((ID eq 201 or ID eq 252) and ((contains(tolower(descr),tolower('Edgar'))))))`)
```
johannes-vogel, line 228:
```suggestion
    const res = await GET(`/browse/Books?$apply=filter(((ID eq 251 or ID eq 252) and ((contains(tolower(descr),tolower('Edgar'))))))`)
```
201 does not include Edgar, only 251 and 252.

johannes-vogel (approved):

PR #633, opened 2024-05-08
chore: Streamlined eslint usage

danjoa:


patricebender (changes requested):
why remove the eslint config?

patricebender:
→ we need the dedicated eslint config in our repository, because outside contributors (without cap/dev setup) must be able to lint their project according to our guidelines.

PR #587, opened 2024-04-15
chore: release main

cap-bots:
:robot: I have created a release *beep* *boop*
---

<details><summary>db-service: 1.9.0</summary>

### Added

* Add missing `func` cqn structures ([#629](https://github.com/cap-js/cds-dbs/issues/629)) ([9d7539a](https://github.com/cap-js/cds-dbs/commit/9d7539ab0fc7e70a6a00c0bd9cb4b3e362976e16))

### Fixed

* **`order by`:** reject non-fk traversals of own columns in order by ([#599](https://github.com/cap-js/cds-dbs/issues/599)) ([3288d63](https://github.com/cap-js/cds-dbs/commit/3288d63f0ee6a96580a3b2138ecb24a944371cf1))
* Align all quote functions with @sap/cds-compiler ([#619](https://github.com/cap-js/cds-dbs/issues/619)) ([42e9828](https://github.com/cap-js/cds-dbs/commit/42e9828baf11ec55281ea634ce56ce93e6741b91))
* assign artificial alias if selecting from anonymous subquery ([#608](https://github.com/cap-js/cds-dbs/issues/608)) ([e1a7711](https://github.com/cap-js/cds-dbs/commit/e1a77119f0a5241cfe4f50a37a473f2325ba5bde))
* avoid spread operator ([#630](https://github.com/cap-js/cds-dbs/issues/630)) ([a39fb65](https://github.com/cap-js/cds-dbs/commit/a39fb65f9419fe60e0324741039d004b40082903))
* flat update with arbitrary where clauses ([#598](https://github.com/cap-js/cds-dbs/issues/598)) ([f108798](https://github.com/cap-js/cds-dbs/commit/f108798c6c8035f9cdd0b9c6b8f334f1454c2faa))
* improved `=` and `!=` with val `null` for HANA and Postgres ([#626](https://github.com/cap-js/cds-dbs/issues/626)) ([cbcfe3b](https://github.com/cap-js/cds-dbs/commit/cbcfe3b15e8ebcf7e844dc5406e4bc228d4c94c9))
* Improved placeholders and limit clause ([#567](https://github.com/cap-js/cds-dbs/issues/567)) ([d5d5dbb](https://github.com/cap-js/cds-dbs/commit/d5d5dbb7219bcef6134440715cf756fdd439f076))
[...]

PR #639, opened 2024-05-13
fix: Support multi byte characters

BobdenOs:


patricebender (approved):
I would omit the scope `hana`, as the PR title will be put into the @cap-js/hana changelog, it is not necessary to mention the database in the commit message. In https://github.com/cap-js/cds-dbs/pull/587/commits/34aa381c81bae76188f12b5cb8933f5d1d497cb8 I had to manually cleanup the changelog

test/compliance/resources/db/basic/literals/basic.literals.string.js
```diff
@@ -31,6 +31,9 @@ module.exports = [
   {
     char: 'A',
   },
+  {
+    char: '대', // Ensure multi byte utf-8 characters also fit into a single character column
```
oklemenz2, line 35: Are we sure, that the character that is not utf-8 converted is still the same, when read from database?
Maybe a test can be added, that INSERTed value is the same then the afterwards selected value...

hana/lib/HANAService.js
```diff
@@ -1028,6 +1028,7 @@ class HANAService extends SQLService {
       Int64: () => `BIGINT`,
       UUID: () => `NVARCHAR(36)`,
       Boolean: () => `NVARCHAR(5)`,
+      String: e => `NVARCHAR(${(e.length || 5000) * 4})`,
```
oklemenz2, line 1031: Is this enough for all use-cases? What is the max character that can occur...

PR #641, opened 2024-05-15
feat: Allow hex engine to be used

BobdenOs:
When using `SELECT FROM DUMMY FOR JSON` the hex engine falls back to the legacy engines, but when replacing it with an equivalent statement it is possible to have the queries run in the hex engine.

Any suggestions for `FROM DUMMY` replacements are welcome.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -206,7 +206,7 @@ class HANAService extends SQLService {
       })
 
     const withclause = withclauses.length ? `WITH ${withclauses} ` : ''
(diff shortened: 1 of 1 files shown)


BobdenOs:
Tests running inside the Github action are using legacy engine as `HXE` is using `SP2.0` which only includes the legacy engines.

patricebender (approved):
Cool way to create your own dummy table, I do not have an alternative 😁 It's only the combination of using the `DUMMY` table and feeding the `FOR JSON` function with the result, which causes the fallback to the legacy engine, isn't it?

johannes-vogel:
@BobdenOs merge?

PR #654, opened 2024-05-28
feat: simple queries

BobdenOs:
Adds `hana_simple_queries` feature flag. Which will make `@cap-js/hana` run all possible queries without `FOR JSON`. Which allows `hdb` and `@sap/hana-client` to returns different results again. 

Output converters are kept in place to reduce the likely hood of this happening. Current impacted test is `Bookshop - Order By › collations for aggregating queries with subselect`. Where the result of `sum(price) as pri` comes back as `"150"` for `@sap/hana-client` and `"1.5e+2"` for `hdb`. While expected is `150` in the `FOR JSON` result.

Additionally this PR removes the duplicate `"_json_"` output definition. Simplifying the SQL in all cases.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -133,8 +133,10 @@ class HANAService extends SQLService {
     const { cqn, sql, temporary, blobs, withclause, values } = this.cqn2sql(query, data)
     delete query.SELECT.expand
 
+    const isSimple = temporary.length + blobs.length + withclause.length === 0
+
     // REVISIT: add prepare options when param:true is used
-    const sqlScript = isLockQuery ? sql : this.wrapTemporary(temporary, withclause, blobs)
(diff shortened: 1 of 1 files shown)


BobdenOs:
This is for customers which believe these queries will be faster for their use cases. As the other databases are very fast in rendering json. There is no reason to do this. Also the output result is unstable.

oklemenz2:
@BobdenOs How is the case handled, where a query has an `where ID in <array>` clause. With new JSON concept only a single parameter was used and an json array was passed. Previously in old HANA adapter for each array entry a placeholder was used.

How is it now with your changes of this PR?

We liked the approach, that only a single placeholder is used because this would make the execution plan stable, no matter how many entries are in the array (which is often not under our control, as depending on user data)...

Having some toggles, that would provide different version would be good, so that we can experiment with that. 
For example also the Create was done with a single JSON object. How is this done now?

What do you think?

FYI @D069515

PR #655, opened 2024-05-28
test(`cds 8`): explicitly set `fewerLocalizedViews = false`

patricebender:
the `cds-compiler` will not produce `localized` views for transitive dependencies of `localized` artifacts.

That means an `Author` will not get a `localized.Authors` view just because there is an `author` association in the `localized` entity `Books`.

at some point those tests can be rewritten, but to test the deep replacement mechanism for `localized` artifacts it is fine for now to just explicitly set the `fewerLocalizedViews` option.

### DIFF
--- a/db-service/test/cqn4sql/expand.test.js
+++ b/db-service/test/cqn4sql/expand.test.js
@@ -282,36 +282,6 @@ describe('Unfold expands on associations to special subselects', () => {
     expect(JSON.parse(JSON.stringify(res))).to.deep.equal(expected)
   })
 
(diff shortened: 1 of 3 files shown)


bugwelle (approved):

BobdenOs:
Probably just remove these `transitive_` values. To make the linter happy.

stewsk (approved):

patricebender:
> Probably just remove these `transitive_` values. To make the linter happy.

ah youre right, will do

PR #664, opened 2024-05-29
fix: enable reference to query element in filter with `$self`

patricebender:
this is also supported by the cds-compiler.
With this change, it also works for queries during runtime.


patricebender:
don't do this, throw a nice error for `$self` in filter

PR #667, opened 2024-05-30
chore: Cleanup and fixes for cds8

danjoa:


danjoa:
Let's merge that now, please...

BobdenOs:
The `@capire/sflight` tests have the `IEEE754` header enabled by default for all `axios` requests, but the expectations don't expect `Decimal` values to be `strings` ([model](https://github.com/SAP-samples/cap-sflight/blob/bebfc86571f108e328d736b698f3ab0ef94d2782/db/schema.cds#L17) -> [expectation](https://github.com/SAP-samples/cap-sflight/blob/bebfc86571f108e328d736b698f3ab0ef94d2782/test/odata.test.js#L61)).

BobdenOs (approved):

danjoa:
> The `@capire/sflight` tests have the `IEEE754` header enabled by default for all `axios` requests, but the expectations don't expect `Decimal` values to be `strings` ([model](https://github.com/SAP-samples/cap-sflight/blob/bebfc86571f108e328d736b698f3ab0ef94d2782/db/schema.cds#L17) -> [expectation](https://github.com/SAP-samples/cap-sflight/blob/bebfc86571f108e328d736b698f3ab0ef94d2782/test/odata.test.js#L61)).

We indeed need to revisit that, and check these points:

- [ ] Why did that SFlight test work before?
- [ ] We need to revert the `/^90/.test(v)` again
- [ ] Tests should not have to monkey patch converters

PR #673, opened 2024-06-03
fix: Improved behavioral consistency between the database services

BobdenOs:
This PR contains a few fixes that where required to make the new compliance tests to pass:
- Decimal precision behavior is aligned between the databases when defined
- `@cap-js/hana` now properly supports scalar `SELECT` queries in the columns

The goal of this PR is to remove as much dependencies as possible now that `cds@8` provides the `cds-test` command.

- Removing `jest` by switching to `cds-test`
- Removed `jest.config.js` files
- Removing `chai` by switching to `cds-test`
- Adjust all tests to fit with the `cds-test` `expect` pattern
- Removing `HANA` test action
- Move `HXE` Github action steps into a reuse file
- Added a lock for `HANA` database creation
- Added a fallback for `Postgres` during database creation
- Removed compliance import tests
- Added symbolic link to the compliance test folder


.github/workflows/release-please.yml
```diff
@@ -21,56 +21,25 @@ jobs:
...
-      - name: Start HXE image
-        if: ${{ steps.release.outputs.releases_created }}
-        run: |
-          echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin;
-          { npm start -w hana; } &
-        env:
-          TAG: ${{ steps.find-hxe.outputs.TAG }}
-          IMAGE_ID: ${{ steps.find-hxe.outputs.IMAGE_ID }}
-      - run: npm test -w hana -- --maxWorkers=1
+        uses: ./.github/actions/hxe
```
patricebender, line 36: lgtm

patricebender (approved):
overall lgtm, could you please elaborate a little more what was done here in the description of this change?

danjoa (approved):

hana/package.json
```diff
@@ -16,23 +16,26 @@
...
     "@sap/hana-client": ">=2",
-    "@sap/cds": ">=7.6"
+    "@sap/cds": ">=7.6",
+    "hdb": "^0.19.5"
   },
   "peerDependenciesMeta": {
     "@sap/hana-client": {
       "optional": true
+    },
+    "hdb": {
```
johannes-vogel, line 37: please revert, hana-service should brind hdb as dep

hana/test/compliance
```diff
@@ -0,0 +1 @@
+../../test
```
johannes-vogel, line 1: Why is that required for compliance but not for bookshop?

hana/test/lean-draft.test.js
```diff
@@ -1,3 +1,4 @@
+require('../../test/cds.js')
```
johannes-vogel, line 1: Why do we need to import our cds extensions here?

test/compliance/INSERT.test.js
```diff
@@ -1,3 +1,5 @@
+require('@sap/cds')
```
johannes-vogel, line 1: why?

test/scenarios/sflight/integration.test.js
```diff
@@ -2,6 +2,11 @@ const childProcess = require('child_process')
 const path = require('path')
 const cds = require('../../cds')
 
+// REVISIT: @capire/sflight tests don't expect IEEE754 compliant responses, but do send the IEEE754 header according to the test
+// As it is not possible to configure the IEEE754 flag for @cap-js/hana it is required to inject this configurable:true
+// As the cqn2sql output converter injection doesn't allow for it to be overwritten at a later stage, but configurable will stay true
+Object.defineProperty(cds.builtin.types.Decimal.constructor.prototype, "CQN2HANA:convertOutput", { get: () => a => a, configurable: true })
```
johannes-vogel, line 8: What is this test suite doing if no tests are run?

hana/package.json
```diff
@@ -16,15 +16,15 @@
     "CHANGELOG.md"
   ],
   "scripts": {
-    "test": "npm start && jest --silent",
-    "test:remote": "jest --silent",
+    "test": "npm start && cds-test $(find -L test -type f -regex .*\\.test\\.js)",
```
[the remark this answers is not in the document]
BobdenOs: `cds-test` comes with its own `find` command, but this does not include the `-L` arguments which makes it include the `symbolic` links that are used for re using the `compliance` test suites. The regex just looks for `.test.js`, but as both `\` and `.` are special characters it has to be escaped like this.

hana/test/compliance
BobdenOs: this is a symbolic link inside the `hana/test` folder which makes it look like the root `test` folder is also located on `hana/test/compliance` and the root `test` folder contains `compliance`, `bookshop` and `sflight`. So it is required for `bookshop` as well as this includes the `bookshop`.

test/scenarios/sflight/integration.test.js
BobdenOs: This test suite is importing the `jest` tests from the `sflight` repository ([here](https://github.com/cap-js/cds-dbs/pull/673/files/7cfa9374865d51ecd2224d3b90983e3c3889b622#diff-345b318913531034f8b42e7b87680c450bc1c9a00b255f4ddd804f4299b505d9R24)). Which contain the wrong expectations as they claim to request `IEEE` compliance, but don't expect them and don't receive them.

hana/test/lean-draft.test.js
BobdenOs: When `cds` is not loaded in `describe` does not exist as a `global`

johannes-vogel: can't we omit the wrapping describe in the hana test file instead?
don't see a big value as we mainly want to run the same tests as on sqlite.

sqlite/lib/SQLiteService.js
```diff
@@ -220,7 +220,10 @@ class SQLiteService extends SQLService {
       Int64: cds.env.features.ieee754compatible ? expr => `CAST(${expr} as TEXT)` : undefined,
       // REVISIT: always cast to string in next major
       // Reading decimal as string to not loose precision
-      Decimal: cds.env.features.ieee754compatible ? expr => `CAST(${expr} as TEXT)` : undefined,
+      Decimal: cds.env.features.ieee754compatible ? (expr, elem) => elem?.scale
```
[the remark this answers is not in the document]
BobdenOs: While adjusting all the test it became clear that nothing knows what to expect for decimals. So there where many places that use `regex` definitions to accommodate for this. Now when a `Decimal` has a `scale` defined all database services will read according to the defined scale. Allowing for all the tests to be adjusted like [this](https://github.com/cap-js/cds-dbs/pull/673/files/7cfa9374865d51ecd2224d3b90983e3c3889b622#diff-a98702b5be88ee4621a15b2ec0b2fbe79ae4ce51b1a834e6437e76e743fab86dR246).

hana/package.json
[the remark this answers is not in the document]
johannes-vogel: I feel that symbolic links should become part of `cds-test`.

test/scenarios/sflight/integration.test.js
johannes-vogel: isn't the sflight tests to be adapted instead?

test/compliance/INSERT.test.js
BobdenOs: Defines the global `describe`

test/scenarios/sflight/integration.test.js
BobdenOs: yes that is why the `REVISIT` comment is there

hana/test/lean-draft.test.js
BobdenOs: Other option would be to move `lean-draft` to `compliance/scenarios/sflight`. Would probably be more fitting as it is basically an `sflight` test suite and it would run using the symbolic link mechanism.

johannes-vogel: probably makes more sense as effectively a single server start is sufficient for running the tests

sqlite/lib/SQLiteService.js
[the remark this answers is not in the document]
johannes-vogel: I think we should rather ask ourselves whether we want to be the ones that do number formatting. So far, we said that's up to the clients.

If databases stringify these values anyway, that seems like overhead at runtime that we only do for our tests.

hana/test/lean-draft.test.js
johannes-vogel: Will you do that before merging?

hana/package.json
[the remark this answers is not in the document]
johannes-vogel: Can you create a follow up for this?
Then we can merge before cds test works with symbolic links

test/scenarios/sflight/integration.test.js
johannes-vogel: Can you please also craete a follow up?
I don't find the time to prepare the tests.

db-service/package.json
```diff
@@ -22,7 +22,7 @@
     "CHANGELOG.md"
   ],
   "scripts": {
-    "test": "jest --silent"
+    "test": "cds-test --skip toMatchSnapshot"
```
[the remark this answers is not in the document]
johannes-vogel: please also create a follow up. Otherwise these tests are only run with our internal pipeline.

sqlite/lib/SQLiteService.js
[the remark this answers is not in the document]
BobdenOs: When large numbers are involved there is an arbitrary cut off point where `SQLite` switches its rendering style. Which is generally fine, but when we are serving the `IEEE754` spec it is problematic. So by providing a proper formatter it becomes controllable and will behave according to specification.

```sql
SELECT json_object('exp',10000000000000000000,'normal',1000000000000000000)
returns {"exp":1.0e+19,"normal":1000000000000000000}
parsed {exp: 10000000000000000000, normal: 1000000000000000000}

SELECT json_object('exp',cast(10000000000000000000 as TEXT),'normal',CAST(1000000000000000000 AS TEXT))
returns {"exp":"1.0e+19","normal":"1000000000000000000"}
parsed {"exp":"1.0e+19","normal":"1000000000000000000"}
```

db-service/package.json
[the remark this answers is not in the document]
BobdenOs: I personally don't see any value in the static snapshot tests. The problem of that databases is not their ability to render `SQL` it will always be determined by the integration with the database.

Here are some examples why the current state of the static `cqn2sql` tests don't provide any value.

1. test found a gap in the `CREATE` cqn got skipped instead of fixed ([code](https://github.com/cap-js/cds-dbs/blob/d25af70b23688cd22c7b87f20a7848c9653ae9a9/db-service/test/cqn2sql/create.test.js#L45))
1. tests the `not in` operator, but never found that it does not work on HANA ([code](https://github.com/cap-js/cds-dbs/blob/d25af70b23688cd22c7b87f20a7848c9653ae9a9/db-service/test/cqn2sql/delete.test.js#L57))
1. Uses legacy `cqn` format in many expression `['(',...,')']` ([code](https://github.com/cap-js/cds-dbs/blob/d25af70b23688cd22c7b87f20a7848c9653ae9a9/db-service/test/cqn2sql/delete.test.js#L80))

So while there is value in the `cqn` that the tests define. They are not actually being tested to determine functional correctness. Therefor the most beneficial action here would be to go over all the `cqn2sql` tests and move them into their respective `compliance` test suites.

sqlite/lib/SQLiteService.js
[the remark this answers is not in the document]
johannes-vogel: Is that only required for sqlite or for all dbs? atm, it's implemented for all.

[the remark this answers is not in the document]
johannes-vogel: Can you please refer to that change in PR title to make it appear in changelog?

PR #674, opened 2024-06-04
chore: Simple queries with lazy booleans

danjoa:
if `cds.env.features.sql_simple_queries` === ...
- 0 → always use json functions
- 1 → use json fns only if there are expands, structs/arrays or booleans
- 2 → use json fns only if there are expands, structs/arrays


oklemenz2:
How is currently a simple query defined?

What about `where x in <array>` clause? Is this simple or not. Because we liked the approach of only a single placeholder, passing a single JSON array (keep in the execution plan stable). But this of course then uses JSON...

Is there a specification, what a simple query is?

I would imagine a list of use-cases, where JSON function is used (like in the example above), and that would be then a complex query, right?

Is there such as list? 

Do we need fine granular toggles, when to use JSON and when not? We could then do some experiments with AFC, what behaves the best ;-)

db-service/lib/cqn2sql.js
```diff
@@ -9,6 +11,7 @@ const BINARY_TYPES = {
 }
 
 const { Readable } = require('stream')
+const { c } = require('@sap/cds/lib/utils/tar')
```
BobdenOs, line 14:
```suggestion
```

BobdenOs:
@oklemenz2 the `simple` references the complexity of the `result` of the query. So the query itself can be as complex as always, but the result it produces has to be `simple`. Which mostly means that it returns `flat` data. So no `expands` or `array of` data types. This makes the `complex` data type `boolean` for SQLite to be considered `simple` as well.

So the feature flag itself determines automatically whether it can remove the `json` functions from the query and still receive the same result from the driver directly.

oklemenz2:
So `where x in <array>` would still be a single placeholder with JSON array, right?

oklemenz2:
And a Create would be a single placeholder with JSON object, right?

BobdenOs:
The effect this feature toggle has is the following:

With the feature toggle `off` the databases always produce a SQL structure like:
```SQL
SELECT 
  json(.*) as "_json_" -- Postgres like syntax as it the easiest to understand
FROM (
  SELECT
    <columns>
    <expands>
  FROM
    <target>
  <clauses>
)
```

With the feature toggle `on` the databases try to produce only the sub query:
```SQL
SELECT
  <columns> -- <!expands>
FROM
  <target>
<clauses>
```

### Consistency

The big difference between these two statements is that the `_json_` column is compliant* with the `JSON` specification. Where as the result of the `simple` statement is determined by the driver and database protocol. For example `Decimal` when using `hana-client` will be `safe` javascript numbers or `unsafe` javascript strings. Where as with `hdb` it mostly will be a javascript string, but with `1.2e+2` syntax. While the `_json_` result is produced by the database and therefor is driver independent.

### JavaScript Objects

Another effect the `_json_` column has on drivers (e.g. `better-sqlite`). It removes the amount of converting the driver has to do between network / internal data structures to javascript objects. While `SQLite` has a very efficient native json/jsonb implementation to construct the result `c` string. The `better-sqlite` driver has to integrate with the `N-API` APIs to construct the javascript equivalent objects. So when returning the `_json_` column a single call to the `N-API` can be made to wrap the original raw memory into a `Buffer`. As `JSON.parse` is natively integrated into the `v8` source code. It has the most optimal methods of constructing javascript objects possible. 

This effect is also `documented` inside the `hana-client` source code ([npmjs](https://www.npmjs.com/package/@sap/hana-client?activeTab=code) `/lib/index.js:58:1`). `Done this way since the performance of creating N-API objects for the result set is very poor due to the N-API interaface itself`. 

This has the least effect on the Postgres node driver as the Postgres network protocol is very simple, but as it has native json support it automatically call `JSON.parse` in the driver.

### Skipping JavaScript Objects

With the nature of JavaScript being memory managed the amount of values created has a direct impact on overall performance. While this is in most cases not a problem. It has compounding effects when handling large amounts of JavaScript values. As the result inside the `_json_` column is focused on being OData v4 compliant it is possible to skip `JSON.parse` and greatly reduce the amount of JavaScript Values required to be created and cleaned up. This is only possible when the database result is directly consumed by the protocol adapter and does not have post processing like `after` handlers or `lean-draft`.

As all the databases have result set streaming implemented it prevents the whole result from being loaded into memory. Allowing smaller application instances to handle more requests at the same time. With the default `highWaterMark` being `16kb` ([docs](https://nodejs.org/api/stream.html#streamgetdefaulthighwatermarkobjectmode)) it *should* only require `~16mb` of memory to process `1000` parallel requests independent of the size of the result of the request. This behavior requires support from the protocol adapters so these benefits are not yet visible.

oklemenz2:
Thanks for this extensive explanations ;-)

BobdenOs (approved):

johannes-vogel:
@BobdenOs merge?

PR #676, opened 2024-06-06
feat: reject update/delete without where

SamuelBrucksch:
To discuss:
- singletons -> check needed?
- DELETE Foo(123)/bar -> would fail right now, should we check for to-one?


db-service/lib/SQLService.js
```diff
@@ -33,8 +41,8 @@ class SQLService extends DatabaseService {
           operation.columns ||
           Object.keys(
             operation.data || operation.entries?.reduce((acc, obj) => {
-              return Object.assign(acc, obj)
-            }, {}),
+                return Object.assign(acc, obj)
+              }, {}),
```
SamuelBrucksch, line 45:
```suggestion
              return Object.assign(acc, obj)
            }, {}),
```

schiwekM:
> To discuss:
> 
> * singletons -> check needed?
> * DELETE Foo(123)/bar -> would fail right now, should we check for to-one?

I'd say for singletons not needed and to-one should be excluded as it is spec compliant

db-service/lib/SQLService.js
```diff
@@ -21,6 +21,14 @@ const BINARY_TYPES = {
 
 class SQLService extends DatabaseService {
   init() {
+    this.on(['UPDATE', 'DELETE'], ({ query}, next) => {
+      const cqn = query.UPDATE ?? query.DELETE
+      if (!cqn.where && !cqn.from?.ref?.at(-1).where && !cqn.entity?.ref?.at(-1).where) {
```
danjoa, line 26:
```suggestion
      if (!cqn.where && !(cqn.from || cqn.entity)?.ref?.at(-1).where) {
```

db-service/lib/SQLService.js
```diff
@@ -21,6 +21,14 @@ const BINARY_TYPES = {
...
+        const op = query.DELETE ? 'delete' : 'update'
+        return cds.error(`Trying to ${op} all entites. Call .where(...) explicitely, to ${op} all entities.`)
```
danjoa, line 28:
```suggestion
        return cds.error(`${query.DELETE ? 'DELETE' : 'UPDATE'} without a where clause, which would affect all entites in the table. Add where 1=1 if your really want to do so`)
```

PR #678, opened 2024-06-07
fix: Improve error message for disconnected connections

BobdenOs:
`this._done` is used from the default `tx` behavior. This can contain `committed` or `rolled back`. If `this._done` is not set it means the connection was never connected.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -100,7 +100,7 @@ class HANAService extends SQLService {
   }
 
   ensureDBC() {
-    return this.dbc || cds.error`Database is disconnected`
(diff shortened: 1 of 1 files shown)


hana/lib/HANAService.js
```diff
@@ -100,7 +100,7 @@ class HANAService extends SQLService {
   }
 
   ensureDBC() {
-    return this.dbc || cds.error`Database is disconnected`
+    return this.dbc || cds.error`Database connection is ${this._done || 'disconnected'}`
```
patricebender, line 103:
```suggestion
    return this.dbc || cds.error`Database connection ${this._done ? 'is ' + this._done || 'failed'}`
```

patricebender (approved):
I think the error message I suggested reads better but no hard feelings

PR #681, opened 2024-06-11
feat: Add quoted mode support

BobdenOs:


db-service/lib/cqn2sql.js
```diff
@@ -34,6 +34,12 @@ class CQN2SQLRenderer {
     this.class = new.target // for IntelliSense
     this.class._init() // is a noop for subsequent calls
     this.model = srv?.model
+
+    // Overwrite smart quoting
+    if (cds.env.cdsc.sqlMapping === 'quoted') {
+      this.class.prototype.name = (name) => name.id || name
+      this.class.prototype.quote = (s) => `"${s.replace(/"/g, '""')}"`
+    }
```
patricebender, line 42: This is the setting which lets the compiler use quoted mode as well. Would it be useful to have a runtime independent setting for quotation?

BobdenOs: The Database Service has to follow the quoting behaviors of the compiler. So from my perspective it only makes sense to have them both react to the exact same configuration.

patricebender: 👍

patricebender (approved):

db-service/lib/cqn2sql.js
```diff
@@ -34,6 +34,12 @@ class CQN2SQLRenderer {
     this.class = new.target // for IntelliSense
     this.class._init() // is a noop for subsequent calls
     this.model = srv?.model
+
+    // Overwrite smart quoting
+    if (cds.env.cdsc.sqlMapping === 'quoted') {
```
danjoa, line 39: Use official setting `cds.env.sql.names` instead

danjoa (commented):
What exactly are we doing here?

ThePlenkov:
Hi! let me please just share the importance of such a feature. In our app we used quoted on purposes - because we wanted our artifacts to be created using namespace patterns as we used to have in Hana XSC. So now because of this we cannot use @cap-js/hana in our production app without significant refactoring. I'm afraid creating synonyms will not be enough. Therefore - we stay now with CAP native hana driver unless this mode is supported

db-service/lib/cqn2sql.js
```diff
@@ -36,9 +36,9 @@ class CQN2SQLRenderer {
     this.model = srv?.model
 
     // Overwrite smart quoting
-    if (cds.env.cdsc.sqlMapping === 'quoted') {
+    if (cds.env.sql.names === 'quoted') {
       this.class.prototype.name = (name) => name.id || name
-      this.class.prototype.quote = (s) => `"${s.replace(/"/g, '""')}"`
+      this.class.prototype.quote = (s) => `"${(s + '').replace(/"/g, '""')}"`
```
danjoa, line 41:
```suggestion
      this.class.prototype.quote = (s) => `"${String(s).replace(/"/g, '""')}"`
```
more obvious intent

db-service/lib/cqn2sql.js
```diff
@@ -336,7 +342,9 @@ class CQN2SQLRenderer {
     const { ref, as } = from
     const _aliased = as ? s => s + ` as ${this.quote(as)}` : s => s
     if (ref) {
-      const z = ref[0]
+      const { target } = SELECT.from(from)
```
danjoa, line 345: Why do we have to create a new query to get the target? 
Can't we just take it from the rendered query?

hana/lib/HANAService.js
```diff
@@ -600,11 +600,11 @@ class HANAService extends SQLService {
 
         // Calculate final output columns once
         let outputColumns = ''
-        outputColumns = `_path_ as "_path_",${blobs} as "_blobs_",${expands} as "_expands_",${jsonColumn} as "_json_"`
+        outputColumns = `${this.quote('_path_')} as "_path_",${blobs} as "_blobs_",${expands} as "_expands_",${jsonColumn} as "_json_"`
```
danjoa, line 603: Why do we have to quote `_path_`? Isn't that a name chosen by us anyways?

ThePlenkov:
Hi! Is there any progress with this PR? Thanks!

db-service/lib/cqn2sql.js
```diff
@@ -336,7 +342,9 @@ class CQN2SQLRenderer {
...
+      // localized views don't have @cds.persistence.name so all resolve to non localized entities
+      const z = target?.['@cds.persistence.name'] || ref[0]
```
johannes-vogel, line 347:
```suggestion
      let z = ref[0]
      if (cds.env.sql.names === 'quoted') {
        // use SELECT.from to infer query, cds.infer also expects a query
        const { target } = SELECT.from(from)
        z ??= target?.['@cds.persistence.name']
      }
```

hana/lib/HANAService.js
johannes-vogel: `_path_` is used as alias in CQN [here](https://github.com/cap-js/cds-dbs/blob/3e7dc2473e8298512f18dc07cb1d2402d7e8e3b8/hana/lib/HANAService.js#L419) hence also these snippets have to be quoted.

johannes-vogel (approved):

PR #683, opened 2024-06-12
chore: make use of optional catch binding

patricebender:
Optional catch binding is a feature introduced in ECMAScript 2019 (ES10) that allows you to use catch clauses in try/catch statements without specifying an error variable


schiwekM:
> _ No description provided. _

Can't you just leave out the argument to avoid this error, instead of adjusting linting?

patricebender:
> > _ No description provided. _
> 
> Can't you just leave out the argument to avoid this error, instead of adjusting linting?

you are right, that would be better :)

schiwekM (approved):

BobdenOs (approved):

PR #685, opened 2024-06-12
fix: iso function creation

johannes-vogel:
I've verified with this implementation that the creation of ISO function works during productive deployment.
However, the connection creation at runtime did not work because of the dynamic creation of the function at runtime. We should move that to the tests until there's a mechanism to hook into plain sql deployment.

@BobdenOs Could you check whether my change for test deployments is sufficient?

### DIFF
--- a/hana/cds-plugin.js
+++ b/hana/cds-plugin.js
@@ -8,34 +8,36 @@ if (cds.requires.db?.impl === '@cap-js/hana') {
   cds.env.sql.dialect = 'hana'
 }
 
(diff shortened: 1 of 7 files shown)


johannes-vogel:
I just checked the remaining failures and this is due to compliance create test suite to set _deployed to true  which skips the ISO function generation.

BobdenOs:
@johannes-vogel I will add a check that makes sure that the `ISO` function is deployed or not. If not it will be excluded from the input converters. I am currently adjusting the `Timestamp` and `DateTime` test data.

PR #689, opened 2024-06-14
fix: Binary columns now return as Buffer for HANAService

BobdenOs:


sjvans (approved):
lgtm

BobdenOs:
> What did they return before?

The binary type columns where accidentally returned as `Readable` as they where being confused a `LOB`

PR #690, opened 2024-06-14
chore: Remove Buffer.prototype.toJSON overwrite

BobdenOs:


danjoa (approved):

sebastien-savalle:
Hello, could you give more details related to this override ? 
We spent a lot of time to identify a regression on our side caused by this override.

When do you plan to release this fix ?

Thanks,
Sebastien

PR #697, opened 2024-06-18
chore: coercify decimals expects

sjvans:


David-Kunz (approved):

David-Kunz:
@BobdenOs , the tests are indeed testing a bit too much, I don't think there's a need to test _all_ properties. We could certainly remove some of those in the `$select` clause. It's only important that the `$select` clause is 1) taken into account and 2) properly handles draft properties. "Normal" properties don't need to be check extensively.

@sjvans, your PR is fine for me, alternatively you can just skip testing those fields, up to you.

PR #707, opened 2024-06-21
feat: Enable native HANA fuzzy search for `search` function queries

BobdenOs:
This is the initial change to leverage the `SCORE` function from HANA to perform fuzzy searches. This is a greatly simplified implementation of the `SCORE` function as it comes with a large amount of configuration options.

### DIFF
--- a/hana/lib/cql-functions.js
+++ b/hana/lib/cql-functions.js
@@ -27,7 +27,7 @@ const StandardFunctions = {
     // REVISIT: remove once the protocol adapter only creates vals
     if (Array.isArray(arg.xpr)) arg = { val: arg.xpr.filter(a => a.val).map(a => a.val).join(' ') }
     // REVISIT: make this more configurable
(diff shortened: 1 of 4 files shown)


hana/lib/cql-functions.js
```diff
@@ -24,14 +24,10 @@ const StandardFunctions = {
...
-      { toString } = ref
-    return (
-      '(CASE WHEN (' +
-      refs.map(ref2 => `coalesce(locate(${this.tolower(toString(ref2))},${this.tolower(arg)}),0)>0`).join(' or ') +
-      ') THEN TRUE ELSE FALSE END)'
-    )
+    // REVISIT: remove once the protocol adapter only creates vals
+    if (Array.isArray(arg.xpr)) arg = { val: arg.xpr.filter(a => a.val).map(a => a.val).join(' ') }
+    // REVISIT: make this more configurable
+    return (`(CASE WHEN SCORE('${arg.val}' IN ${ref} FUZZY MINIMAL TOKEN SCORE 0.5 minimal score 0.5 SEARCH MODE 'string') > 0.0 THEN TRUE ELSE FALSE END)`)
```
patricebender, line 30: 👍
could we add a test for this?

BobdenOs: We have a test for this as you can see it is failing.

johannes-vogel (approved):

PR #715, opened 2024-06-27
fix: rewrite assoc chains if intermediate assoc is not fk

patricebender:
in one case a ref `genre.parent.ID` was mistaken for a foreign key access because `ID` is fk of `parent`. But because `parent` is not fk of `genre`, it is actually not an fk access.
Hence, the reference actually needs to point to the correct join node.

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -1573,7 +1573,7 @@ function cqn4sql(originalQuery, model) {
           transformedFrom.args.push(f)
         }
       })
-      return { transformedFrom }
+      return { transformedFrom, transformedWhere: existingWhere }
     } else if (from.SELECT) {
       transformedFrom = transformSubquery(from)
       if (from.as) {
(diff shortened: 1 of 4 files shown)


db-service/lib/cqn4sql.js
```diff
@@ -1583,7 +1583,7 @@ function cqn4sql(originalQuery, model) {
         // select from anonymous query, use artificial alias
         transformedFrom.as = Object.keys(originalQuery.sources)[0]
       }
-      return { transformedFrom }
+      return { transformedFrom, transformedWhere: existingWhere }
```
patricebender, line 1586: this is a "side" fix → Dont ignore where on original query if we select from a subquery.

db-service/lib/infer/index.js
```diff
@@ -1198,6 +1200,8 @@ function infer(originalQuery, model) {
         firstStepIsEntity = true
         if (arg.$refLinks.length === 1) return `${cur.definition.name}`
         return `${cur.definition.name}`
+      } else if (cur.definition.SELECT) {
```
patricebender, line 1203: yield proper error message `<element> not found in <subquery>`

stewsk (approved):

PR #726, opened 2024-07-03
chore(hana-client): add warning for missing thread pool configuration

David-Kunz:
Also asked for review in mail thread

### DIFF
--- a/hana/lib/drivers/hana-client.js
+++ b/hana/lib/drivers/hana-client.js
@@ -4,7 +4,7 @@ const cds = require('@sap/cds')
 const hdb = require('@sap/hana-client')
 const { driver, prom, handleLevel } = require('./base')
 const LOG = cds.log('@sap/hana-client')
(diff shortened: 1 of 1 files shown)


BobdenOs (approved):

danjoa (approved):
Looks good 👍 
Could we add a learn more link to some docs by hana-client?

hana/lib/drivers/hana-client.js
```diff
@@ -17,6 +19,7 @@ class HANAClientDriver extends driver {
    * @param {import('./base').Credentials} creds The credentials for the HANAClientDriver instance
    */
   constructor(creds) {
+    if (process.env.NODE_ENV === 'production' && !process.env.HDB_NODEJS_THREADPOOL_SIZE && !process.env.UV_THREADPOOL_SIZE) LOG.warn("When using @sap/hana-client, it's strongly recommended to adjust its thread pool size with environment variable `HDB_NODEJS_THREADPOOL_SIZE`, otherwise it might lead to performance issues.")
```
BobdenOs, line 22: I think we should remove the `NODE_ENV` check and move it to the root of the file. So it only is printed on the first time connecting to the database. This file is only loaded when `@sap/hana-client` is actually used. That way it is not printed once.

David-Kunz: Sure, I can move everything

David-Kunz: done

PR #733, opened 2024-07-08
fix: add placeholder for string values

mariayord:
Fixes : cap/issues/issues/16220


johannes-vogel (approved):
@BobdenOs looks good to me? What do you think?

BobdenOs:
@johannes-vogel there are `val` scenarios where the final result will be a `string`, but at this point of the code it might still be a `Date` objects. So while this fulfills the original description. There might be more scenarios where they should be created as place holders. Additionally it might make sense to check whether the incoming `x` has `params` already defined as `true`. Which we should allow to overwrite this logic.

db-service/lib/cqn2sql.js
```diff
@@ -311,8 +311,8 @@ class CQN2SQLRenderer {
    */
   column_expr(x, q) {
     if (x === '*') return '*'
-    
-    let sql = this.expr({ param: false, __proto__: x })
+  
+    let sql = x.param !== true && (typeof x.val === 'number' || x.val === null) ? this.expr({ param: false, __proto__: x }): this.expr(x)
```
BobdenOs, line 315: Why the `null` it doesn't have a type. So whatever type the `?` is assigned by the database `null` is accepted and will be returned correctly as a `null`.

mariayord: In the case  `{ val: null, as: 'Null' },` it doesn't have a type. See [select with column aliases](https://github.com/cap-js/cds-dbs/blob/13ece9e29ea17c3ac52f36afd4e4932986bbb27d/db-service/test/cqn2sql/select.test.js#L463).

mariayord: @BobdenOs  Do you think the generated snapshot is wrong and in this case there should be a placeholder for `null`

BobdenOs: When changing behavior all snapshots are wrong by default. It is required to check the difference between the snapshots when changing the behavior. I would rather look at the integration tests whether the behavior is correct. If Postgres and HANA don't have problem with handling placeholders for `null` values in the select columns. I would interpretate `null` as placeholders.

mariayord: > It is required to check the difference between the snapshots when changing the behavior.
I checked this , the question is if the old result of the test is correct, with no placeholder for `null `value.
If you say it is not the problem for db to have a `null` placeholder, the test before was wrong!

test/scenarios/bookshop/read.test.js
```diff
@@ -53,6 +53,13 @@ describe('Bookshop - Read', () => {
     expect(columns).to.contain('title')
   })
 
+  test('Path expression with null value', async () => {
```
BobdenOs, line 61: 1. the test name is incorrect
2. there is already a test doing this as was shared with you before [here](https://github.com/cap-js/cds-dbs/blob/9f8906ef1e916aaff3bdcf3faed22740e814b7dc/test/compliance/SELECT.test.js#L45)

mariayord: 1. Please make a suggestion for the test name if you don't  like it.
2. This is an integration test (I suppose it should run with Hana, Postgres..), the other one is unit.

BobdenOs (approved):

Issue #708, opened 2024-06-25
odata `groupBy` causes `subquery uses ungrouped column "__select__.author_id" from outer query`


hakimio:
### Description of erroneous behaviour 

CDS server crashes with error `TypeError: transformedQuery.SELECT.from.ref is not iterable` when trying to `groupBy` on a child association.

Stacktrace:
```
[cds] - TypeError: transformedQuery.SELECT.from.ref is not iterable
    at expandColumn (node_modules\@cap-js\db-service\lib\cqn4sql.js:790:41)
    at handleExpand (node_modules\@cap-js\db-service\lib\cqn4sql.js:420:40)
    at node_modules\@cap-js\db-service\lib\cqn4sql.js:354:32
    at getTransformedColumns (node_modules\@cap-js\db-service\lib\cqn4sql.js:382:21)
    at transformSelectQuery (node_modules\@cap-js\db-service\lib\cqn4sql.js:156:41)
    at cqn4sql (node_modules\@cap-js\db-service\lib\cqn4sql.js:73:26)
    at PostgresService.cqn4sql (node_modules\@cap-js\db-service\lib\SQLService.js:375:12)
    at PostgresService.cqn2sql (node_modules\@cap-js\db-service\lib\SQLService.js:351:18)
    at PostgresService.onSELECT (node_modules\@cap-js\db-service\lib\SQLService.js:127:39)
    at PostgresService.onSELECT (node_modules\@cap-js\postgres\lib\PostgresService.js:314:18) {
  id: '1043785',
  level: 'ERROR',
  timestamp: 1719308262527
}
```

### Detailed steps to reproduce

Try to run an odata query similar to the following:
```
$apply: filter(foo eq 'bar')/groupby((user/name),aggregate(amount with sum as totalAmount))
```
If you change `groupby((user/name)` with `groupby((user_ID)` it works as expected.

### Details about your project

| Package | Version |
|:------------------|---------------------------------------|
| OData version     | v4                               |
| @cap-js/cds-typer      | 0.22.0                                                     |
| @cap-js/cds-types      | 0.2.0                                                      |
| @cap-js/postgres       | 1.8.0                                                      |
| @sap/cds               | 7.9.2                                                      |
| @sap/cds-compiler      | 4.9.4                                                      |
| @sap/cds-dk            | 7.9.3                                                      |
| @sap/cds-dk (global)   | 7.9.2                                                      |
| @sap/cds-fiori         | 1.2.4                                                      |
[...]

PR #734, opened 2024-07-08
fix: optimize foreign key access for expand with aggregations

patricebender:
this is another follow up for #708

-> if the expand column was a foreign key access, the column was not properly re-written which resulted in the reported error.

we now make use of the common functionality which ensure proper column re-write. Along the way, support for structured columns was added as well.


BobdenOs (approved):
Looks good. Only `delete fc.as` looks suspicious. I guess `fc` is already a clone.

patricebender:
> Looks good. Only `delete fc.as` looks suspicious. I guess `fc` is already a clone.

yeah true the function should ideally have a parameter which makes it possible to omit aliases.

patricebender:
> Looks good. Only `delete fc.as` looks suspicious. I guess `fc` is already a clone.

will do this separately

PR #736, opened 2024-07-08
fix: make search impl okra compatible

johannes-vogel:


db-service/lib/cql-functions.js
```diff
@@ -23,7 +25,16 @@ const StandardFunctions = {
   search: function (ref, arg) {
     if (!('val' in arg)) throw new Error(`Only single value arguments are allowed for $search`)
     // only apply first search term, rest is ignored
-    arg.val = arg.__proto__.val = arg.val.split(' ')[0].replace(/"/g, '')
+    let first = arg.val.split(' ')[0]
+    if (/^".+"$/s.test(first)) first = first.substring(1, first.length - 1)
+
+    if (cds.env.features.okra_compat) {
+      first = first.replace(/\\\\/g,"\\").replace(/\\"/g,'"')
```
BobdenOs, line 32: Should we do `JSON.parse(first)` ? As it looks like that is what is going one here.

PR #739, opened 2024-07-10
fix: for aggregated `expand` always set explicit alias

patricebender:
to have the proper result structure, we need to rename foreign key references in the inner `expand` subquery to match the element name of the target.

instead of `author { <outer>.author_ID }` it needs to be `author { <outer>.author_ID as ID` to match the expected result structure.

for #708

### DIFF
--- a/test/scenarios/bookshop/read.test.js
+++ b/test/scenarios/bookshop/read.test.js
@@ -40,13 +40,27 @@ describe('Bookshop - Read', () => {
   })
 
   test('Books with groupby with path expression and expand result', async () => {
-    const res = await GET('/admin/Books?$apply=filter(title%20ne%20%27bar%27)/groupby((author/name),aggregate(price with sum as totalAmount))', admin)
+    const res = await GET(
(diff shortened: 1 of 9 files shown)


johannes-vogel (approved):

test/scenarios/bookshop/read.test.js
```diff
@@ -40,13 +40,27 @@ describe('Bookshop - Read', () => {
...
-    const res = await GET('/admin/Books?$apply=filter(title%20ne%20%27bar%27)/groupby((author/name, author/ID),aggregate(price with sum as totalAmount))', admin)
+    const res = await GET(
+      '/admin/Books?$apply=filter(title%20ne%20%27bar%27)/groupby((author/name, author/ID),aggregate(price with sum as totalAmount))',
+      admin,
+    )
     expect(res.data.value.length).to.be.eq(4) // As there are two books which have the same author
+    expect(
+      res.data.value.every(
+      item =>
+        Object.prototype.hasOwnProperty.call(item, 'author') &&
```
BobdenOs, line 59: Why not `'author' in item ...` ?

PR #749, opened 2024-07-15
chore: Improve debugging information for `INSERT` entries

BobdenOs:
Currently when setting `DEBUG=sql` it will print `Readable ...` for `entries` and `rows`. With this change the debug information will show the source `entries` and `rows` when they where provided by the original query. In the case the entries where already a `Readable` from the start it will simply print `[object Readable]` to keep the logs readable. 

Additionally one level of nesting has been removed so when using `jest` or `node:util` to format the `rows` it shows the data instead of `[[[[Array]]]]`. As the default nested printing level seems to be `3`.


patricebender:
could also be considered a fix, IMO :)

David-Kunz (approved):

johannes-vogel:
@BobdenOs Do you know what's missing here?

johannes-vogel:
Screenshot 2024-07-25 at 16 23 36

it seems this PR causes a race condition. @BobdenOs Can you double check?

BobdenOs:
> Screenshot 2024-07-25 at 16 23 36
> it seems this PR causes a race condition. @BobdenOs Can you double check?

But it can't as it doesn't change the behavior. It only stores a value on the steam object.

patricebender:
I would vote that we disable this test for the time being, it happens quite frequently. We should come up with a stable test instead.

PR #753, opened 2024-07-19
Fix search with multiple values

larslutz96:
Fixes search for search terms with multiple values like: `'/admin/Books?$search=foo bar'`

### DIFF
--- a/hana/lib/cql-functions.js
+++ b/hana/lib/cql-functions.js
@@ -25,7 +25,7 @@ const StandardFunctions = {
   concat: (...args) => `(${args.map(a => (a.xpr ? `(${a})` : a)).join(' || ')})`,
   search: function (ref, arg) {
     // REVISIT: remove once the protocol adapter only creates vals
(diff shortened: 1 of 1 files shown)


BobdenOs:
Please prevent duplicate or parallel work on the same task: https://github.com/cap-js/cds-dbs/pull/736

johannes-vogel:
> Please prevent duplicate or parallel work on the same task: #736

I think that's a bit different. #736 was for sqlite and likely solved with #754.
This PR here likely provides compat for okra + cap-js/hana.

patricebender (commented):
should we have a test for this?

BobdenOs:
@patricebender We had to disable the `$search` tests, because `HXE` doesn't support `SCORE`.

llanghammer-sap:
Hey all, 

Is there any update on this topic? On Fiori Element List Reports the Search is usually a standard feature and as of now any multi value search will result in an unexpected error.

PR #754, opened 2024-07-19
fix: quotations in vals

vkozyura:


db-service/lib/cql-functions.js
```diff
@@ -23,7 +23,7 @@ const StandardFunctions = {
   search: function (ref, arg) {
     if (!('val' in arg)) throw new Error(`Only single value arguments are allowed for $search`)
     // only apply first search term, rest is ignored
-    arg.val = arg.__proto__.val = arg.val.split(' ')[0].replace(/"/g, '')
+    arg.val = arg.__proto__.val = arg.val.split(' ')[0].replace(/^["](.*)["]$/s, "$1")
```
BobdenOs, line 26: The combination of `split(' ')` and `"` logic afterwards is definitely going to create behaviors which users don't expect.

Here the `regex` looks for quoted strings with escaped quotes inside or just takes anything that is not white spaces. The `JSON.parse` will correctly remove `\` characters where applicable. It should probably be caught if `JSON.parse` throws.
```suggestion
    const sub= /("")|("(?:[^"]|\\")*[^\\]")|(\S*)/.exec(arg.val)
    arg.val = arg.__proto__.val = (sub[1] ? JSON.parse(sub[1]) : sub[2]) || ''
```

vkozyura: done

BobdenOs (approved):

PR #757, opened 2024-07-22
fix: custom cqn INSERT with alias

larslutz96:
Add alias to INSERT statement in case of custom CQN INSERT with alias.


BobdenOs (commented):
But does it change the behavior?

larslutz96:
> But does it change the behavior?

No, this should not change anything.
I added it just because of a test and to look if the changes have any impact on other tests.

PR #758, opened 2024-07-24
deps: build script generates cds8 dependency in deployer app

patricebender:
this should not impact the dependencies of the actual node project, hence this is not a breaking change


LotharBender (approved):

johannes-vogel:
@BobdenOs it seems there's a race condition with your previous PR, can you please check?

PR #770, opened 2024-08-05
chore: Migrate to docker compose v2

BobdenOs:
`docker-compose` (v1) has been replaced for a little while now with `docker compose` (v2), but for `Docker Desktop` there is an alias `docker-compose` which actually is (v2) which created a miss match between the GitHub Action and the local development setup as the container name changed from `hana_1` to `hana-1`. 

This PR aligns the two by switching to `docker compose` (v2) for the GitHub Actions as well. As the GitHub Action default runner seems to have removed the `docker-compose` (v1) command.

### DIFF
--- a/hana/tools/docker/hce/ready.sh
+++ b/hana/tools/docker/hce/ready.sh
@@ -1,5 +1,5 @@
-docker exec hce_hana_1 /bin/bash -c "while ! ./check_hana_health ; do sleep 10 ; done;"
-docker exec -it hce_hana_1 /bin/bash -c "\
+docker exec hce-hana-1 /bin/bash -c "while ! ./check_hana_health ; do sleep 10 ; done;"
+docker exec -it hce-hana-1 /bin/bash -c "\
 cd /usr/sap/H00/HDB00;\
 . ./hdbenv.sh;\
 hdbuserstore -i SET SYSDBKEY localhost:30013@SYSTEMDB SYSTEM Manager1;\
--- a/hana/tools/docker/hce/start.sh
+++ b/hana/tools/docker/hce/start.sh
@@ -1,6 +1,6 @@
 exists=$(docker images hana-master:current -q);
 if [ $exists ]; then
-    docker-compose -f hana.yml up -d;
+    docker compose -f hana.yml up -d;
     ./ready.sh;
 else
     ./update.sh;
--- a/hana/tools/docker/hxe/ready.sh
+++ b/hana/tools/docker/hxe/ready.sh
@@ -1,9 +1,9 @@
-until docker cp ./start-hdi.sql hxe_hana_1:/usr/sap/HXE/start-hdi.sql
+until docker cp ./start-hdi.sql hxe-hana-1:/usr/sap/HXE/start-hdi.sql
 do
   sleep 1
 done
 
-docker exec hxe_hana_1 bash -c "until /check_hana_health -n -e ready-status > /dev/null; do sleep 1; done;"
+docker exec hxe-hana-1 bash -c "until /check_hana_health -n -e ready-status > /dev/null; do sleep 1; done;"
 echo "HANA has started"
-docker exec hxe_hana_1 bash -c "/usr/sap/HXE/HDB90/exe/hdbsql -i 90 -d SYSTEMDB -u SYSTEM -p Manager1 -I /usr/sap/HXE/start-hdi.sql > /dev/null && sleep 10"
+docker exec hxe-hana-1 bash -c "/usr/sap/HXE/HDB90/exe/hdbsql -i 90 -d SYSTEMDB -u SYSTEM -p Manager1 -I /usr/sap/HXE/start-hdi.sql > /dev/null && sleep 10"
 echo "HDI has been enabled"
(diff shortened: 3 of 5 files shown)


sjvans (approved):

BobdenOs:
Hi @gregorwolf,

The HANA Express images are publicly available [here](https://hub.docker.com/r/saplabs/hanaexpress/tags) (Last pushed 25 days ago). It might not be clear in the YAML files as for action startup times we use a pre-started image. Which saves `~4` minutes of start up time.

I specifically decided against using a HANA Cloud instance, because of the following reasons:
- It is easy to leak secrets ([docs](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections))
- Secrets are not available inside PRs from forks (*1)
- Having HANA Express as baseline allows anyone to contribute to the project
- Using a local HANA instance greatly improves test performance (*2)
- It would require a special HANA Cloud instance for this repository (*3)
- (redacted)

*1 It is possible to enable secrets in PRs from forks, but this makes leaking secrets trivial.

*2 As GitHub Actions are a black box to us. It is not possible to determine where the Action will be executed. So even when using a HANA Cloud instance that is running on Azure. There is no guarantee that the HANA Cloud instance and the Action runner will be in the same data center / continent. Which comes with a large impact on the networking performance as a side effect of how the HANA network protocol is defined. Where internal test suites have shown a variance in runtime of 3x the base line. While local HANA Cloud instances have around 0.1x variance in runtime. Even when running on the default GitHub Action runners which don't meet the minimal requirements of HANA Express. Which means that there is no need to set the test timeout to 60seconds which means that if something goes wrong the test run will fail quickly.

*3 While we have access to development and testing HANA Cloud instances it would be required to run an additional instance to just serve this repository. As it is possible for anyone to make PR and push commits it would be required that the HANA Cloud instance serving these GitHub Actions would scale accordingly. Which is currently difficult to do as scaling up an HANA Cloud instance takes quite some time and additional privileges (BTP). Which if done using existing HANA Cloud instances would create the cascading effect of preventing internal PRs to fail when the HANA Cloud instance is being attacked through this method. Effectively creating a Denial of Development attack on this repository and all internal repositories that share the HANA Cloud instance.

Sorry for the long explanation, but I hope it provides some insight to the decision.

Regards,

gregorwolf:
Thank you @BobdenOs for the details. Seems the search for https://hub.docker.com/search?q=hanaexpress is not providing the best result list. saplabs/hanaexpress appears, but it's not the top result and even after a imaged named the same way. Do you have any contact to the saplabs team maintaining the image? Maybe they can push Dockerhub to get the search result fixed.

BobdenOs:
@gregorwolf I haven't had any direct contact with the SAP labs colleagues. I am also not sure how big of a deal it is that the image is not Search engine optimized. Especially as the first hit on Docker Hub has `1B+` downloads. While the SAP labs image has `10K+` downloads. Even the 6 year old image with the same name from an arbitrary user has `100k+` downloads.

Additionally when using a proper search engine the image page is the first hit(s) ([google](https://www.google.com/search?q=docker+hub+hana+express&sca_esv=e9f3e6ede1937afe&source=hp&ei=qK-wZqesMdCVxc8P34SvmA8&iflsig=AL9hbdgAAAAAZrC9uNlk-0kmwZDbJiQcFfD76zgklT1x&ved=0ahUKEwinoOai2N2HAxXQSvEDHV_CC_MQ4dUDCBA&uact=5&oq=docker+hub+hana+express&gs_lp=Egdnd3Mtd2l6Ihdkb2NrZXIgaHViIGhhbmEgZXhwcmVzczIFECEYoAEyBRAhGKABSMMWUABYyxVwAHgAkAEAmAGAAaABrRGqAQQxNy42uAEDyAEA-AEBmAIXoALUEcICCxAuGIAEGLEDGIMBwgILEAAYgAQYsQMYgwHCAg4QABiABBixAxiDARiKBcICERAuGIAEGLEDGNEDGIMBGMcBwgIOEC4YgAQYsQMYgwEYigXCAg4QLhiABBixAxjRAxjHAcICCBAuGIAEGLEDwgIFEC4YgATCAggQABiABBixA8ICBRAAGIAEwgIGEAAYFhgewgIHEAAYgAQYE8ICCBAAGBMYFhgewgIIEAAYgAQYogTCAgcQIRigARgKmAMAkgcEMTcuNqAHpoYB&sclient=gws-wiz), [bing](https://www.bing.com/search?q=docker+hub+hana+express&form=QBLH&sp=-1&ghc=1&lq=0&pq=docker+hub+hana+expre&sc=6-21&qs=n&sk=&cvid=28FF34949F0C431EA37677C12C967205&ghsh=0&ghacc=0&ghpl=)). Sure Bing has 4 hits in the top 5 results that point to instructions on how to install deleted version of HANA Express using Docker. Not sure how much can be done about that at this point.

It is still a goal of mine to expose the HANA Express and HANA Cloud local development scripts from the `@cap-js/hana` package and the GitHub Action. To make it significantly simpler for developers to setup their own CI/CD and local development setups with full fledges HANA features.

PR #774, opened 2024-08-05
fix: Enforce `LargeStrings` to be `Strings` for `sql_simple_queries=1`

BobdenOs:
when using `sql_simple_queries=1` and `hdb` the `NCLOB` columns are returned as `Buffers` not as `Strings`. With this change these columns include an output converters that informs `hdb` to handle them as `Strings`.


hana/lib/HANAService.js
```diff
@@ -1124,6 +1124,7 @@ class HANAService extends SQLService {
 
     static OutputConverters = {
       ...super.OutputConverters,
+      LargeString: cds.env.features.sql_simple_queries === 1 ? e => `TO_NVARCHAR(${e})` : undefined,
```
soccermax, line 1127: what about `cds.env.features.sql_simple_queries: 2`?
```suggestion
      LargeString: cds.env.features.sql_simple_queries > 0 ? e => `TO_NVARCHAR(${e})` : undefined,
```

johannes-vogel (approved):

PR #781, opened 2024-08-21
fix(deep): prevent false unique constraint errors

David-Kunz:


db-service/lib/deep-queries.js
```diff
@@ -45,17 +45,17 @@ async function onDeep(req, next) {
...
   const queries = getDeepQueries(query, beforeData, target)
-  const res = await Promise.all(queries.map(query => {
-    if (query.INSERT) return this.onINSERT({ query })
-    if (query.UPDATE) return this.onUPDATE({ query })
-    if (query.DELETE) return this.onSIMPLE({ query })
-  }))
+  // first delete, then update, then insert because of unique constraints
+  await Promise.all(queries.filter(q => q.DELETE).map(query => this.onSIMPLE({ query })))
+  await Promise.all(queries.filter(q => q.UPDATE).map(query => this.onUPDATE({ query })))
+  const inserts = await Promise.all(queries.filter(q => q.INSERT).map(query => this.onINSERT({ query })))
```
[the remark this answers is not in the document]
David-Kunz: Yes, I prefer the latter

db-service/lib/deep-queries.js
```diff
@@ -246,7 +258,24 @@ const _getDeepQueries = (diff, target, root = false) => {
...
+        .filter(key => !target.keys[key].virtual && !target.keys[key].isAssociation)
+
+      const keyVals = keys.map(k => ({ val: diffEntry[k] }))
+      const currDelete = deletes.get(target.name)
+      if (currDelete)
+        keys.length === 1
+          ? currDelete.DELETE.where[2].list.push(...keyVals)
+          : currDelete.DELETE.where[2].list.push({ list: keyVals })
+      else {
+        // if we have only one key, queries should be simpler
```
BobdenOs, line 274: I don't see the need for this simplification. Additionally this comes with an additional side effect for HANA where it will create a different execution plan if other parts of `CAP` don't add this simplification.

The final SQL different will be this for HANA:
 ```SQL
DELETE WHERE ID in JSON_TABLE(?,'$' COLUMNS(VAL STRING PATH '$.val'))

DELETE WHERE (ID) in JSON_TABLE(?,'$' COLUMNS(VAL1 STRING PATH '$.val1'))
```

So while I like the sentiment it might not achieve what it is trying to achieve. I would rather investigate what the actual behavior is before adding this additional logic.

File: db-service/lib/deep-queries.js [hunk 3]
```diff
@@ -246,7 +258,24 @@ const _getDeepQueries = (diff, target, root = false) => {
...
+
+      const keys = cds.utils
+        .Object_keys(target.keys)
+        .filter(key => !target.keys[key].virtual && !target.keys[key].isAssociation)
+
+      const keyVals = keys.map(k => ({ val: diffEntry[k] }))
+      const currDelete = deletes.get(target.name)
+      if (currDelete)
+        keys.length === 1
+          ? currDelete.DELETE.where[2].list.push(...keyVals)
```
BobdenOs, line 271:
```suggestion
          ? currDelete.DELETE.where[2].list.push(keyVals[0])
```

db-service/lib/deep-queries.js
```diff
@@ -195,7 +203,10 @@ const getDeepQueries = (query, dbData, target) => {
     diff = [diff]
   }
 
-  return _getDeepQueries(diff, target, true)
+  const deletes = new Map()
```
BobdenOs, line 206: I like the idea, but it shows a clear problem in the overall implementation. As the `INSERT` map is not re-used and rather generated after each `_getDeepQueries` call. So from my perspective it would be best if `DELETE` and `INSERT` both use the same approach and when taking this approach it should probably also just include `UPDATE` and return all of them as separate results as this would also remove the additional split on the called of `getDeepQueries`.

test/compliance/UPDATE.test.js
```diff
@@ -49,4 +50,93 @@ describe('UPDATE', () => {
       throw new Error('not supported')
     })
   })
+
+  describe('uniques in deep updates', () => {
+    cds.test(__dirname + '/resources')
```
BobdenOs, line 55: Please move the `describe` so that the file only has a single `cds.test`. It works for `cds-dbs`, but it does an additional deployment which should not be required for this test suite.

test/compliance/UPDATE.test.js
```diff
@@ -49,4 +50,93 @@ describe('UPDATE', () => {
...
+                ID: 5,
+                number: 1, // would fail without the update below first
+              },
+              {
+                ID: 3,
+                number: 999, // 1 -> 999
+              },
+            ],
+          }),
+        )
```
BobdenOs, line 138: With the amount of duplicate data being defined it is very difficult to identify what is the actual changes that are being applied. So I would go for an approach that re uses the original data to make it a bit easier to follow what the test is trying to do. While keeping the comments that explain the conflicts the data is trying to create / avoid.

```suggestion
        const data = {
              ID: 1,
              title: 'foo',
              pages: [
                {
                  ID: 1,
                  number: 1,
                },
                {
                  ID: 2,
                  number: 2,
                },
              ],
            }
        await tx.run(
          INSERT.into(BooksUnique).entries([ data ]),
        )
        
        // Create new entries with conflicting numbers
        data.pages[0].ID = 3
        data.pages[1].ID = 4
        await tx.run(UPDATE(BooksUnique).data(data)) // first, old entries are deleted, so no violation
        
        data.pages[0].ID = 5
        data.pages[0].number = 1  // would fail without the update below first
        data.pages[1].number = 999
        await tx.run(UPDATE(BooksUnique).data(data))
```

BobdenOs (commented):
Looks functionally correct, but shows the weaknesses in the existing implementation.

db-service/lib/deep-queries.js
David-Kunz: Yes, did the same as in lean draft impl, if the SQL queries are nice with the additional list, we can keep it like that also for single keys

db-service/lib/deep-queries.js
David-Kunz: Yes, we could do that for `INSERT`, too

test/compliance/UPDATE.test.js
David-Kunz: I found the first one a bit more readable, but this change is fine for me

db-service/lib/deep-queries.js
David-Kunz: I just noticed that it's a little bit more subtle:
There's a query hierarchy which needs to be taken into account w.r.t. managed fields. If a given entity has sub queries (i.e. queries for its children), it is considered 'modified' and the managed fields (`@cds.on.update`) must be updated. That is currently wrong in this PR (deleting children wouldn't count as 'modified').

I think the simplest solution would be to keep the query hierarchy intact and collect same-target queries afterwards, as it's already done for `INSERT`.

A more performant solution would be to introduce a 'dirty' flag to mark modified children. I'll think about this.

David-Kunz: prototype solution is https://github.com/cap-js/cds-dbs/blob/fff3ed46f42016c9a5c7d440289ebcaf7391e206/db-service/lib/deep-queries.js#L215

(still need to fix a bug with root insert result)

db-service/lib/deep-queries.js
```diff
@@ -205,16 +209,17 @@ const _hasManagedElements = target => {
 /**
  * @param {unknown[]} diff
  * @param {import('@sap/cds/apis/csn').Definition} target
+ * @param {Map<String, Object>} deletes
```
patricebender, line 212: why is `updates` not a map?:)
```suggestion
 * @param {Map<String, Object>} deletes
 * @param {Map<String, Object>} inserts
 * @param {Object[]} updates
```

David-Kunz: Yeah, it's a bit inconsistent, but for updates, I can't group by target.name :(

David-Kunz: and constructing an identifier for each key combination seems a bit excessive :)

patricebender (approved):

PR #784, opened 2024-08-23
fix: Add `{list:[{val}]}` json function support

BobdenOs:


johannes-vogel (approved):

hana/lib/HANAService.js
```diff
@@ -977,6 +977,13 @@ class HANAService extends SQLService {
         this.values.push(JSON.stringify(list.list.map(l => l.list.reduce((l, c, i) => { l[`V${i}`] = c.val; return l }, {}))))
         return `(SELECT * FROM JSON_TABLE(?, '$' COLUMNS(${extraction})))`
       }
+      // If the list only contains of vals it is replaced with a json function and a placeholder
+      if (this.values && first.val) {
```
David-Kunz, line 981: Hi @BobdenOs , for the[ PR about combining deep DELETE statements](https://github.com/cap-js/cds-dbs/pull/781), I also send it for multiple keys, like

```
      const left = { list: keys.map(k => ({ ref: [k] })) }
      const op = not ? ['not', 'in'] : ['in']
      const right = { list: dataArray.map(r => ({ list: keys.map(k => ({ val: r[k] })) })) }
      return [left, ...op, right]
``` 

Would that be also something you could move into one placeholder? If not, what's the max. amount of those for HANA?

David-Kunz: Also happens in lean draft

David-Kunz: Ok, just saw the code above

PR #787, opened 2024-08-27
test: Add tests for computed elements

BobdenOs:


johannes-vogel:
Is there also a runtime query that tests this?

BobdenOs:
> Is there also a runtime query that tests this?

All the objects defined in the `module.exports` are being inserted into the table definitions. Then afterwards they are being read and checked for correctness. Where all properties starting with `=` are used as expectation when the inserted property is different then the returned result. ([test](https://github.com/cap-js/cds-dbs/pull/781#pullrequestreview-2271712503), [example](https://github.com/cap-js/cds-dbs/actions/runs/10577143838/job/29304451893#step:6:80))

patricebender (approved):

PR #789, opened 2024-08-27
deps: requires @sap/cds >= 8.2

danjoa:
Requires next release of @sap/cds with internal PR 4589 merged


BobdenOs (approved):

johannes-vogel:
we should also include cds version bump in hana service

PR #800, opened 2024-09-07
fix: dont use virtual key for `UPDATE … where (<key>) in <subquery>`

patricebender:
For a draft enabled model:

```cds
namespace my;

entity Books {
  key ID : Integer;
  title  : String;
  stock  : Integer;
  author : Association to Authors;
}

entity Authors {
  key ID : Integer;
  name   : String;
  alive  : Boolean;
}

service CatalogService {
    @odata.draft.enabled
    @readonly entity Books as projection on my.Books;

    @readonly entity Authors as projection on my.Authors;
}
```

`my.Books` will have a field:
```json
        "IsActiveEntity": {
          "type": "cds.Boolean",
          "key": true,
          "default": {
            "val": true
          },
          "@UI.Hidden": true
        },
```

which is **marked as virtual by the lean draft implementation**. This virtual key must not be part of the primary key matching which is done for a path expression in a `UPDATE.where`:

https://github.com/cap-js/cds-dbs/blob/982b8b796da4b577c5641039d2036816209c0437/db-service/test/cqn4sql/UPDATE.test.js#L73-L97

TODO:

- [x] test this


patricebender:
also check if UPDATEs on tables w/o primary keys cause issues

db-service/test/cqn4sql/model/update.cds
```diff
@@ -0,0 +1,23 @@
...
+
+entity Authors {
+  key ID : Integer;
+  name   : String;
+  alive  : Boolean;
+}
+
+service CatalogService {
+    @odata.draft.enabled
+    @readonly entity Books as projection on bookshop.Books;
```
David-Kunz, line 20:
```suggestion
    entity Books as projection on bookshop.Books;
```

Not relevant for this test, but it's cleaner (readonly draft-enabled entities don't make much sense)

David-Kunz (approved):
lgtm 🎸

PR #803, opened 2024-09-10
fix(deep-queries): properly return insert result

David-Kunz:
For `INSERT`s, `beforeData` is always `[]` and

```js
return beforeData.length ?? rootResult
```

would always return `0`.


patricebender (commented):
could you add a test for this scenario?

oklemenz2 (approved):

David-Kunz:
@patricebender , done!

patricebender (approved):

BobdenOs (approved):

PR #806, opened 2024-09-11
fix: reject all path expressions w/o foreign keys

patricebender:
reject all path expressions which navigate along a managed association which does not provide foreign keys. In this case, we can't construct join conditions, where exists subqueries or correlated subqueries for expands.


db-service/test/cqn4sql/keyless.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+  beforeAll(async () => {
+    model = await cds.load(__dirname + '/model/keyless').then(cds.linked)
+  })
+
+  it('no foreign keys for join', () => {
+    const { Books } = model.entities
+    const q = SELECT.from(Books).where(`author[ID = 42].book[ID = 42].author.name LIKE 'King'`)
+    expect(() => cqn4sql(q, model)).to.throw(
+      'Path step “author” of “author[…].book[…].author.name” has no valid foreign keys',
```
stewsk, line 20:
```suggestion
      'Path step “author” of “author[…].book[…].author.name” has no foreign keys',
```
to make it consistent with the other messages

patricebender: good catch :)

db-service/test/cqn4sql/model/keyless.cds
```diff
@@ -0,0 +1,17 @@
+// path expressions along `Authors:author` are not possible
```
stewsk, line 1:
```suggestion
// path expressions along `Books:author` are not possible
```

db-service/test/cqn4sql/model/keyless.cds
```diff
@@ -0,0 +1,17 @@
...
+entity Books {
+  key ID : Integer;
+  title  : String;
+  stock  : Integer;
+  author : Association to Authors;
+  authorWithExplicitForeignKey: Association to Authors { ID };
```
stewsk, line 7: If there was a calculated element
```
  authorName : author.name;
```
Would selecting `authorName` in a query have to be handled separately?

stewsk (approved):

db-service/test/cqn4sql/model/keyless.cds
patricebender: good point, I added this as a test case → does not need to be handled separately :)

PR #808, opened 2024-09-12
chore: use `for … of` + `if` instead of arrow func chaining

patricebender:
[cap/dev] etc/perf.js

for (each of keys) if ...                  196 ms
for (each of keys) || ...                  188 ms
keys.filter.forEach (...)                  323 ms


David-Kunz (approved):

David-Kunz:
@danjoa , might also be relevant for our [srv-dispatch](https://github.tools.sap/cap/cds/blob/8d937622cdb1fffd7b92cc48757107559128a150/lib/srv/srv-dispatch.js#L50).

PR #814, opened 2024-09-20
Refactor: import of compareJSON

fliellerjulian:
Simplified the import of compareJSON by removing the old path

forgot the tag `refactor:` in the commit message


patricebender (approved):
thanks!

johannes-vogel:
I think we should increase the minimum cds version in addition

PR #823, opened 2024-10-01
chore: cleanup legacy path for compareJSON (#814)" 

johannes-vogel:


db-service/package.json
```diff
@@ -28,7 +28,7 @@
     "generic-pool": "^3.9.0"
   },
   "peerDependencies": {
-    "@sap/cds": ">=7.9"
+    "@sap/cds": ">=8"
```
hakimio, line 31: With this change maybe there should also be a major version bump for `db-service`?

PR #831, opened 2024-10-09
feat: assoc-like calc elements after exists predicate

patricebender:
as documented in #830, assoc-like calculated elements are re-written by the compiler to include the calculation directives in their on-conditions. Hence, we can allow those assoc-like calculated elements everywhere, where unmanaged associations can be used:

1. `exists` predicate
2. scoped queries
3. nested projections

this PR implements 1. and 2.


db-service/test/cqn4sql/calculated-elements.test.js
```diff
@@ -701,19 +701,23 @@ describe('Unfolding calculated elements in select list', () => {
...
-    // calc elements can't end in an association, hence this does not work, yet.
     expect(() => cqn4sql(CQL`SELECT from booksCalc.Books { ID } where exists youngAuthorName`, model)).to.throw(
-      `Calculated elements cannot be used in “exists” predicates in: “exists youngAuthorName”`,
+      `Expecting path “youngAuthorName” following “EXISTS” predicate to end with association/composition, found “cds.String”`,
+    )
+  })
+  it('exists cannot leverage calculated elements which is an expression', () => {
+    // at the leaf of a where exists path, there must be an association
+    expect(() => cqn4sql(CQL`SELECT from booksCalc.Books { ID } where exists authorFullName`, model)).to.throw(
+      `Expecting path “authorFullName” following “EXISTS” predicate to end with association/composition, found “expression”`,
```
patricebender, line 713: Should we just omit the `, found “expression”` suffix, if the type cannot be determined?

stewsk (approved):

db-service/test/cqn4sql/calculated-elements.test.js
patricebender: I can do it, why not use the nice prosa :)

PR #837, opened 2024-10-10
fix: Improved behavioral consistency between the database services

BobdenOs:
This PR contains a few fixes that where required to make the new compliance tests to pass:
- Decimal precision behavior is aligned between the databases when defined
- `@cap-js/hana` now properly supports scalar `SELECT` queries in the columns

The goal of this PR is to remove as much dependencies as possible now that `cds@8` provides the `cds-test` command.

- Removing `jest` by switching to `cds-test`
- Removed `jest.config.js` files
- Removing `chai` by switching to `cds-test`
- Adjust all tests to fit with the `cds-test` `expect` pattern
- Removing `HANA` test action
- Move `HXE` Github action steps into a reuse file
- Added a lock for `HANA` database creation
- Added a fallback for `Postgres` during database creation
- Removed compliance import tests
- Added symbolic link to the compliance test folder


johannes-vogel (approved):

sebastien-savalle:
Can you please report this fix to a a version which support cds 7 ?

Actually, there are some inconsistencies:
- Latest version of the db-service still support cds 7.9
- cds-hana requires at least cds 8.2

So, when you fix a bug (like in this PR) which involves both dependencies, only project using cds > 8.2 can benefit from the fix.

johannes-vogel:
> Can you please report this fix to a a version which support cds 7 ?
> 
> Actually, there are some inconsistencies:
> 
> * Latest version of the db-service still support cds 7.9
> * cds-hana requires at least cds 8.2
> 
> So, when you fix a bug (like in this PR) which involves both dependencies, only project using cds > 8.2 can benefit from the fix.

I do not see any inconsistencies.
There are parts in `@cap-js/hana` which require a newer version even though the base package `@cap-js/db-service` does not need this. The base package is also used for sqlite and postgres where we do not yet require cds8.
In addition, `@cap-js/hana` is officially supported with cds8.

Are you running on cds7 + cap-js/hana? If yes, please consider updating the cds version.

sebastien-savalle:
No, we are running with cds7 and cap-js/hana, but we would like to first migrate to cap-js/hana (which is not easy) and then to cds8 to avoid too many regressions.

johannes-vogel:
> No, we are running with cds7 and `cap-js/hana`, but we would like to first migrate to cap-js/hana (which is not easy) and then to cds8 to avoid too many regressions.

We do recommend to first upgrade to cds8. You can adopt that and turn off the new odata protocol adapter. With that, there should not be much upgrade effort. Then you can adopt @cap-js/hana and later on the new protocol adapter or vice versa.

> Why did you publish a final version (1.0 & 1.1) of @cap-js/hana if the support of cds-7 was not official ?

There simply was no technical reason to restrict it to cds8 before.

PR #844, opened 2024-10-15
fix: Ensure that `Iterator` and `Readable` work correctly for `INSERT`

BobdenOs:
Currently a lot of the `cds-dbs` logic relies on non consuming `Arrays` which are required to be kept fully in memory. It should be allowed to `INSERT` any kind of `Iterator` for scenarios where it is possible to not load the whole dataset into the application.

Examples:

```js
const { Books } = cds.entities

// Generator
const gen = function*(amount) {
  for(let i = 0; i < amount; i++) {
    yield { ID: cds.utils.uuid() }
  }
}
await INSERT(gen(100)).into(Books)

// Generator (async)
const genAsync = async function*(amount) {
  for(let i = 0; i < amount; i++) {
    yield { ID: await fetchKey() }
  }
}
await INSERT(genAsync(100)).into(Books)

// Readable (raw)
express.post('/upload', async (req,res) => {
  await INSERT(req).into(Books)
  res.end('done')
})

// Readable (object)
await INSERT(Readble.from(genAsync(100))).into(Books)

```

### DIFF
--- a/test/compliance/INSERT.test.js
+++ b/test/compliance/INSERT.test.js
@@ -1,6 +1,12 @@
-require('../cds')
+const cds = require('../cds')
+
+const { Readable } = require('stream')
 
 describe('INSERT', () => {
+  const { data, expect } = cds.test(__dirname + '/resources')
+  data.autoIsolation(true)
+  data.autoReset()
+
   describe('into', () => {
     test.skip('missing', () => {
       throw new Error('not supported')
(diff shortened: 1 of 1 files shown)


danjoa (approved):
👍 love that (assuming it's correct, didn't check in detail :)

db-service/lib/cqn2sql.js
```diff
@@ -605,8 +607,9 @@ class CQN2SQLRenderer {
         } else {
           if (val != null && elements[this.columns[key]]?.type in this.BINARY_TYPES) {
             val = Buffer.from(val, 'base64').toString(binaryEncoding)
+          } else {
+            buffer += `${sepsub}${val == null ? 'null' : JSON.stringify(val)}`
           }
```
BobdenOs, line 612:
```suggestion
          buffer += `${sepsub}${val == null ? 'null' : JSON.stringify(val)}`
```

db-service/lib/cqn2sql.js
```diff
@@ -605,8 +607,7 @@ class CQN2SQLRenderer {
         } else {
           if (val != null && elements[this.columns[key]]?.type in this.BINARY_TYPES) {
             val = Buffer.from(val, 'base64').toString(binaryEncoding)
-          }
-          buffer += `${sepsub}${val === undefined ? 'null' : JSON.stringify(val)}`
+          buffer += `${sepsub}${val == null ? 'null' : JSON.stringify(val)}`
```
BobdenOs, line 610:
```suggestion
          }
          buffer += `${sepsub}${val == null ? 'null' : JSON.stringify(val)}`
```

Issue #773, opened 2024-08-05
@cap-js/hana - Problem with insert on tables with columns with default null - they are filled with current timestamp


DinoWinter:
### Insert a row in a table with columns with default null leads to current timestamps in these columns.

> 1. Create a table with some columns with a default null and deploy to HANA
```cds
namespace my.test;

using {
    managed
} from '@sap/cds/common';

entity Test : managed {
    key id          : UUID;
        description : String(255);
        anotherText : String(255) default null;
        loggedAt    : Timestamp default null;

};
```

> 2. Create a service and run local:
```cds
using {my.test as test} from '../db/test';

service TestService @(path : '/test')  {
    entity Test @(restrict : [{grant : [
        'READ',
        'WRITE'
    ]}]) as projection on test.Test;
}

```
> 3. Create a new row in the table with only a description set:
```
POST http://localhost:4004/test/Test
Content-Type: application/json

{
  "description": "a new description"
}
```

> The response is:
```
HTTP/1.1 201 Created
X-Powered-By: Express
X-Correlation-ID: 7f46ab8b-41b2-4e92-b392-7be6e985075b
OData-Version: 4.0
location: Test(87bb7c82-c059-4ebe-aa9a-62becc5cc04e)
Content-Type: application/json; charset=utf-8
Content-Length: 328
Date: Mon, 05 Aug 2024 14:33:24 GMT
Connection: close

{
  "@odata.context": "$metadata#Test/$entity",
  "anotherText": "2024-08-05T14:33:24.882Z",
  "createdAt": "2024-08-05T14:33:24.882Z",
  "createdBy": "privileged",
  "description": "a new description",
  "id": "87bb7c82-c059-4ebe-aa9a-62becc5cc04e",
  "loggedAt": "2024-08-05T14:33:24.882Z",
  "modifiedAt": "2024-08-05T14:33:24.882Z",
  "modifiedBy": "privileged"
}
```

> I found the problem in the @cap-js/hana
It seems to be in the file HANAService.js in the line 1040: const defaultValue...:
```js
        if (!isUpdate) {
          const d = element.default
          if (d && (d.val !== undefined || d.ref?.[0] === '$now')) {
            const defaultValue = d.val ?? (cds.context?.timestamp || new Date()).toISOString()
            managed = typeof defaultValue === 'string' ? this.string(defaultValue) : defaultValue
          }
        }
```
If the colunns has a default value of null then d.val is null and the date is set as the value for the column.

[...]

PR #845, opened 2024-10-15
fix: null as default value

johannes-vogel:
even though modeling is weird as the db default is anyway null, our implementation should go along with it.
 
closes https://github.com/cap-js/cds-dbs/issues/773

### DIFF
--- a/test/compliance/INSERT.test.js
+++ b/test/compliance/INSERT.test.js
@@ -1,6 +1,9 @@
-require('../cds')
+const cds = require('../cds')
 
 describe('INSERT', () => {
+  const { data, expect } = cds.test(__dirname + '/resources')
+  data.autoIsolation(true)
+
   describe('into', () => {
     test.skip('missing', () => {
       throw new Error('not supported')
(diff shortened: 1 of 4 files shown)


test/compliance/resources/db/basic/literals.cds
```diff
@@ -63,6 +63,14 @@ entity binaries {
     largebinary : LargeBinary;
 }
 
+entity defaults {
+    key ID: Integer;
+    string      : String default '';
```
BobdenOs, line 68: Could use the cds formatter

test/compliance/INSERT.test.js
```diff
@@ -32,4 +36,13 @@ describe('INSERT', () => {
       throw new Error('not supported')
     })
   })
+
+  describe('default values', () => {
+    test('default values are generated', async () => {
+      const { 'basic.literals.defaults': entity } = cds.db.entities
+      await cds.run(INSERT.into(entity).entries({ID: 1}))
+      const result = await cds.run(SELECT.from(entity, 1))
+      assert.deepEqual(result, { ID: 1, boolean: false, integer: 0, nulls: null, string: ''})
```
BobdenOs, line 45: I also like `assert`, but I do think it would be best to make it work with `expect` for overall consistency.

johannes-vogel: we do use `assert` already quite heavily, don't we? I changed from `expect` to `assert` as we want to go towards `assert` in the future?

BobdenOs (approved):

PR #858, opened 2024-10-22
fix: Throw error if rows in limit is missing for expand, same as in limit()

SamuelBrucksch:
OData requests like `$expand=xxx($skip=5)` currently fail, because rows is missing which causes a TypeError on HANA. We already have a missing rows check for limit(), so i added the same here.


BobdenOs (approved):

hana/lib/HANAService.js
```diff
@@ -421,6 +422,11 @@ class HANAService extends SQLService {
         }
 
         if (parent && (limit || one)) {
+          if (limit && limit.rows == null) {
```
johannes-vogel, line 425: isn't this also an issue for pg and sqlite?

BobdenOs: This is already handled in the `super.limit` function, but was missing from the HANA specific expand `limit` logic.

Issue #859, opened 2024-10-22
`@cap-js/hana` support for calling `Procedures` outside of the current schema


BobdenOs:
### Description of erroneous behavior 

Currently it is not possible to call procedures that are outside of the `CURRENT_SCHEMA` or `SYS` schema.

The problematic logic is defined [here](https://github.com/cap-js/cds-dbs/blob/main/hana/lib/HANAService.js#L1298) where the only option for the schema is `SYS` if any other schema is used it uses `CURRENT_SCHEMA`. This results in the procedure metadata not properly being retrieved and the output values won't be correctly returned.

### Detailed steps to reproduce

```js
// works:
await cds.run(`CALL PROC(OUT1 => ?, OUT2 => ?)`);
await cds.run(`CALL SYS.PROC(OUT1 => ?, OUT2 => ?)`);

// broken:
await cds.run(`CALL OTHER_SCHEMA.PROC(OUT1 => ?, OUT2 => ?)`);
```

### Details about your project

| Your Project Name | https://github.com/your/repo |
|:------------------|---------------------------------------|
| @cap-js/hana| 1.3.2 |

PR #862, opened 2024-10-23
feat: Improve procedure metadata lookup logic

BobdenOs:
When using `@cap-js/hana` it is possible to send `SQL` statements like:

```sql
CALL PROC()
CALL SYS.PROC()
CALL "PROC"()
CALL "SYS"."PROC"()
```

With this change it is also possible to call procedures outside of the current schema or sys schema like:

```sql
CALL SCHEMA.PROC()
CALL "SCHEMA"."PROC"()
CALL "Schema"."Proc"()
```

The logic always was case sensitive for the procedure name therefor the same is applied to the schema name. This only applies to the metadata lookup for improved output parameter handling. As the native HANA drivers don't behave the same for procedure.

fixes: https://github.com/cap-js/cds-dbs/issues/859

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -1306,7 +1306,7 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction})) AS NEW LE
     // name delimited with "" allows any character
     const match = sql
       .match(
-        /^\s*call \s*(("(?<schema_delimited>\w+)"\.)?("(?<delimited>.+)")|(?<schema_undelimited>\w+\.)?(?<undelimited>\w+))\s*\(/i
+        /^\s*call \s*(("(?<schema_delimited>\w+)"\.)?("(?<delimited>.+)")|((?<schema_undelimited>\w+)\.)?(?<undelimited>\w+))\s*\(/i
       )
     return (
       match && {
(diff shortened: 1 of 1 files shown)


hana/lib/HANAService.js
```diff
@@ -1295,8 +1295,10 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction})) AS NEW LE
   }
 
   async _getProcedureMetadata(name, schema) {
-    const query = `SELECT PARAMETER_NAME FROM SYS.PROCEDURE_PARAMETERS WHERE SCHEMA_NAME = ${schema?.toUpperCase?.() === 'SYS' ? `'SYS'` : 'CURRENT_SCHEMA'
-      } AND PROCEDURE_NAME = '${name}' AND PARAMETER_TYPE IN ('OUT', 'INOUT') ORDER BY POSITION`
+    const sqlString = this.class.CQN2SQL.prototype.string
+    name = typeof name === 'string' ? sqlString(name) : `'${name}'`
+    schema = typeof schema === 'string' ? sqlString(schema) : 'CURRENT_SCHEMA'
```
johannes-vogel, line 1300: What does `sqlString` do?
Might be related to the failing test

BobdenOs: `sqlString` is coming from `cqn2sql` which implements how to inline strings into a SQL statement ([code](https://github.com/cap-js/cds-dbs/blob/641c3b97d0571b8b8b7f44e038ffc34ac17c49e5/db-service/lib/cqn2sql.js#L1013-L1015)).

BobdenOs: The test was failing, because it was testing functionalities not implemented and the original regexp would return `<schema>.` as the schema name.

johannes-vogel (approved):

PR #865, opened 2024-10-24
feat: support current_utctimestamp

johannes-vogel:
works locally with SAP HANA Cloud but does not work with HXE


db-service/lib/cqn2sql.js
```diff
@@ -990,7 +990,7 @@ class CQN2SQLRenderer {
     } else {
       cds.error`Invalid arguments provided for function '${func}' (${args})`
     }
-    const fn = this.class.Functions[func]?.apply(this.class.Functions, args) || `${func}(${args})`
+    const fn = this.class.Functions[func]?.apply(this.class.Functions, args) || `${func}${args ? `(${args})`: ''}`
```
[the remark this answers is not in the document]
johannes-vogel: revert

hana/test/hana-functions.test.js
```diff
@@ -0,0 +1,50 @@
+const cds = require('../../test/cds')
```
BobdenOs, line 1: Should we add this function definition into the other databases? 
Then the test can be move here:
https://github.com/cap-js/cds-dbs/blob/45bcdcf30ed103c1db8283d30ce5d84a7748a1d5/test/compliance/functions.test.js#L339-L348

johannes-vogel: I initially started there but the other databases don't support the argument

johannes-vogel: https://github.com/cap-js/cds-dbs/pull/909 but I don't like the different behavior on the dbs

BobdenOs (approved):

hana/test/hana-functions.test.js
BobdenOs: It is a good finding that the databases have different response formats. The purpose of the cal-functions is to define them in a way that they response in the correct format. Most likely adding a `cast` or `element` to the original func can apply the correct output converter for each database.

PR #866, opened 2024-10-24
feat: deep `UPDATE` without reading

BobdenOs:
tests are currently failing only on `HXE` everything passed on `HCE`


db-service/lib/SQLService.js
```diff
@@ -214,7 +214,7 @@ class SQLService extends DatabaseService {
     // REVISIT: It's not yet 100 % clear under which circumstances we can rely on db constraints
     return (super.onDELETE = /* cds.env.features.assert_integrity === 'db' ? this.onSIMPLE : */ deep_delete)
     async function deep_delete(/** @type {Request} */ req) {
-      let { compositions } = req.target
+      let { compositions } = (req.target ??= req.query.target)
```
David-Kunz, line 217: when is this not set? https://github.tools.sap/cap/cds/blob/402a114f8968c2bc25c5f1e2de2208f4cac0c66d/lib/srv/srv-dispatch.js#L29

db-service/lib/cqn2sql.js
```diff
@@ -750,7 +751,7 @@ class CQN2SQLRenderer {
     const managed = this._managed.slice(0, columns.length)
 
     const extractkeys = managed
-      .filter(c => keys.includes(c.name))
+      // .filter(c => keys.includes(c.name))
```
David-Kunz, line 754: ?

db-service/lib/cqn2sql.js
```diff
@@ -1005,6 +1007,20 @@ class CQN2SQLRenderer {
     return `(${list.map(e => this.expr(e))})`
   }
 
+  json(arg) {
+    const { props, elements, json } = arg
+    const { _convertInput } = this.class
+    let val = typeof json === 'string' ? json : (arg.json = JSON.stringify(json))
+    if (val[val.length - 1] === ',') val = arg.json = val.slice(0, -1) + ']'
+    if (val[val.length - 1] === '[') val = arg.json = val + ']'
```
David-Kunz, line 1015:
```suggestion
    if (val.at(-1) === ',') val = arg.json = val.slice(0, -1) + ']'
    if (val.at(-1) === '[') val = arg.json = val + ']'
```

db-service/lib/deep-genres.sql
```diff
@@ -0,0 +1,126 @@
+-- DEBUG => this.dbc._native.prepare(cds.utils.fs.readFileSync(__dirname + '/deep-genres.sql','utf-8')).exec([JSON.stringify(query.UPDATE.data)])
+DO (IN JSON NCLOB => ?) BEGIN
+
+  -- Extract genres with a depth of 3 (like: '$.children[*].children[*]')
```
David-Kunz, line 4: depth can be determined based on input json, right? we must just `union all` all provided rows

db-service/lib/deep-genres.sql
```diff
@@ -0,0 +1,126 @@
...
+        ID INT PATH '$.ID',
+        "$.ID" NVARCHAR(2147483647) FORMAT JSON PATH '$.ID',
+        parent_ID INT PATH '$.parent_ID',
+        "$.PARENT_ID" NVARCHAR(2147483647) FORMAT JSON PATH '$.parent_ID',
+        "$.CHILDREN" NVARCHAR(2147483647) FORMAT JSON PATH '$.children'
+      )
+      ERROR ON ERROR
+    ) AS NEW;
+
+  -- DELETE all children of parents that are no longer in the dataset
```
David-Kunz, line 86: where are rows deleted which are not children of parents, but which are not part of the dataset?

E.g.

JSON input:

| ID  | parent_ID |
|-----|-----------|
| 1   | null      |
| 11  | 1         |
| 12  | 1         |
| 2   | null      |
| 21  | 2         |
| 22  | 2         |


Existing data:

| ID  | parent_ID | will be deleted? |
|-----|-----------|----------------|
| 13  | 1         |             yes |
| 3   | null      |             no |

David-Kunz: Note: They might be children of _other_ entities and would not have to be deleted.

David-Kunz: Thinking about it again, we can exclude non-children entities since compositions, by definition, are children and one needs to consider only children and children's children of the root instance.

We should also look at the case of a composition of one where the foreign key is part of the parent. Here, one cannot identify the previous child purely based on the payload and one needs to take into account the foreign keys of existing data.

PR #867, opened 2024-10-24
fix: encoding of large strings

johannes-vogel:


johannes-vogel:
closing in favor of #840

PR #868, opened 2024-10-24
fix: enable `cesu8` by default for `hdb` driver and encode entries streams 

BobdenOs:
fixes: https://github.com/cap-js/cds-dbs/blob/da629d98192cf1b196cceb172cb01f4a39acd887/test/compliance/resources/db/basic/literals/basic.literals.string.js#L41

The tests run with `hana-client` inside the pipelines. Tested locally works as expected.


johannes-vogel:
did the test also fail for plain string?

PR #882, opened 2024-11-05
fix: wrap values in array if it is object, so it can be spreaded

SamuelBrucksch:
A query with no properties remaining caused a TypeError earlier, as we try to destructure an object:
```
Uncaught TypeError: Found non-callable @@iterator
```
So if it is wrapped as array, the TypeError does not exist anymore.


db-service/lib/cqn2sql.js
```diff
@@ -90,10 +82,16 @@ class CQN2SQLRenderer {
     if (vars?.length && !this.values?.length) this.values = vars
     if (vars && Object.keys(vars).length && !this.values?.length) this.values = vars
     const sanitize_values = process.env.NODE_ENV === 'production' && cds.env.log.sanitize_values !== false
-    DEBUG?.(
-      this.sql,
-      ...(sanitize_values && (this.entries || this.values?.length > 0) ? ['***'] : this.entries || this.values || []),
-    )
+
+    if (DEBUG_SQL_JSON) {
```
SamuelBrucksch, line 86:
```suggestion
    if (DEBUG) {
```

db-service/lib/cqn2sql.js
```diff
@@ -90,10 +81,17 @@ class CQN2SQLRenderer {
     if (vars?.length && !this.values?.length) this.values = vars
     if (vars && Object.keys(vars).length && !this.values?.length) this.values = vars
     const sanitize_values = process.env.NODE_ENV === 'production' && cds.env.log.sanitize_values !== false
-    DEBUG?.(
-      this.sql,
-      ...(sanitize_values && (this.entries || this.values?.length > 0) ? ['***'] : this.entries || this.values || []),
-    )
+
+    const DEBUG_ENABLED = cds.log('sql')._debug || cds.log('sqlite')._debug
```
BobdenOs, line 85: calling `cds.log` this often will make it much more heavy then just always calling `DEBUG`. Just make these two `cds.log`s at the start of the file.

db-service/lib/cqn2sql.js
```diff
@@ -6,16 +6,9 @@ const _strict_booleans = _simple_queries < 2
...
-  return cds.debug('sql|sqlite')
-  //if (DEBUG) {
-  //  return DEBUG
-  // (sql, ...more) => DEBUG (sql.replace(/(?:SELECT[\n\r\s]+(json_group_array\()?[\n\r\s]*json_insert\((\n|\r|.)*?\)[\n\r\s]*\)?[\n\r\s]+as[\n\r\s]+_json_[\n\r\s]+FROM[\n\r\s]*\(|\)[\n\r\s]*(\)[\n\r\s]+AS )|\)$)/gim,(a,b,c,d) => d || ''), ...more)
-  // FIXME: looses closing ) on INSERT queries
-  //}
-})()
+const DEBUG = cds.debug('sql|sqlite')
+const LOG_SQL = cds.log('sql')
+const LOG_SQLITE =  cds.log('sqlite')
```
SamuelBrucksch, line 11: Why do we actually explicitly check for sqlite here? Shouldn't sql only be good enough as it is DB agnostic?

BobdenOs (approved):

PR #883, opened 2024-11-05
chore: Using local sflight clone instead of git dependency

danjoa:
Simplifies our setup, and speeds up pipelines


danjoa:
Agreed, but we anyways have to find better ways to deal with that ... i.e. we need some umbrella-setup like cap/dev available externally in public GitHub. That also is the preliminary conclusion from respective considerations with Daniel around Incidents Mgmt and the Calesi repos (which all live in public Github). 

This PR is a near-term measure to fix problems we have with always having a duplicate install of SFlight in cap/dev today, as well as the problem that Steffen encountered recently when he had to revert a change he thought to do locally in SFlight which broke tests in cds-dbs. → I'd propose to merge that for now and follow up on the bigger thing near term.

BobdenOs (approved):

PR #884, opened 2024-11-05
fix: Overly expensive test

danjoa:
This test eats up 16 sec, blocking a worker thread for that time. 
These changes reduce the runtime to 2.5 sec. 
Something in that direction would be great.


danjoa:
Ready 2b merged → @BobdenOs pls check & approve if ok

BobdenOs (approved):

PR #885, opened 2024-11-06
test: switch to `sflight` and `bookshop` sub modules

BobdenOs:
Tries to address the issues pointed out by https://github.com/cap-js/cds-dbs/pull/884 and https://github.com/cap-js/cds-dbs/pull/883


danjoa:
As commented in #883 the approach is the right direction, but we should do so in some kind of umbrella setup like in cap/dev.

PR #886, opened 2024-11-08
fix: boolean for sql_simple_queries 

larslutz96:
For sql_simple_queries=2 elements of Boolean type are returned as strings due to the output converter. To correct this, we disable the boolean output converter for this feature flag.

### DIFF
--- a/sqlite/lib/SQLiteService.js
+++ b/sqlite/lib/SQLiteService.js
@@ -208,10 +208,10 @@ class SQLiteService extends SQLService {
       struct: expr => `${expr}->'$'`,
       array: expr => `${expr}->'$'`,
       // SQLite has no booleans so we need to convert 0 and 1
-      boolean: expr =>
+      boolean:
         cds.env.features.sql_simple_queries === 2
           ? undefined
-          : `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
+          : expr => `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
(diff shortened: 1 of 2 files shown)


sqlite/lib/SQLiteService.js
```diff
@@ -208,7 +208,10 @@ class SQLiteService extends SQLService {
       struct: expr => `${expr}->'$'`,
       array: expr => `${expr}->'$'`,
       // SQLite has no booleans so we need to convert 0 and 1
-      boolean: expr => `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
+      boolean: expr =>
+        cds.env.features.sql_simple_queries === 2
+          ? undefined
+          : `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
```
BobdenOs, line 214:
```suggestion
      boolean: cds.env.features.sql_simple_queries === 2 ? undefined : expr => `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
```

patricebender:
just out of curiosity: what does `sql_simple_queries=2` do? why is the value of this feature flag `2`?

johannes-vogel (approved):

johannes-vogel:
> just out of curiosity: what does `sql_simple_queries=2` do? why is the value of this feature flag `2`?

sql_simple_queries>0 allows to use SQL statements without JSON functions if possible.
sql_simple_queries=1 uses JSON functions for expands and boolean conversion
sql_simple_queries=2 uses JSON functions for expands

PR #887, opened 2024-11-08
perf: Add foreign key indexes to tables

BobdenOs:
With the latest release of `better-sqlite3@11.5` the `SQLite3` version was bumped to `3.47.0` ([changes](https://sqlite.org/releaselog/3_47_0.html)).

> No attempt is made to create [automatic indexes](https://sqlite.org/optoverview.html#autoindex) on a column that is known to be non-selective because of its use in other indexes that have been analyzed.

this seems to be the most important change for CAP applications as we create many `join` conditions using `associations` and `compositions` which `SQLite3` tries to process to the best of its ability by creating temporary automatic indexes, but with this optimization the likeliness of these indexes being created is greatly reduced. Causing some of the more complex join queries to take `5000x` longer then when using `better-sqlite@11.4`. This PR tries to mitigate the effect of the reduced number of automatically generated indexes by using the model definition to provide explicit indexes for these modeled join condition sources / targets.

An example of the impact that providing explicit indexes has on complex SFlight requests:

```
analytics/Bookings?$expand=to_Travel($expand=CurrencyCode,TravelStatus,to_Agency,to_Customer,to_Booking),to_Carrier,to_Flight&$top=30
```

|   | sqlite | postgres | hana |
| --- | --- | --- | --- |
| before | ~50 req/s | ~50 req/s | ~140 req/s |
| after | ~130 req/s | ~110 req/s | ~140 req/s |

So while it doesn't have an impact on the HANA performance it provides additional optimization information for `SQLite` and `Postgres`. Closing the gap that `SQLite` and `Postgres` always have had when being compared with `HANA` for complex join heavy queries.

### DIFF
--- a/db-service/lib/SQLService.js
+++ b/db-service/lib/SQLService.js
@@ -294,8 +294,39 @@ class SQLService extends DatabaseService {
       DEBUG?.(query, data)
       const ps = await this.prepare(query)
       const exec = this.hasResults(query) ? d => ps.all(d) : d => ps.run(d)
-      if (Array.isArray(data) && Array.isArray(data[0])) return await Promise.all(data.map(exec))
-      else return exec(data)
+      let ret
+      if (Array.isArray(data) && Array.isArray(data[0])) ret = await Promise.all(data.map(exec))
+      else ret = await exec(data)
+
+      // Leverage modeling structure to create table indexes
+      if (this.model && /^\s*CREATE TABLE/i.test(query)) {
+        // Extracts the name from the incoming SQL statment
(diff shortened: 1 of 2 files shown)


patricebender (approved):

danjoa (changes requested):
While I certainly appreciate that... shouldn't that rather go to the compiler 2sql backend that creates the DDL?
Would also allow us to avoid the expensive reengineering clumsinesses for finding the original definitions from `_`-slugified names, I think?

PR #889, opened 2024-11-12
feat: support for cds.Map

SamuelBrucksch:
As Map type also has `elements`, we get into that `if` and then it is ignored. So I added a check, to pass through map type, so it is handled like a normal ref.


patricebender:
I think for this change to work we need at least compiler v5.4.0

johannes-vogel:
@BobdenOs I've tried to fix the remaining HANA failure but I think we need to have a look together.

BobdenOs (approved):

johannes-vogel:
double check conversion of map type with simple queries

PR #890, opened 2024-11-13
chore(deps): update package-lock.json

johannes-vogel:
Updating the package-lock revealed that `db-service/test/assocs/unmanaged-assocs.test.js` uses an own mechanism for testing. I've worked around that now but is that test of value @patricebender?

### DIFF
--- a/db-service/test/assocs/unmanaged-assocs.test.js
+++ b/db-service/test/assocs/unmanaged-assocs.test.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 const cds = require('@sap/cds')
+cds.test.in(__dirname)
 require('../../index') // to extend cds.ql query objects with .forSQL() and alike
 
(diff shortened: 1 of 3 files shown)


db-service/test/assocs/unmanaged-assocs.test.js
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 const cds = require('@sap/cds')
+cds.test.in(__dirname)
```
BobdenOs, line 3: I have seen this issue before. It happens because `cds-test` now lazy loads `cds.test` and therefor `require` is not defined until `cds.test` is accessed once. One way to solve this issue is by using the extended `cds.test` definition in `test/cds.js`.

like:
```suggestion
const cds = require('../../../test/cds')
```

BobdenOs (approved):

PR #893, opened 2024-11-13
fix: enable nulls first

David-Kunz:
OData demands `NULLS FIRST` for `ASC` and `NULLS LAST` for `DESC`. SQLite, HANA behave like that, for Postgres the default is the opposite which is now changed.

### DIFF
--- a/test/compliance/SELECT.test.js
+++ b/test/compliance/SELECT.test.js
@@ -154,7 +154,7 @@ describe('SELECT', () => {
       assert.strictEqual(res.length, 1, 'Ensure that all rows are coming back')
       assert.strictEqual(res[0].count_star, 3, 'Ensure that the function is applied and aliased')
       assert.strictEqual(res[0].count_one, 3, 'Ensure that the function is applied and aliased')
-      assert.strictEqual(res[0].count_string, 3, 'Ensure that the function is applied and aliased')
+      assert.strictEqual(res[0].count_string, 2, 'Ensure that the function is applied and aliased')
       assert.strictEqual(res[0].count_char, 0, 'Ensure that the function is applied and aliased')
     })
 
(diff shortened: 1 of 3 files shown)


patricebender (commented):
could you add a test for this in our compliance test suite (then it would be tested for all dbs) or at whatever place you may think is appropriate? Thanks :)

David-Kunz:
> could you add a test for this in our compliance test suite (then it would be tested for all dbs) or at whatever place you may think is appropriate? Thanks :)

of course, I just forgot to put this PR as a draft PR

test/compliance/SELECT.test.js
```diff
@@ -782,7 +789,7 @@ describe('SELECT', () => {
       cqn.SELECT.one = true
       const res = await cds.run(cqn)
       assert.strictEqual(!Array.isArray(res) && typeof res, 'object', 'Ensure that the result is an object')
-      assert.strictEqual(res.string, 'no', 'Ensure that the first row is coming back')
+      assert.strictEqual(res.string, null, 'Ensure that the first row is coming back') // NULL value must come first (as per OData)
```
David-Kunz, line 792:
```suggestion
      assert.strictEqual(res.string, null, 'Ensure that the first row is coming back and null values come first')
```

test/compliance/SELECT.test.js
```diff
@@ -531,6 +531,9 @@ describe('SELECT', () => {
   })
 
   describe('orderby', () => {
+
+    const _localeSort = (a, b) => a === null ? -1 : b === null ? 1 : String.prototype.localeCompare.call(a, b)
```
BobdenOs, line 535: I would expect a check that both `a` and `b` are `null`, but this is a test and only has single `null` row. So ...

BobdenOs (approved):

test/compliance/SELECT.test.js
David-Kunz: yes, not needed, but in case someone adds another test I've adapted it 👍

PR #901, opened 2024-11-18
docs(hana): add guide on how to run tests

patricebender:
I keep forgetting how to run the tests in a remote SAP HANA system of my choice,  so I wanted to write it down for next time and for others.

### DIFF
--- a/hana/README.md
+++ b/hana/README.md
@@ -27,7 +27,7 @@ To run the tests on a a remote SAP HANA system, you need to adapt either your en
 Then execute the following command:
 
 ```sh
-npm run test:remote
+npm run test
 ```
 
 ## Support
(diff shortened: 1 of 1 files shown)


hana/README.md
```diff
@@ -12,6 +12,24 @@ npm add @cap-js/hana
...
+```
+
+*NOTE: The docker image may not be available for ARM-based systems.*
+
+To run the tests on a a remote SAP HANA system, you need to adapt either your environment variables or the `hana` configuration in the [`server.js`](https://github.com/cap-js/cds-dbs/blob/c61a04aa4394511100f97cfebd362a2298221d96/hana/test/service.js) file.
+
+Then execute the following command:
+
+```sh
+npm run test:remote
```
patricebender, line 30:
```suggestion
npm run test
```

PR #903, opened 2024-11-20
fix: foreignKeys aspect on association may not be set

patricebender:


johannes-vogel (approved):
even though I do not understand why you check in `assoc.elements` for foreign keys

patricebender:
> even though I do not understand why you check in `assoc.elements` for foreign keys

I think it is not necessary anymore. From the js doc I can tell that this was for Universal CSN, where associations actually have a `.elements`:

```js
/**
 * Returns true if e is a foreign key of assoc.
 * this function is also compatible with unfolded csn (UCSN),
 * where association do not have foreign keys anymore.
 */
```

PR #906, opened 2024-11-21
fix: Default `false` for `sslUseDefaultTrustStore`

swaldmann:
Possible workaround for https://github.com/SAP/node-hdb/issues/240

### DIFF
--- a/hana/package.json
+++ b/hana/package.json
@@ -50,7 +50,14 @@
           "impl": "@cap-js/hana"
         }
       },
-      "db": "sql"
+      "db": {
+        "kind": "sql",
+        "[production]": {
+          "credentials": {
+            "sslUseDefaultTrustStore": false
+          }
+        }
+      }
     }
   },
   "license": "SEE LICENSE"
--- a/hana/package.json
+++ b/hana/package.json
@@ -43,21 +43,15 @@
       "kinds": {
         "sql": {
           "[production]": {
-            "kind": "hana"
+            "kind": "hana",
+            "sslUseDefaultTrustStore": false
           }
         },
         "hana": {
           "impl": "@cap-js/hana"
         }
       },
-      "db": {
-        "kind": "sql",
-        "[production]": {
-          "credentials": {
-            "sslUseDefaultTrustStore": false
-          }
-        }
-      }
+      "db": "sql"
     }
   },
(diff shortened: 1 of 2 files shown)


hana/package.json
```diff
@@ -43,7 +43,8 @@
       "kinds": {
         "sql": {
           "[production]": {
-            "kind": "hana"
+            "kind": "hana",
+            "sslUseDefaultTrustStore": false
```
patricebender, line 47: why do we need that?:)

BobdenOs: I am pretty sure that this property would be required to be part of the `credentials` object. As that is what being send to the database driver. Also this seems to only be a problem with `hdb` it shouldn't be a problem with `@sap/hana-client`. Which would mean that it should probably go into:

https://github.com/cap-js/cds-dbs/blob/732a2f385074f50b87ff9715b8bdf48d28a36309/hana/lib/drivers/hdb.js#L23-L41

As the change was made in `hdb` to resolve a long running issue. I am hesitant to have our implementation overwrite the default. The main problem with this change is when the `HANAService` has a large connection pool or a lot of tenants. As out default `max` pool size is `10`. It would require ~1000 tenants to reach the reported state in the issue.

Additionally it is possible to use the previous `hdb` version which does not have this behavior to achieve the same state as this PR is trying to achieve. Which only becomes a problem when root issue shows up as it would require the `hdb` upgrade, but then having it disabled by `@cap-js/hana` would not allow CAP applications to connect to their HANA system at all anymore.

swaldmann: Agreed that this is the better place. I moved it to `hdb.js` now.

johannes-vogel:
is that already fixed by [hdb 0.19.11](https://github.com/SAP/node-hdb/releases/tag/v0.19.11)?

patricebender:
> is that already fixed by [hdb 0.19.11](https://github.com/SAP/node-hdb/releases/tag/v0.19.11)?

https://github.com/cap-js/cds-dbs/pull/918

oklemenz2:
No, with hdb 0.19.11 does not fix the issue, just change certificates that are mixed in...

I069405:
is that already fixed by [hdb 0.19.12](https://github.com/SAP/node-hdb/releases/tag/v0.19.12)?

PR #910, opened 2024-11-22
fix: SAP HANA functions should also be callable in upper case on SAP HANA

johannes-vogel:
According to https://cap.cloud.sap/docs/guides/databases#functions-mappings-for-runtime-queries

> For the SAP HANA functions, both usages are allowed: all-lowercase as given above, as well as all-uppercase.

### DIFF
--- a/test/cds.js
+++ b/test/cds.js
@@ -72,7 +72,7 @@ cds.test = Object.setPrototypeOf(function () {
         hash.update(isolateName)
         ret.data.isolation = isolate = {
           // Create one database for each overall test execution
-          database: process.env.TRAVIS_JOB_ID || process.env.GITHUB_RUN_ID || require('os').userInfo().username.repeat(6) || 'test_db',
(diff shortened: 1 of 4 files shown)


test/cds.js
```diff
@@ -72,7 +72,7 @@ cds.test = Object.setPrototypeOf(function () {
         hash.update(isolateName)
         ret.data.isolation = isolate = {
           // Create one database for each overall test execution
-          database: process.env.TRAVIS_JOB_ID || process.env.GITHUB_RUN_ID || require('os').userInfo().username || 'test_db',
+          database: process.env.TRAVIS_JOB_ID || process.env.GITHUB_RUN_ID || require('os').userInfo().username.repeat(6) || 'test_db',
```
johannes-vogel, line 75:
```suggestion
          database: process.env.TRAVIS_JOB_ID || process.env.GITHUB_RUN_ID || require('os').userInfo().username || 'test_db',
```

BobdenOs (approved):

PR #921, opened 2024-11-28
feat: pass options to bettersqlite3 constructor

### LINKED ISSUE #920: [sqlite] Pass options to constructor of bettersqlite3
The constructor of bettersqlite3 is currently called without any options https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#new-databasepath-options in `SQLiteService.js`:

```js
const dbc = new sqlite(database);
```

Users should be able to specify options in `cds.requires.db.sqlite` which are then passed to the constructor of bettersqlite3:

```js
const dbc = new sqlite(database, this.options.sqlite);
```

aschmidt93:
Resolves #920

### DIFF
--- a/sqlite/lib/SQLiteService.js
+++ b/sqlite/lib/SQLiteService.js
@@ -22,7 +22,7 @@ class SQLiteService extends SQLService {
       options: { max: 1, ...this.options.pool },
       create: tenant => {
         const database = this.url4(tenant)
-        const dbc = new sqlite(database)
+        const dbc = new sqlite(database, this.options.sqlite)
 
         const deterministic = { deterministic: true }
         dbc.function('session_context', key => dbc[$session][key])


sqlite/lib/SQLiteService.js
```diff
@@ -22,7 +22,7 @@ class SQLiteService extends SQLService {
       options: { max: 1, ...this.options.pool },
       create: tenant => {
         const database = this.url4(tenant)
-        const dbc = new sqlite(database)
+        const dbc = new sqlite(database, this.options.sqlite)
```
patricebender, line 25: @BobdenOs looks like a useful addition, what do you think?

patricebender:
Hi @aschmidt93

thanks for contributing. We agree that it makes sense to make it possible to pass options to sqlite. We also have been requested to provide a way to supply options for other packages which are used by us, e.g. [`hdb`](https://www.npmjs.com/package/hdb).

That being said, we already figuring out a generic approach to provide options for our dependencies. Hence I am closing this PR, as we prefer a generic - CAP wide - way of doing this.

PR #923, opened 2024-11-28
fix: wildcard expand with aggregation

patricebender:
for the case where we `group by` expanded columns, we must ignore the `expand` from the columns


db-service/test/cqn4sql/expand.test.js
```diff
@@ -1050,6 +1050,35 @@ describe('Expands with aggregations are special', () => {
     const res = cqn4sql(q, model)
     expect(JSON.parse(JSON.stringify(res))).to.deep.equal(qx)
   })
+  it.skip('simple aggregation expand ref wrapped in func', () => {
```
johannes-vogel, line 1053: is this also representing an issue?

johannes-vogel (approved):

PR #925, opened 2024-11-29
fix: expressions with not on HANA

vkozyura:


johannes-vogel:
@vkozyura Do you think you'll be able to fix the case expressions until tomorrow?
If not, I propose to split this in two PRs in order to be able to deliver the already implemented fix.

BobdenOs (approved):

vkozyura:
> @vkozyura Do you think you'll be able to fix the case expressions until tomorrow? If not, I propose to split this in two PRs in order to be able to deliver the already implemented fix.

@johannes-vogel: We merged the fix. I will create follow-up for the remaining CASE issues

PR #934, opened 2024-12-04
perf: unify `sql_simple_queries` for HANA queries

BobdenOs:
For applications using `sql_simple_queries=2` and `@sap/hana-client` the query results will properly reflect `true` and `false` for `boolean` fields, but when using `hdb` as driver the `boolean` datatype is not returned by the HANA protocol. While `hdb` has the data type defined ([code](https://github.com/SAP/node-hdb/blob/49c69673279655aa6534a9a5f38248ebdcfe36d3/lib/protocol/common/TypeCode.js#L45https://github.com/SAP/node-hdb/blob/49c69673279655aa6534a9a5f38248ebdcfe36d3/lib/protocol/common/TypeCode.js#L45)) for some reason HANA does not recognize `hdb` as a client capable of handling `boolean` column types. Therefor the data type returned is `tinyint` this change adjusts the data converter to output `true` and `false` when using `sql_simple_queries=1`. Allowing the database service to generate the same SQL statements for both `1` and `2`. Saving an additional loop on the application to convert the numeric results to booleans.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -1349,7 +1349,6 @@ const logicOperators = {
   'THEN': 1,
   'AND': 1,
   'OR': 1,
-  'NOT': 1,
 }
 const compareOperators = {
   '=': 1,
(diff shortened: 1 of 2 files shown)


oklemenz2:
PR https://github.com/cap-js/cds-dbs/pull/929 introduces a second occurrence to be removed ;-)

johannes-vogel (approved):

PR #937, opened 2024-12-06
fix: quoted mode

johannes-vogel:


db-service/lib/cqn2sql.js
```diff
@@ -21,10 +21,13 @@ class CQN2SQLRenderer {
     this.class = new.target // for IntelliSense
     this.class._init() // is a noop for subsequent calls
     this.model = srv?.model
-
     // Overwrite smart quoting
     if (cds.env.sql.names === 'quoted') {
-      this.class.prototype.name = (name) => name.id || name
+      this.class.prototype.name = (name) => {
+        let e = name.id || name
+        const pn = this.model?.definitions[e]?.['@cds.persistence.name']
```
BobdenOs, line 28: I don't like the model lookup. It might be a bigger refactor, but passing the `target` from the outside would make this much more reliable.

johannes-vogel: I tried to solve "quoted mode" at a single point.

Why do you think that it's more reliable once passed from the outside?

BobdenOs (approved):

PR #940, opened 2024-12-09
feat: Add support for `odbc` driver

BobdenOs:
While setting up the `ABAP` database service (https://github.com/cap-js/cds-dbs/pull/936) we where using the `odbc` module to create the database driver. There is also an `HANA` odbc driver version available. So this PR is taking the same `odbc` approach as the `ABAP` service to create a database connection.


hana/lib/drivers/odbc.js
```diff
@@ -0,0 +1,162 @@
+const path = require('path')
```
patricebender, line 1: where can I find the source of this code?:)

PR #942, opened 2024-12-09
fix: handle undefined DEBUG

### LINKED ISSUE #941: [postgres] Error 'DEBUG is not a function' when setting env variable 'debug' to 'sqlite'

When the env variable `debug` is set to value `sqlite`, an error is thrown in `cqn2ql.js` where `DEBUG is not a function` as it is undefined.

In `SQLService.js` the optional chaining operator is already used with `DEBUG.`

aschmidt93:
Closes #941


patricebender (approved):
🥇

BobdenOs (approved):

PR #944, opened 2024-12-10
chore: fix db.run('...',1,2,3,4)

johannes-vogel:
otherwise, the transformed `data` is not considered.

introduced in #943

### DIFF
--- a/db-service/lib/common/DatabaseService.js
+++ b/db-service/lib/common/DatabaseService.js
@@ -154,10 +154,7 @@ class DatabaseService extends cds.Service {
    */
   run(query, data, ...etc) {
     // Allow db.run('...',1,2,3,4)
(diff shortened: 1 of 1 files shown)


db-service/lib/common/DatabaseService.js
```diff
@@ -154,7 +154,10 @@ class DatabaseService extends cds.Service {
    */
   run(query, data, ...etc) {
     // Allow db.run('...',1,2,3,4)
-    if (data !== undefined && typeof query === 'string' && typeof data !== 'object') data = [data, ...etc]
+    if (data !== undefined && typeof query === 'string' && typeof data !== 'object') {
+      data = [data, ...etc]
+      return super.run(query, data)
+    }
```
BobdenOs, line 160:
```suggestion
    if (data !== undefined && typeof query === 'string' && typeof data !== 'object') arguments[1] = [data, ...etc]
```

PR #952, opened 2024-12-17
feat: automated test isolation

BobdenOs:
## Concept

The basic concept of this change is to provide test isolation for all the `cds-dbs` databases automatically. Up till now only the `cds-dbs` repository has been able to run tests in parallel while using persistent database services. Therefor many stakeholders have setup their own custom credential handling. This change is aiming to make all custom test isolation solutions obsolete.

To make this solution viable for all CAP applications the following criteria have to be considered:
1. deploy only when **_required_**
1. allow for parallel test executions
1. allow for parallel test runs

### test run isolation

The current implementation comes with support for `Travis` and `Github Actions` by using the automatic environment variables to create test run isolations given the generic name `database`. Additionally for local development the default operating system user name is used as `database` isolation. This allows for using shared database instances when running local instances of the database are not available. If none of these stable isolation names are found a default fallback name is used. This comes with the benefit that deployment reductions are functional, but might come with side effects when multiple users are relying on the fallback `database` isolation.

### Minimal deployment

The test isolation logic relies upon the project sources to determine what needs to be deployed and what deployments can be re used. In most applications the source is defined as `*`. Therefor the test isolation combines this with the `cds.root` to determine unique source definition. 

For all tests a shared read only isolation is created which all tests will initially use. Until a write operation is detected at this point the operations are paused and the isolation is switched to a test specific isolation. When switching to a test unique isolation the database is checked for any available previously deployed isolations. If no previously deployed isolations are available a new isolation is created and deployed. After the test has finished the isolation is reset to its original deployment state and marked as available for any follow up write tests.

By assuming all tests are read only tests at the start it allows the test to avoid triggering deployment as much as possible. Even when all tests running in parallel are write tests. As long as one of them finishes before another test starts writing they are able to share the same deployment. Where if the tests are assumed to be writing tests they will all always trigger their own deployment at the start. This approach results in a maximum number of deployments equal to `number of parallel tests


db-service/lib/common/DatabaseService.js
```diff
@@ -3,15 +3,183 @@ const ConnectionPool = require('./generic-pool')
 const infer = require('../infer')
 const cds = require('@sap/cds')
 
+const databaseModel = cds.linked({
+  definitions: {
+    schemas: new cds.entity({
```
David-Kunz, line 8: maybe a more distinct name, like `cds_test_schemas` to avoid clashes?

BobdenOs: The model is inside the `database` isolation. It actually clashes with `SYS.SCHEMAS` in `HANA`, but as this is isolated it doesn't matter.

- SYS
  - Database 1 <- database model
     - Tenant 1 <- application model
     - ...
   - ...

db-service/lib/common/DatabaseService.js
```diff
@@ -3,15 +3,192 @@ const ConnectionPool = require('./generic-pool')
...
+databaseModel.meta = {}
+
 /** @typedef {unknown} DatabaseDriver */
 
 class DatabaseService extends cds.Service {
 
-  init() {
+  async init() {
     cds.on('shutdown', () => this.disconnect())
+    if (Object.getOwnPropertyDescriptor(cds, 'test') || this.options.isolate) {
```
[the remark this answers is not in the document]
oklemenz2: should such test DB setup not be a module outside of the productive code?

hm23 (changes requested):
Looks good to me!

However, I only manually tested with a single `cds.test` test suite in the @sap/cds module so far, so I can't say if the "shared deployment" approach really works.

Obviously, one still needs to provide the hana credentials. I did so by modifying `cds.env.requires.db` in a beforeAll, which I added as part of the jest hana setup. I don't know if there is a nicer solution for this...

danjoa:
Is that still to be followed up?

hm23:
> Is that still to be followed up?

Probably @BobdenOs / @johannes-vogel can answer this, I was not involved in recent discussions about this PR.

I'd be happy if I could run `cds.test` tests on hana without having to take care about deployment/isolation. I played around with this approach last week, but I only tested with a single `cds.test` test suite. I saw that it interferes with our old and ugly `cds/tests/_runtime/utils/setup.js`, and I could imagine that the same could happen for stakeholders. So I would vote for hiding this behind a flag for now.

danjoa (changes requested):
As mentioned, also for `cds.test.data`, I'm really not in favour of these kinds of free lunch automations. Tests should care about their concrete setup and cleanup needs and not just offload that duty.

BobdenOs:
> As mentioned, also for `cds.test.data`, I'm really not in favour of these kinds of free lunch automations. Tests should care about their concrete setup and cleanup needs and not just offload that duty.

I agree with this sentiment. It seems to me that the current state of `cds.test` with sqlite has created a different expectation from the stakeholders. Where no tests care about anything. The test just calls the functions and don't know what it is actually doing.

For the cds HANA tests I will start discussions with the team for potential alternative solutions.

PR #954, opened 2024-12-18
fix: Force `sql_simple_queries` to quote in outer select

BobdenOs:


johannes-vogel (approved):

hana/lib/HANAService.js
```diff
@@ -426,7 +426,7 @@ class HANAService extends SQLService {
           q.as = q.SELECT.from.as
         }
 
-        if (rowNumberRequired || q.SELECT.columns.length !== aliasedOutputColumns.length) {
+        if (cds.env.features.sql_simple_queries || rowNumberRequired || q.SELECT.columns.length !== aliasedOutputColumns.length) {
```
soccermax, line 429: wouldn't it be enough to only do this for `orderBy` or `having` clauses? Or are there more cases where the nesting is really needed?

PR #960, opened 2024-12-19
fix: Improve `sql_simple_queries` flag

BobdenOs:
`sql_simple_queries` shouldn't exist.

This change makes it so that only `HANA` knows about the `sql_simple_queries` flag and it is no longer numeric. It now is either `truthy` or `falsy`. There is a gap in `hdb` which uses an older protocol version which uses `TINYINT` as a fallback for the lack of `BOOLEAN` support when using `@sap/hana-client` this limitation is no longer there. 

Which means that if applications have columns modeled as `cds.uint8` they are not allowed to use `sql_simple_queries` and `hdb` as driver. Either do the default behavior of not using `sql_simple_queries` or use `@sap/hana-client` as driver.

### DIFF
--- a/hana/lib/drivers/hdb.js
+++ b/hana/lib/drivers/hdb.js
@@ -9,7 +9,7 @@ const iconv = require('iconv-lite')
 const { driver, prom, handleLevel } = require('./base')
 const { isDynatraceEnabled: dt_sdk_is_present, dynatraceClient: wrap_client } = require('./dynatrace')
 
-if (cds.env.features.sql_simple_queries === 3) {
+if (cds.env.features.sql_simple_queries) {
   // Make hdb return true / false
   const Reader = require('hdb/lib/protocol/Reader.js')
   Reader.prototype._readTinyInt = Reader.prototype.readTinyInt
--- a/postgres/lib/PostgresService.js
+++ b/postgres/lib/PostgresService.js
@@ -397,22 +397,11 @@ GROUP BY k
         }
         return col
       })
-      const isRoot = SELECT.expand === 'root'
-      const isSimple = cds.env.features.sql_simple_queries &&
-        isRoot && // Simple queries are only allowed to have a root
-        !Object.keys(q.elements).some(e =>
-          q.elements[e].isAssociation || // Indicates columns contains an expand
-          q.elements[e].$assocExpand || // REVISIT: sometimes associations are structs
-          q.elements[e].items // Array types require to be inlined with a json result
-        )
-
-      const subQuery = `SELECT ${cols} FROM (${sql}) as ${queryAlias}`
-      if (isSimple) return subQuery
 
       // REVISIT: Remove SELECT ${cols} by adjusting SELECT_columns
       let obj = `to_jsonb(${queryAlias}.*)`
-      return `SELECT ${SELECT.one || isRoot ? obj : `coalesce(jsonb_agg (${obj}),'[]'::jsonb)`
-        } as _json_ FROM (${subQuery}) as ${queryAlias}`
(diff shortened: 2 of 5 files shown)


hana/lib/drivers/hdb.js
```diff
@@ -9,7 +9,7 @@ const iconv = require('iconv-lite')
 const { driver, prom, handleLevel } = require('./base')
 const { isDynatraceEnabled: dt_sdk_is_present, dynatraceClient: wrap_client } = require('./dynatrace')
 
-if (cds.env.features.sql_simple_queries === 3) {
+if (cds.env.features.sql_simple_queries) {
```
oklemenz2, line 12: **OK, let's see:**

`features.sql_simple_queries: 1`: Not necessary, as always coming as boolean, right? And is default. Here we don't want to patch hdb
`features.sql_simple_queries: 2`: Does not make sense anymore at all, as not needed for HANA (sq3) and accepted for sqlite (sq1)
`features.sql_simple_queries: 3`: Yes, here it should be applied

sqlite/lib/SQLiteService.js
```diff
@@ -213,10 +213,7 @@ class SQLiteService extends SQLService {
       struct: expr => `${expr}->'$'`,
       array: expr => `${expr}->'$'`,
       // SQLite has no booleans so we need to convert 0 and 1
-      boolean:
-        cds.env.features.sql_simple_queries === 2
-          ? undefined
-          : expr => `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
+      boolean: expr => `CASE ${expr} when 1 then 'true' when 0 then 'false' END ->'$'`,
```
oklemenz2, line 216: OK!

postgres/lib/PostgresService.js
```diff
@@ -397,22 +397,11 @@ GROUP BY k
...
-        )
-
-      const subQuery = `SELECT ${cols} FROM (${sql}) as ${queryAlias}`
-      if (isSimple) return subQuery
 
       // REVISIT: Remove SELECT ${cols} by adjusting SELECT_columns
       let obj = `to_jsonb(${queryAlias}.*)`
-      return `SELECT ${SELECT.one || isRoot ? obj : `coalesce(jsonb_agg (${obj}),'[]'::jsonb)`
-        } as _json_ FROM (${subQuery}) as ${queryAlias}`
+      return `SELECT ${SELECT.one || SELECT.expand === 'root' ? obj : `coalesce(jsonb_agg (${obj}),'[]'::jsonb)`
```
oklemenz2, line 403: Cannot judge

hana/lib/drivers/hdb.js
oklemenz2: > This PR achieves the same behavior for all databases without having to differ between sql_simple_queries:1|2|3 by having all databases return true and false for Booleans.

Yes, that's good ! But I would not extra patch `hdb`, when anyways the booleans come correctly from query ...

BobdenOs: They don't come correctly from the query when using `hdb` as the protocol version it uses give the metadata as tinyint not as Boolean. That is why it is being patched instead of rendering `json` queries.

I tried to find the place where `hdb` defines the protocol version, but was not successful to change the behavior of HANA to match that for hana-client. A protocol upgrade probably comes with more features then just native Boolean support. So a feature request to `hdb` will most definitely be rejected.

oklemenz2: I thought that boolean comes from JSON function for `simple_queries: 1`, isn't it?

danjoa (changes requested):
I wanted all CAP db services to behave the same by _not_ converting booleans at all, but behave like all db drivers in this regard and handle them as truthy/falsy values. 

Seeing that PR brings me to the conclusion that we should seriously do as described above with cds9. All apps which have a problem with that can add generic post processing

PR #975, opened 2025-01-09
fix functions always return true or false

larslutz96:


db-service/lib/cql-functions.js
```diff
@@ -75,15 +75,15 @@ const StandardFunctions = {
    * @param {string} y
    * @returns {string}
    */
-  startswith: (x, y) => `instr(${x},${y}) = 1`, // sqlite instr is 1 indexed
+  startswith: (x, y) => `coalesce(instr(${x},${y}) = 1,false)`, // sqlite instr is 1 indexed
```
patricebender, line 78: this covers only SQLite, is this also relevant for other dbs?:)

larslutz96: added same for postgres

David-Kunz (approved):
👍

patricebender (approved):

PR #980, opened 2025-01-15
fix: contains not evaluting to bool

PDT42:
Compare the result of sqlite contains realization using `instr` to 0, to make it evaluate to a boolean. The same problem does not exist for HANA or Postgres.


patricebender (approved):
💯

PR #981, opened 2025-01-16
fix: nested ternary in calculated element

patricebender:
calculated elements may reference other calcuated elements and so forth.. that is why we have logic in place which constructs subpaths for such scenarios which enables us to calculate joins deeply built into a given calculated element hierarchy. However, the logic had a flaw for e.g. nested ternary expressions.

For `(1 > 0 ? 1 : (book.stock > 10 ? value : 3));`

we accidentally constructed subpaths like `book.stock.value` which led to an error because we tried to create a join for `value`, where no association prefix is present.

fix cap/issue#17660

### DIFF
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -859,7 +859,7 @@ function infer(originalQuery, model) {
       } else if (arg.xpr || arg.args) {
         const prop = arg.xpr ? 'xpr' : 'args'
         arg[prop].forEach(step => {
-          const subPath = { $refLinks: [...basePath.$refLinks], ref: [...basePath.ref] }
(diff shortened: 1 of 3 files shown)


stewsk (approved):

stewsk:
@patricebender can you please checks whether this works in the compiler?

patricebender:
> @patricebender can you please checks whether this works in the compiler?

I did and it doesn't work. Gute Nase! Opened an internal issue.

stewsk:
@patricebender 
> Opened an internal issue.
can you send me the link?

patricebender:
> @patricebender 
> 
> > Opened an internal issue.
> 
> 
> 
> can you send me the link?

Fixed already /cap/cds-compiler/issues/13397

PR #982, opened 2025-01-17
chore: update all dev and peer deps

johannes-vogel:


patricebender (approved):
This should resolve https://github.com/cap-js/cds-dbs/security/dependabot/26

what command did you use?

PR #985, opened 2025-01-21
refactor: standard database functions

patricebender:
- [x] factor out SQLite specific functions from base class
- [x] `HANAService` also requires `db-service/cql-functions` and then also supports `floor`, `ceiling`, `average` … 
- [x] kill `totalseconds`
- [ ] (later) `cql-functions` in the db-service becomes `sql-functions` and a new interface `cql-functions` is added containing the set of functions which shall be implemented by the dbs themselves this would enable also overwriting/re-using `cql-functions` for non-sql dbs


db-service/lib/cql-functions.js
```diff
@@ -92,266 +75,116 @@ const StandardFunctions = {
    * @example
    * // returns 'b'
    * {func:'substring',args:[{val:'abc'},{val:1},{val:1}]}
-   * @param {string} x
-   * @param {string} y
-   * @param {string} z
-   * @returns {string}
+   * @param {string} x - The string input
+   * @param {string} y - The starting position
+   * @param {string} z - The length of the substring
```
BobdenOs, line 80: `z` is optional

db-service/lib/cql-functions.js
```diff
@@ -92,266 +75,116 @@ const StandardFunctions = {
...
-   * @returns {string}
-   */
-  nano100_between: (x, y) => `(julianday(${y}) - julianday(${x})) * 864000000000`,
   /**
-   * Generates SQL statement that calculates the difference in seconds between two timestamps
-   * @param {string} x left timestamp
-   * @param {string} y right timestamp
-   * @returns {string}
+   * Generates SQL statement for the current date
+   * @param {string} [p] - Optional parameter
```
BobdenOs, line 170: is `p` allowed according to ANSI SQL ? My understanding is that `p` is only allowed by certain databases.

patricebender: No, at least not for `current_date` (I didn't add this in this PR though),

this and excerpt from ANSI SQL 2003:

```
<current date value function> ::= CURRENT_DATE

<current time value function> ::=
CURRENT_TIME [ <left paren> <time precision> <right paren> ]

<current local time value function> ::=
LOCALTIME [ <left paren> <time precision> <right paren> ]

<current timestamp value function> ::=
CURRENT_TIMESTAMP [ <left paren> <timestamp precision> <right paren> ]

<current local timestamp value function> ::=
LOCALTIMESTAMP [ <left paren> <timestamp precision> <right paren> ]
```

Looks like the precision is ok for the others. I would then remove it from `current_date` would you agree?

BobdenOs (approved):

PR #989, opened 2025-01-22
fix(`UPDATE`): only perform subselect matching if necessary

patricebender:
the issue was that while calculating the `UPDATE.elements`, we accidentially assumed that the calculated element would lead to a join, hence we did our subselect matching. This logic is not necessary for non `SELECT` queries --> there we need joins if we do a `SELECT * from <entity with calculated element /w path expression>`

---

fix(`UPDATE`): no assocs in list which matches subquery results


BobdenOs (approved):

danjoa (approved):

danjoa:
P.S. We might want to cove all the cases above also in our tests

PR #990, opened 2025-01-22
fix: cqn transformation is OData agnostic

patricebender:


db-service/lib/cqn4sql.js
```diff
@@ -796,8 +796,10 @@ function cqn4sql(originalQuery, model) {
     // columns of the expand must be in the groupBy
     if (transformedQuery.SELECT.groupBy) {
       const baseRef = column.$refLinks[0].definition.SELECT || ref
-
-      return _subqueryForGroupBy(column, baseRef, columnAlias)
+      const wildcardIndex = column.expand.findIndex(e => e === '*')
+      if (wildcardIndex === -1) {
```
danjoa, line 800:
```suggestion
      if (column.expand.includes('*')) {
```

danjoa (approved):

danjoa (changes requested):
Would that mean that:

1. For `SELECT Items {*} from Orders group by Items.ID` we would _not_ skip the expand?
2. For `SELECT Items { ID } from Orders group by Items.ID` we would skip the expand?

→ still seems wrong

danjoa:
As discussed we need to revisit the way we translate OData `$apply` requests to _valid_ queries, and have those queries translated to native SQL without any assumption of its origin from OData. → https://github.com/cap-js/cds-dbs/pull/990/commits/bed7329671f3a82b5220ab78784d0ab1e6c4bb6a adds comments to relevant tests about expected valid queries.

danjoa:
Moreover, and more importantly, we have to ensure that we correctly implement group by, also for things which OData cannot express. In particular, for expands of to-many associations like that: 

```sql
SELECT from Orders {
   ID, sum (Items.product.price * Items.quantity) as total,
   Items { product.name, quantity }
} group by ID
```

test/scenarios/bookshop/read.test.js
```diff
@@ -42,6 +42,7 @@ describe('Bookshop - Read', () => {
   test('Books with groupby with path expression and expand result', async () => {
     const res = await GET(
       '/admin/Books?$apply=filter(title%20ne%20%27bar%27)/groupby((author/name),aggregate(price with sum as totalAmount))',
+      //> SELECT author { name }, sum(price) as totalAmount from Books where title != 'bar' GROUP BY author.name
```
agoerler, line 45: or, you just need to ensure that you select the author's name in SQL

```suggestion
      //> SELECT author.name, sum(price) as totalAmount from Books where title != 'bar' GROUP BY author.name
```

test/scenarios/bookshop/read.test.js
```diff
@@ -50,6 +51,7 @@ describe('Bookshop - Read', () => {
   test('same as above, with foreign key optimization', async () => {
     const res = await GET(
       '/admin/Books?$apply=filter(title%20ne%20%27bar%27)/groupby((author/name, author/ID),aggregate(price with sum as totalAmount))',
+      //> SELECT author { name, ID }, sum(price) as totalAmount from Books where title != 'bar' GROUP BY author.name, author.ID
```
agoerler, line 54: see above:

```suggestion
      //> SELECT author.name, author.ID, sum(price) as totalAmount from Books where title != 'bar' GROUP BY author.name, author.ID
```

patricebender:
> As discussed we need to revisit the way we translate OData `$apply` requests to _valid_ queries, and have those queries translated to native SQL without any assumption of its origin from OData. → [bed7329](https://github.com/cap-js/cds-dbs/commit/bed7329671f3a82b5220ab78784d0ab1e6c4bb6a) adds comments to relevant tests about expected valid queries.

Agree 👍



> Moreover, and more importantly, we have to ensure that we correctly implement group by, also for things which OData cannot express. In particular, for expands of to-many associations like that:
> 
> ```sql
> SELECT from Orders {
>    ID, sum (Items.product.price * Items.quantity) as total,
>    Items { product.name, quantity }
> } group by ID
> ```

Also agree, however I would make vote for a two step approach:

1. revert changes in db-service which implemented OData stuff on cqn transformation level, to be OData agnostic again → coordinate with runtime to make the OData adapter come up with a proper CQN which is then transformed as usual.
2. Come up with a plan to implement group by, which also depends on the actual DB-Driver we are talking to

patricebender:
@johannes-vogel I reverted the special handling of expand + group by which we added due to issues of OData $apply + $expand. Now there are some tests failing (the ones Daniel already commented on)

I think the next step would be for the runtime to patch the OData adapter in a way, that the CQNs are generated so that the tests will work again, even without cqn4sql doing some magic for those cases. Will you dispatch that?

PR #991, opened 2025-01-22
feat: path expressions inside filter of scoped queries

patricebender:


db-service/test/cqn4sql/where-exists.test.js
```diff
@@ -1567,6 +1567,107 @@ describe('path expression within infix filter following exists predicate', () =>
...
+      inner join bookshop.Genres as genre on genre.ID = books.genre_ID
+      { books.ID }
+        WHERE EXISTS (
+          SELECT 1 from bookshop.Authors as Authors where Authors.ID = books.author_ID
+        )
+        and genre.name = NULL`,
+    )
+  })
+  it('renders nested inner joins for the path expression at the leaf of scoped queries', () => {
+    let query = CQL`SELECT from bookshop.Authors:books[genre.parent.name = null] { ID }`
```
stewsk, line 1585: What about books that don't have a genre? Should they also appear in the result?

stewsk:
There should be a test where there is a to-many association in the filter of the last path step.
Simply translating that to a JOIN would not be right, in my opinion, as this would potentially increase the size of the result set rather than reducing it (this is a mistake that we have already made when dealing with to-many assocs in the WHERE clause

patricebender:
> There should be a test where there is a to-many association in the filter of the last path step. Simply translating that to a JOIN would not be right, in my opinion, as this would potentially increase the size of the result set rather than reducing it (this is a mistake that we have already made when dealing with to-many assocs in the WHERE clause

👍 having an inner join and not an left outer join for all filter path expressions should take care of that, shouldnt it?

stewsk:
> having an inner join and not an left outer join for all filter path expressions should take care of that, shouldnt it?

no, concerning "duplication" there is no difference between INNER and LEFT OUTER

db-service/test/cqn4sql/where-exists.test.js
patricebender: hmm now that I think about it… it should have the same semantics as if you would write the expression into the queries `.where` clause. For this case, also books without a `genre` will be part of the result set due to the `left join`. What do you think?

PR #992, opened 2025-01-23
fix: groupby path expression with overlapping identifier

PDT42:
Currently, the HANA service is unable to `GROUP BY foobar.foobar`, raising an error like `invalid column name: foobar.foobar`, if the query includes an aggregation and `sql syntax error: incorrect syntax near "FROM"`, if it does not. The problem is caused by the overlapping column names and the HANA service not correctly injecting the column groupby requires into the parent select - if the parent includes a column with the same `column_name` as the 'grouped-by' column. 

This fix enables the required injection, by skipping the overlapping names check for columns that contain a select.

### DIFF
--- a/test/compliance/SELECT.test.js
+++ b/test/compliance/SELECT.test.js
@@ -638,8 +638,7 @@ describe('SELECT', () => {
 
     test('navigation with duplicate identifier in path', async () => {
       const { Books } = cds.entities('complex.associations')
-      const cqn = CQL`SELECT name { name } FROM ${Books} GROUP BY name.name`
-      const res = await cds.run(cqn)
+      const res = await cds.ql`SELECT name { name } FROM ${Books} GROUP BY name.name`
       assert.strictEqual(res.length, 1, 'Ensure that all rows are coming back')
     })
 
--- a/test/compliance/SELECT.test.js
+++ b/test/compliance/SELECT.test.js
@@ -638,7 +638,14 @@ describe('SELECT', () => {
 
     test('navigation with duplicate identifier in path', async () => {
       const { Books } = cds.entities('complex.associations')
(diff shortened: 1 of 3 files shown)


test/compliance/resources/db/complex/associations.cds
```diff
@@ -4,6 +4,7 @@ entity Books {
   key ID : Integer;
   title  : String(111);
   author : Association to Authors;
+  name   : Association to Authors on $self.author.ID = ID;
```
BobdenOs, line 7: I think that the `ID` ref is referencing the `Books.ID` is should be `name.ID` to reference the `Authors.ID`.
```suggestion
  name   : Association to Authors on $self.author.ID = name.ID;
```

test/compliance/SELECT.test.js
```diff
@@ -635,6 +635,20 @@ describe('SELECT', () => {
       const res = await cds.run(cqn)
       assert.strictEqual(res.length, 3, 'Ensure that all rows are coming back')
     })
+
+    test('navigation with duplicate identifier in path', async () => {
+      const { Books } = cds.entities('complex.associations')
+      const cqn = CQL`SELECT name { name } FROM ${Books} GROUP BY name.name`
+      const res = await cds.run(cqn)
+      assert.strictEqual(res.length, 1, 'Ensure that all rows are coming back')
+    })
```
BobdenOs, line 644: Just an heads up there is a new API for `CQL` which is `cds.ql` which can be executed directly making the test look like:

```suggestion
    test('navigation with duplicate identifier in path', async () => {
      const { Books } = cds.entities('complex.associations')
      const res = await cds.ql`SELECT name { name } FROM ${Books} GROUP BY name.name`
      assert.strictEqual(res.length, 1, 'Ensure that all rows are coming back')
    })
```

johannes-vogel (approved):

PR #1004, opened 2025-01-29
feat: support for `SELECT.hint`

schwma:
part of `cdsnode/issues/2186`

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -142,7 +142,7 @@ class HANAService extends SQLService {
 
     // REVISIT: add prepare options when param:true is used
     let sqlScript = isLockQuery || isSimple ? sql : this.wrapTemporary(temporary, withclause, blobs)
-    const{ hints } = query.SELECT
+    const { hints } = query.SELECT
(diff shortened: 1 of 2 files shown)


hana/lib/HANAService.js
```diff
@@ -141,7 +141,9 @@ class HANAService extends SQLService {
     const isSimple = temporary.length + blobs.length + withclause.length === 0
 
     // REVISIT: add prepare options when param:true is used
-    const sqlScript = isLockQuery || isSimple ? sql : this.wrapTemporary(temporary, withclause, blobs)
+    let sqlScript = isLockQuery || isSimple ? sql : this.wrapTemporary(temporary, withclause, blobs)
+    const{ hints } = query.SELECT
+    if (hints) sqlScript += ` WITH HINT (${Array.isArray(hints) ? hints.join(',') : hints})`
```
johannes-vogel, line 146:
```suggestion
    if (hints) sqlScript += ` WITH HINT (${hints.join(',')})`
```

hints is always an array, isn't it?

schwma: Yes, the ql `.hints()` API always creates an array. So as long as the property isn't set to something else by hand the check isn't needed.

PR #1008, opened 2025-02-03
test: add jaeger docker image with HANA trace information

BobdenOs:
This PR adds a `jaeger` and `prometheus` docker image to the `hce` docker setup. This allow the HANA Cloud instance to share its trace information with the `jaeger` Open Telemetry trace APIs. The value that the `prometheus` image provides is the ability to use the `monitoring` view of `jaeger`. Which aggregates the trace information to highlight possible performance pain points.

image

By leveraging `@cap-js/telemetry`, `SAP Passport` and `HANA` traces. It is possible to get a full stack performance overview while in a local development environment. Which is usually only possible in productive landscapes by using `dynatrace`. By having the detailed trace information locally it enables experimental analysis which is not possible in productive landscapes. Additionally removing the hyper scaler from the equation reduces the number of un controllable factors from the measurement (e.g. noise neighbors, network latency, etc.).

image

This feature combined with `explain` (https://github.com/cap-js/cds-dbs/pull/462) create a powerful foundation for ensuring a good performance base line and analysis of poor performing requests.

### DIFF
--- a/hana/lib/drivers/base.js
+++ b/hana/lib/drivers/base.js
@@ -17,7 +17,7 @@ class HANADriver {
    * @returns {import('@cap-js/db-service/lib/SQLService').PreparedStatement}
    */
   async prepare(sql) {
-    const prep = prom(
+    const prep = module.exports.prom(
       this._native,
       'prepare',
     )(sql).then(stmt => {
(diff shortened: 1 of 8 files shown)


patricebender (approved):
Sounds good. Would love to try it out but the hce image doesnt run on ARM macs… could I also attach this to a remote HANA?

Issue #994, opened 2025-01-24
Hana communicationTimeout not configurable


kasimos:
Currently in https://github.com/cap-js/cds-dbs/blob/8c86b863a69833d50cff91483150bf0314bb7258/hana/lib/drivers/hana-client.js#L33 the communicationTimeout for the Hana queries is hardcoded to 1 Minute (60.000ms). 
In many cases background jobs running through cds.spawn might tackle db intensive tasks, like cleanups, transformations or calculations that easily exceed the arbitrary 1 minute limit.
It would be good to have an option to configure this, as well as the other options that are currently commented out for the dbConnection.

PR #1024, opened 2025-02-11
feat: pass through of  arbitrary client options

johannes-vogel:
Pass through of arbitrary client options should be possible. The database services do not interpret this but simply pass through information to the respective db clients.

```json
{
  "db": {
     "kind": "hana",
     "credentials": { ... },
     "client": { "packetSize": 1048576},
  }
}
```
This sample refers to the [`packetSize` parameter of `hdb`](https://github.com/SAP/node-hdb?tab=readme-ov-file#controlling-the-maximum-packet-size).

For Hana MT and SQLite the "ugly reuse" of credentials is not possible.
In Hana MT, the credentials for the db connection are not in environment but provided by service manager.
In SQLite, the credentials and options are different parameters for the client.

SQLite options: https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#new-databasepath-options
Postgres options: https://node-postgres.com/apis/client#new-client
hdb options: https://github.com/SAP/node-hdb?tab=readme-ov-file#cesu-8-encoding-support
hana-client options: https://help.sap.com/docs/PRODUCT_ID/f1b440ded6144a54ada97ff95dac7adf/4fe9978ebac44f35b9369ef5a4a26f4c.html?state=PRODUCTION&version=latest&locale=en-US

closes https://github.com/cap-js/cds-dbs/issues/994
closes https://github.com/cap-js/cds-dbs/issues/920


johannes-vogel:
@BobdenOs Any recommendation for testing this? I only see the possibility for a test in each of the submodules.

BobdenOs:
It would be possible to make a test case which provides an object that has a `get` property with enumerable `true` where the test makes sure that the `get` was called. If it was not called it means the options are being ignored.

sqlite/test/deep/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "@test/cds-db-layer",
+  "version": "1.0.0",
+  "description": "Base for db layer validations",
+  "cds": {
+    "requires": {
+      "db": {
+        "impl": "@cap-js/sqlite",
+        "client": {
+          "xy": 55
+        }
```
johannes-vogel, line 11:
```suggestion
```

BobdenOs (approved):

Issue #1020, opened 2025-02-08
Table Constraints not generated for PostgreSQL When Using "cds build --production"


CHELLARAO-M:
### Table Constraints not generated for PostgreSQL
When building a CAP application targeting a PostgreSQL database using the `cds build --production` command, table constraints (e.g., FK constraints) are not generated in the csn file of gen folder. However, when using the `cds deploy --dry --model-only --out cds-model.csn ` command, these constraints are captured in the generated CSN file. 

For example, given the following schema:
```cds
using { managed } from '@sap/cds/common';
namespace sap.capire.bookshop;

entity Books : managed {
  key ID   : Integer;
  title    : localized String(111);
  author   : Association to Authors;
}

entity Authors : managed {
  key ID       : Integer;
  name         : String(111);
  books        : Association to many Books on books.author = $self;
}
```
#### Package.json
*Enabled generation of foreign key constraints*
`cds.features.assert_integrity = db` [db-constaints](https://cap.cloud.sap/docs/guides/databases#database-constraints)

```json
{
  "name": "Test_pg",
  "version": "1.0.0",
  "description": "A simple CAP project.",
  "repository": "<Add your repository here>",
  "license": "UNLICENSED",
  "private": true,
  "dependencies": {
    "@sap/cds": "^8",
    "express": "^4",
    "@sap/xssec": "^4",
    "@cap-js/postgres": "1.10.5"
  },
  "devDependencies": {
    "@cap-js/cds-types": "^0.9.0",
    "@cap-js/sqlite": "^1",
    "@sap/cds-dk": "^8"
  },
  "scripts": {
    "start": "cds-serve"
  },
  "cds": {
    "requires": {
      "auth": "xsuaa",
      "connectivity": true,
      "destinations": true,
      "application-logging": true
    },
    "features": {
      "assert_integrity": "db"
    }
  }
}

```
#### Expected behavior:

The generated csn file for PostgreSQL(`gen/pg`)  should include table constraints.
```json
"sap.capire.bookshop.Books": {
            "kind": "entity",
            "elements": {
                ...
            },
            "$tableConstraints": {
                "referential": {
                    "sap.capire.bookshop.Books_author": {
                        "identifier": "c__sap.capire.bookshop.Books_author",
                        "foreignKey": [
                            "author_ID"
                        ],
                        "parentKey": [
                            "ID"
                        ],
                        "dependentTable": "sap.capire.bookshop.Books",
[...]

PR #1032, opened 2025-02-18
feat: Persist assert_integrity feature

daogrady:
Fixes https://github.com/cap-js/cds-dbs/issues/1020


postgres/test/cds-build.test.js
```diff
@@ -27,4 +27,10 @@ describe('cds build plugin', () => {
     execSync('npx cds build --production', { cwd: workDir })
     expect(fs.existsSync(path.join(dbDest, 'csn.json'))).to.be.true
   })
+
+  test('should retain assert_integrity setting', () => {
+    execSync('CDS_FEATURES_ASSERT__INTEGRITY=db npx cds build --production', { cwd: workDir })
```
patricebender, line 32:
```suggestion
    execSync('CDS_FEATURES_ASSERT_INTEGRITY=db npx cds build --production', { cwd: workDir })
```
Is the double `_` intentional?

I would prefer to add the `cds.features.assert_integrity` directly into the `tiny-sample/package.json`

postgres/cds-plugin.js
```diff
@@ -25,11 +25,23 @@ cds.build?.register?.('postgres', class PostgresBuildPlugin extends cds.build.Pl
...
+        },
+        scripts: {
+          start: 'cds-deploy'
+        }
+      }
+      const assertIntegrity = cds.env?.features?.assert_integrity
+      if (assertIntegrity) {
+        packageJson.cds ??= {}
+        packageJson.cds.features ??= {}
+        packageJson.cds.features.assert_integrity = assertIntegrity
```
patricebender, line 41: We could already prepare for _other_ supported feature flags in a generic way, because once another feature flag will be added, this should be done anyway. Something like

postgres/test/cds-build.test.js
daogrady: Yes, the double `_` [is intentional](https://cap.cloud.sap/docs/guides/databases#database-constraints):
Screenshot 2025-02-19 at 11 08 53

Can gladly add it to the _package.json_, but that this setting is active for all tests. Probably not an issue, just didn't want to intrude in unrelated tests.

postgres/cds-plugin.js
daogrady: I actually started with a generic implementation that would take an arbitrary path into the `cds.env`, prepare all parent objects in _package.json_ based on the type of what `cds.env` dictacted and finally copied over all appropriate values. And before I knew it, I had a 200 loc function, that somewhat paled in comparison to this 5 line solution for the problem at hand. 😃
I therefore suggest to keep it simple as long as we are dealing with only one value and generalise it later, if we need to carry over more settings.

patricebender: I see. However, I am not really happy with just fixing the passthrough for this _one_ feature. I can imagine other feature flags which are important for a proper build, such as: naming modes (quoted or plain names?), standard database function mappings, …

@stewsk @johannes-vogel @chgeo we have the issue that build relevant options are not per default copied over to the postgres deployer app (found in `gen/pg`) do you have some idea how to properly copy over build relevant options (which are those?) to the deployer app? If we don't do that, an `mbt build && cf deploy` will lose all SQL relevant features defined in the app which e.g. has been the case for `cds.features.assert_integrity` in #1020

johannes-vogel: I don't think that there's a dedicated section for deployment relevant things. They are spread over different sections.

patricebender: Would it then in your opinion be sufficient to have a kind of allow-list approach, where we selectively copy over known build relevant options?

johannes-vogel: I don't know if an allow list makes sense if we do not know the options. How is that solved for the hana deployer?

patricebender: during build for HANA the actual db artifacts are already created with all features found in the cds env. Upon a deployment the `hdi` artifacts already have the desired features baked in.  In contrast to that, for postgres, only a inferred CSN is generated during build. Which does not contain any DB specific transformations, hence lacking the DB features set in the `cds env`. Upon deployment, the `gen/pg/csn.json` is transformed to the actual SQL schema. If the deployment is initiated in the `gen/pg/` folder, the root app `cds env` is ignored in favor of the the `cds env` of the deployer app (`gen/pg/package.json`) which is empty, invalidating all features set in the apps root.

postgres/test/tiny-sample/db/schema.cds
```diff
@@ -4,4 +4,9 @@ entity Books {
   key ID : Integer;
   title  : String;
   stock  : Integer;
+  author : Association to Authors;
 }
+
+entity Authors {
+    key ID : Integer;
+}
```
patricebender, line 12:
```suggestion
}

```

patricebender (approved):
I'd be fine if we merge this as-is now to unblock the OP of the issue. IMO we should still think about how the _full_ build configuration is propagated to the deployer app…

daogrady:
> I'd be fine if we merge this as-is now to unblock the OP of the issue. IMO we should still think about how the full build configuration is propagated to the deployer app…

Based on @johannes-vogel reply I don't see an immediate win here. But if you see more demand for further discussion, just let us know. 🙂

PR #1033, opened 2025-02-19
test: 2fast4us?

danjoa:
This assertion frequently fails locally on my M2 machine, presumably because it only works on slow machines?
If so what will happen when I get an M4? ;)


BobdenOs (approved):
🏎️

PR #1041, opened 2025-02-24
chore: Adjusting to consolidated cds.infer and query._target

danjoa:
Requires new release of @sap/cds with 4993


db-service/lib/SQLService.js
```diff
@@ -120,10 +120,10 @@ class SQLService extends DatabaseService {
...
     //          --> make cds.infer properly work with custom joins and remove this
-    if (!query.target) {
+    if (!(query._target instanceof cds.entity)) {
       try { this.infer(query) } catch { /**/ }
     }
-    if (query.target && !query.target._unresolved) {
+    if (!query._target?._unresolved) { // REVISIT: use query._target instead
       // Will return multiple rows with objects inside
       query.SELECT.expand = 'root'
     }
```
patricebender, line 129: @BobdenOs for which cases do we need this coding?

patricebender (commented):
changes in cqn4sql and infer lgtm, tests are green (if cds branch `cds.infer` is checked out)

db-service/lib/SQLService.js
BobdenOs: For when the target doesn't exist in the model it is not possible to provide output converters.

BobdenOs (approved):

PR #1042, opened 2025-02-24
fix: expand + groupby may return null, dont attach `.element`

patricebender:
this resolves an `Uncaught Object.defineProperty called on non-object` dump

this is a workaround for an issue reported in cap/issues#17907

--> we need to put all this logic in the OData protocol adapter, this coding should not be necessary for the DB. This initiative has already started in #990

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -877,8 +877,8 @@ function cqn4sql(originalQuery, model) {
 
       // to be attached to dummy query
       const elements = {}
-      const wildcardIndex = column.expand.includes('*')
-      if (wildcardIndex) {
+      const containsWildcard = column.expand.includes('*')
+      if (containsWildcard) {
         // expand with wildcard vanishes as expand is part of the group by (OData $apply + $expand)
         return null
       }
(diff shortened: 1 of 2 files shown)


db-service/lib/cqn4sql.js
```diff
@@ -877,8 +877,8 @@ function cqn4sql(originalQuery, model) {
 
       // to be attached to dummy query
       const elements = {}
-      const wildcardIndex = column.expand.findIndex(e => e === '*')
-      if (wildcardIndex !== -1) {
+      const wildcardIndex = column.expand.includes('*')
```
BobdenOs, line 880: It is no longer an `index`

db-service/lib/cqn4sql.js
```diff
@@ -910,7 +912,11 @@ function cqn4sql(originalQuery, model) {
           elements[c.as || c.ref.at(-1)] = c.element
         })
         return res
-      })
+      }).filter(c => c !== null)
```
BobdenOs, line 915: can `c` be anything `falsy`, but not `null` ?

patricebender: no, either its an object representing a column, or it is null (as a result of a wildcard expand). :)

BobdenOs: Great then we can simplify the check like: 

```suggestion
      }).filter(c => c)
```

johannes-vogel (approved):

PR #1048, opened 2025-02-26
fix: Support all `SELECT` properties in path expressions

BobdenOs:


test/scenarios/bookshop/read.test.js
```diff
@@ -232,6 +232,31 @@ describe('Bookshop - Read', () => {
...
 
+  test('apply SELECT properties from ref', async () => {
+    const { Books } = cds.entities('sap.capire.bookshop')
+    const query = cds.ql`SELECT FROM ${Books}[
+      where ID > 210
+      group by title
+      having computed
+      order by title desc
+      limit 2
+    ] {
```
patricebender, line 243: What are the semantics of those properties if they are not part of the leaf-ref of a from clause? E.g.:

```js
cds.ql`SELECT FROM Books {
      books[
        where ID > 210
        group by title
        having computed
        order by title desc
        limit 2
    ].title as top2
  }`
```

danjoa: - `order by`, `limit`, and `group by` are only reasonable for expands
- `where` is also reasonable for inlines

patricebender:
added in #1050 and #1049

PR #1049, opened 2025-02-27
feat: query modifiers on expand `ref` are propagated to subquery

patricebender:
enable the following:

```js
cds.ql`SELECT name, books[order by price] { * } FROM Authors`
```

it is still possible to have query modifiers defined as sibling property to `expand`:

```js
{
            ref: ['author'],
            expand: [
              {
                ref: ['name'],
              },
            ],
            limit: {
              offset: {
                val: 1,
              },
              rows: {
                val: 1,
              },
            },
            orderBy: [
              {
                ref: ['name'],
                sort: 'asc',
              },
            ],
          }
```

--> If both are set, the modifiers in the `ref` of the expanded association have precedence

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -807,11 +807,13 @@ function cqn4sql(originalQuery, model) {
     // `SELECT from Authors {  books.genre as genreOfBooks { name } } becomes `SELECT from Books:genre as genreOfBooks`
     const from = { ref: subqueryFromRef, as: uniqueSubqueryAlias }
     const subqueryBase = {}
-    for (const [key, value] of Object.entries(column)) {
-      if (!(key in { ref: true, expand: true })) {
-        subqueryBase[key] = value
-      }
+    const queryModifiers = { ...column, ...ref.at(-1) }
+    for (const [key, value] of Object.entries(queryModifiers)) {
+      if (key in { limit: 1, orderBy: 1, groupBy: 1, excluding: 1, where: 1, having: 1 }) subqueryBase[key] = value
     }
+    // where at leaf already part of subqueryBase
+    if (from.ref.at(-1).where) from.ref[from.ref.length - 1] = [from.ref.at(-1).id]
+
     const subquery = {
       SELECT: {
         ...subqueryBase,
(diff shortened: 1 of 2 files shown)


db-service/lib/cqn4sql.js
```diff
@@ -807,11 +807,13 @@ function cqn4sql(originalQuery, model) {
     // `SELECT from Authors {  books.genre as genreOfBooks { name } } becomes `SELECT from Books:genre as genreOfBooks`
     const from = { ref: subqueryFromRef, as: uniqueSubqueryAlias }
     const subqueryBase = {}
-    for (const [key, value] of Object.entries(column)) {
-      if (!(key in { ref: true, expand: true })) {
-        subqueryBase[key] = value
-      }
+    const queryModifiers = { ...column, ...ref.at(-1) }
+    for (const [key, value] of Object.entries(queryModifiers)) {
+      if (key in { limit: 1, orderBy: 1, groupBy: 1, excluding: 1, where: 1, having: 1 }) subqueryBase[key] = value
```
patricebender, line 812: everything covered?

BobdenOs: compared with [cqn2sql](https://github.com/cap-js/cds-dbs/blob/fddaa0877f6807086861d23abe92287a16be1814/db-service/lib/cqn2sql.js#L229) and lgtm.

danjoa: 👍 but you can take out `excluding` → that is always next to `expand` or `inline`. Also in CQL syntax as supported by the compiler, and hence the CQN produced by it: 

```js
cds.ql `SELECT from Authors { ID, name, books[order by title desc limit 1] {*} excluding { descr } }`
```
```js
⇒ cds.ql {
  SELECT: {
    from: { ref: [ 'Authors' ] },
    columns: [
      { ref: [ 'ID' ] },
      { ref: [ 'name' ] },
      {
        ref: [
          {
            id: 'books',
            orderBy: [ { ref: [ 'title' ], sort: 'desc' } ],
            limit: { rows: { val: 1 } }
          }
        ],
        expand: [ '*' ],
        excluding: [ 'descr' ]
      }
    ]
  }
}
```

danjoa (approved):

PR #1071, opened 2025-03-11
fix: Prioritize element name over implicit alias in path resolution

patricebender:
For recursive entity structures like:

```cds
entity Genres {
  ID:   Integer;
  name: String;
  parent: Association to Genres;
}
```

an OData request such as:

`http://localhost:4004/odata/v4/CatalogService/Genres(42)/dependsOn?$orderby=parent/name`

leads to a cqn like `select from Genres:parent order by parent.name`

this would previously resolve the first path step in `parent.name` as _implicit_ table alias `parent`, even though there is an element (the association `parent`)with the same name. With this change, the first reference of any multi-segment path is resolved using the following order:

1. Explicit table alias: If provided (e.g., in `select from Genres:parent as parent order by parent.name`), use the explicit alias.
2. Element reference: If no explicit alias exists, resolve the element (e.g., `parent` resolves to `Genres:parent`).
3. Implicit table alias: Fallback to the implicit alias.
4. Error: Throw an error if none of the above can be matched.

This update maintains compatibility with standard SQL implicit alias behavior while ensuring that element names take precedence where applicable.

### DIFF
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -423,10 +423,11 @@ function infer(originalQuery, model) {
     // if any path step points to an artifact with `@cds.persistence.skip`
     // we must ignore the element from the queries elements
     let isPersisted = true
-    let firstStepIsTableAlias, firstStepIsSelf, expandOnTableAlias
+    let firstStepIsImplicitTableAlias, firstStepIsExplicitTableAlias, firstStepIsSelf, expandOnTableAlias
     if (!inFrom) {
-      firstStepIsTableAlias = arg.ref.length > 1 && arg.ref[0] in sources
-      firstStepIsSelf = !firstStepIsTableAlias && arg.ref.length > 1 && ['$self', '$projection'].includes(arg.ref[0])
+      firstStepIsImplicitTableAlias = arg.ref.length > 1 && arg.ref[0] in sources
+      firstStepIsExplicitTableAlias = originalQuery.SELECT?.from.as && arg.ref.length > 1 && arg.ref[0] in sources
+      firstStepIsSelf = !firstStepIsImplicitTableAlias && arg.ref.length > 1 && ['$self', '$projection'].includes(arg.ref[0])
       expandOnTableAlias = arg.ref.length === 1 && arg.ref[0] in sources && (arg.expand || arg.inline)
     }
     if(dollarSelfRefs && firstStepIsSelf) {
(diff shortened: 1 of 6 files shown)


danjoa (changes requested):
I did a few quick tests in `cds repl -r @capire/bookshop`, with this added to _bookshop/db/schema.cds_:

```cds

@cds.persistence.skip: false
entity parent as select ID, name from Genres;
```

Works as expected:
```js
await SELECT `ID, name from Genres order by name`
await SELECT `ID, name as genres from Genres order by name`
await SELECT `ID, name, parent.name from Genres order by name`
await SELECT `ID, name, parent.name from Genres order by parent.name`
await SELECT `ID, name, parent.name from Genres as g order by parent.name`
await SELECT `ID, name from parent order by name`
```

Fails with errors:
```js
await SELECT `ID, name, parent.name from parent order by parent.name`
await SELECT `ID, name, parent.name from parent` 
await SELECT `ID, name, parent.name from Genres:parent` 
await SELECT `ID, name, parent.name from Genres as parent`
await SELECT `ID, name from parent as g order by parent.name`
```

Works as fixes for the above, but the above should work ootb: 
```js
await SELECT `ID, name, parent.name as parent_name from parent order by parent.name`
await SELECT `ID, name, parent.name as parent_name from parent`
await SELECT `ID, name, parent.name as parent_name from Genres:parent` 
await SELECT `ID, name, parent.name as parent_name from Genres as parent`
```
> Note: `as parent_name` is anyways the default I assumed we apply also for the failing ones?

Works, but seems wrong precedence:

```js
await SELECT `ID, name from parent order by parent.name`
```
→ seems to do the same as:
```js
await SELECT `ID, name from parent order by name`
```
Didn't we say that only explicit aliases have precedence? → i.e.:
```js
await SELECT `ID, name from Genres as parent order by parent.name`
```

patricebender:
> await SELECT `ID, name, parent.name from parent order by parent.name`

fails with `Uncaught Error: Duplicate definition of element “name”`

patricebender:
> await SELECT `ID, name, parent.name from parent order by parent.name`

there is no element `parent` in the projection, hence the `parent` refers to the _implicit_ table alias. Duplicate definition of element `name` rightfully thrown. 

same for 
```
await SELECT `ID, name, parent.name from parent`
```

> await SELECT `ID, name, parent.name from Genres as parent`

here rule number `1.` kicks in, the explicit table alias ha s precedence over element `Genres:parent` hence you select `name` twice. Duplicate definition of element `name` rightfully thrown.

> await SELECT `ID, name from parent as g order by parent.name`

there is no `parent` in `parent.elements` (the projection does not expose the association.) Also, explicit table alias `g` is assigned, hence `parent.name` can't be resolved. `"parent" not found in the elements of "sap.capire.bookshop.parent"` rightfully thrown.

> await SELECT `ID, name from parent order by parent.name`

"Works, but seems wrong precedence:" → No. There is no element `parent` in `parent.elements` (the projection does not expose the association) so there is only one name.

@danjoa please check

danjoa:
Last update from spec meeting: 

- Precedence rules stay as decided recently
- Implicit aliases should be always `$Author` and alike → as before but prefixed with `$` → **breaking change for Node.js**
- If developers want to use aliases in where, order by, ... clauses they should specify _explicit_ aliases
- If they used JOIN in their queries we can stay with implicit aliases w/o `$` → as of today

PR #1090, opened 2025-03-21
chore: use resolve

larsplessing:
Use cds.ql.resolve.
Adopting changes from pr: https://github.com/cap-js/cds-dbs/pull/977/files#diff-d5bbedeb9d69ec4e9a2a2cb63e32f88b5098749d0e2505bd8fefc6b6d33cf3eaR548


danjoa (dismissed):
As discussed: that should not be in cds.ql

David-Kunz (approved):

PR #1091, opened 2025-03-26
feat: support for HANA history tables

BobdenOs:
In the case that applications are using HANA history tables ([docs](https://help.sap.com/docs/SAP_HANA_PLATFORM/6b94445c94ae495c83a19646e7c3fd56/91302b26f62c4433bbc58e0a951cdc1d.html)) a newly added annotation `@hana.systemversioned` has to be added to the columns which HANA is automatically generating to ensure that write operations still function properly.

Additionally this PR makes it so that the `sap-valid-from`, `sap-valid-to` and `sap-valid-at` OData URL parameters will properly be enforced by the HANA history table results (e.g. time traveling).

### DIFF
--- a/test/compliance/resources/db/hana/versioning.cds
+++ b/test/compliance/resources/db/hana/versioning.cds
@@ -1,21 +1,22 @@
-using {
-  cuid,
-  temporal
-} from '@sap/cds/common';
+using {cuid} from '@sap/cds/common';
 
 namespace edge.hana.versioning;
 
 // The history table has to be defined before the origin table
 // As the compiler doesn't know the dependency between the two entities
 @readonly
-entity versioned.history : temporal {
-  ID   : UUID; // cuid doesn't work as it would make the ID column a key
-  data : String(5000);
+entity versioned.history { // : temporal
+  validFrom : Timestamp;
+  validTo   : Timestamp;
+  ID        : UUID; // cuid doesn't work as it would make the ID column a key
+  data      : String(5000);
 }
 
-entity versioned : cuid, temporal {
+entity versioned : cuid { // : temporal
+  validFrom : Timestamp;
+  validTo   : Timestamp;
   // Expose own history as an association
-  history : Association to many versioned.history
-              on history.ID = ID;
-  data    : String(5000);
+  history   : Association to many versioned.history
+                on history.ID = ID;
+  data      : String(5000);
 }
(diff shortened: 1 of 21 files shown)


test/compliance/resources/db/hana/versioning.cds
```diff
@@ -0,0 +1,21 @@
...
+  cuid,
+  temporal
+} from '@sap/cds/common';
+
+namespace edge.hana.versioning;
+
+// The history table has to be defined before the origin table
+// As the compiler doesn't know the dependency between the two entities
+@readonly
+entity versioned.history : temporal {
```
johannes-vogel, line 11: I think we should omit the temporal aspect to make it more clear that this is a hana native thing that's independent of CAP.

BobdenOs: I don't think that these tests are supposed to be used as documentation of the feature. As my objection would be the `@sql.append` being used to do [SQL injection](https://github.com/cap-js/cds-dbs/pull/1091/files#diff-d858da117a6320e5af8b52f6222fec1974b6d5cc63a2cb3ae2a5296d99743dcaR20).

johannes-vogel: I was simply confused by the additional workaround you had to do in `test/compliance/resources/fts/versioning/versioning.cds`. Surely, this is not supposed to be documentation but I was thinking that this makes it overly complex. We only want to ensure that our DML statement do not conflict with hana system versioning.

BobdenOs: I agree that this is a bit more "advanced" modeling, but it will be simpler then having to make a new mechanism to deploy custom HANA artifacts before doing the normal `cds.deploy`. As it would have to go through the whole process described [here](https://capire.cfapps.sap.hana.ondemand.com/docs/advanced/hana#add-native-objects).

johannes-vogel: I was more referring to `cds.valid.from: false` which stems from our temporal aspect. the `sql.append` magic I understood as trick to circumvent hana native artifacts for deployment.

is there a reason you included the cap temporal aspect in this test?

BobdenOs: My assumption was that it would be required for OData APIs that would be able to leverage the `valid-from/to/at` APIs as integrated [here](https://github.com/cap-js/cds-dbs/pull/1091/files#diff-11bd20d3012fd985095f5a9be4a1da3e94e09eac409545ac67688c9582d1d201R116), but for the test it doesn't have an impact just semantics. So if you think it is the wrong semantics I can pull the column definitions in from the aspect as well.

johannes-vogel (approved):

PR #1095, opened 2025-03-30
refactor: fixed resolve view

danjoa:
Requires @sap/cds 5132


David-Kunz (approved):
Just wanted to point that out in the cds PR 😄

David-Kunz:
To ease the pain, we can support both variants here and get rid of it once @sap/cds is released.

PR #1096, opened 2025-03-31
refactor: do not inline booleans

larsplessing:
Do not inline booleans. Use them as placeholder/value except for placeholders with boolean values in case when statements.


hana/lib/HANAService.js
```diff
@@ -1246,23 +1246,24 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction})) AS NEW LE
...
               const expression = {
-                xpr: ['CASE', 'WHEN', ...compare, 'THEN', { val: true }, 'WHEN', 'NOT', ...compare, 'THEN', { val: false }],
+                xpr: ['CASE', 'WHEN', ...compare, 'THEN', { val: true, cast: { type: 'cds.Boolean' } }, 'WHEN', 'NOT', ...compare, 'THEN', { val: false, cast: { type: 'cds.Boolean' } }],
                 _internal: true,
               }
 
               if (ifNull != null) {
                 // If at least one of the sides is NULL it will go into ELSE
                 // This case checks if both sides are NULL and return their result
+                const val = (v) => (typeof v === 'boolean' ? { val: v, cast: { type: 'cds.Boolean' } } : { val: v })
```
BobdenOs, line 1256:
```suggestion
```

`ifNull` can only be `true`, `false` and `null`. With the `ifNull != null` being the previous line it is garanteed that it is either a `true` or `false`. Please make sure that you read the code around what you are doing before introducing this kind of logic.

hana/lib/HANAService.js
```diff
@@ -1246,23 +1246,24 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction})) AS NEW LE
...
                 expression.xpr.push('ELSE', {
                   xpr: [
                     'CASE',
                     'WHEN',
                     // coalesce is used to match the left and right hand types in case one is a placeholder
                     ...[{ func: 'COALESCE', args: [left, right] }, 'IS', 'NULL'],
                     'THEN',
-                    { val: ifNull },
+                    val(ifNull),
```
BobdenOs, line 1264:
```suggestion
                    { val: ifNull, param: false },
```

hana/lib/HANAService.js
```diff
@@ -1246,23 +1246,24 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction})) AS NEW LE
...
                     'ELSE',
-                    { val: !ifNull },
+                    val(!ifNull),
```
BobdenOs, line 1266:
```suggestion
                    { val: ifNull, param: false },
```

hana/lib/HANAService.js
```diff
@@ -1246,23 +1246,24 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction})) AS NEW LE
               const compare = [left, x, right]
 
               const expression = {
-                xpr: ['CASE', 'WHEN', ...compare, 'THEN', { val: true }, 'WHEN', 'NOT', ...compare, 'THEN', { val: false }],
+                xpr: ['CASE', 'WHEN', ...compare, 'THEN', { val: true, cast: { type: 'cds.Boolean' } }, 'WHEN', 'NOT', ...compare, 'THEN', { val: false, cast: { type: 'cds.Boolean' } }],
```
BobdenOs, line 1249:
```suggestion
                xpr: ['CASE', 'WHEN', ...compare, 'THEN', { val: true, param: false }, 'WHEN', 'NOT', ...compare, 'THEN', { val: false, param: false }],
```

postgres/lib/PostgresService.js
```diff
@@ -454,8 +454,10 @@ GROUP BY k
     }
 
     val(val) {
-      const ret = super.val(val)
-      return ret === '?' ? `$${this.values.length}` : ret
+      let ret = super.val(val)
+      if(ret === '?') ret = `$${this.values.length}`
+      if(typeof val.val === 'boolean') ret = `cast( ${ret} as boolean)` 
+      return ret
```
BobdenOs, line 460:
```suggestion
      const ret = super.val(val)
      return ret === '?' ? `$${this.values.length}${typeof val.val === 'boolean' ? '::boolean' : ''}` : ret
```

BobdenOs (changes requested):

BobdenOs (approved):

danjoa:
Why don't we inline booleans any longer?

danjoa (changes requested):
See my question above

PR #1104, opened 2025-03-31
chore(deps): update dependencies

patricebender:
should resolve https://github.com/cap-js/cds-dbs/security/dependabot/27


patricebender:
nope you are right, however this should do the trick: https://github.com/cap-js/cds-dbs/commit/f1e275fa545c085fbc338b0438f2bc8c76921d66

PR #1093, opened 2025-03-28
fix: Adopt to recurse `DistanceTo` cqn format

BobdenOs:

PR #1111, opened 2025-04-01
chore: don't error on recurse queries

danjoa:
But simply run the supported parts of the query. 
Currently I created mock behaviour by doing this before sending the query to `cds.db` and all looks fine:  
```js
// Avoid error message: Feature "recurse" queries not supported.
delete SELECT.recurse 
delete SELECT.__proto__.recurse
```

Simply ignoring it / not knowing about it at all seems to be just fine

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -258,8 +258,9 @@ class CQN2SQLRenderer {
     return (this.sql = sql)
   }
 
-  SELECT_recurse() {
+  SELECT_recurse(q) {
     // cds.error`Feature "recurse" queries not supported.`
+    return this.from(q.SELECT.from, q)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


danjoa:
@BobdenOs pls check

BobdenOs:
@danjoa I have added the SQLite version for the 3 Hierarchy functions that we use to serve `recurse` queries [here](https://github.com/cap-js/cds-dbs/pull/1093/commits/e9bf15e0bfa5aee43756c49f24bf408f472d58b1#diff-d9ef4ae327f7e5954027eae8acf2866e34a6a365a11e60a3dc5ff535a210c33fR193). Which allows the `SELECT_recurse` implementation to be moved from the `HANAService` into the base `cqn2sql` and serve all databases. While it doesn't provide full functional correctness yet it is able to serve the same tree table requests as the `HANAService` implementation can. With the only current limitation being that it doesn't calculate the correct tree size.

BobdenOs:
Closing as duplicate of https://github.com/cap-js/cds-dbs/pull/1093

Issue #1112, opened 2025-04-02
[REGRESSION] `db-service` TypeError: (transformedQuery.SELECT.from.ref || transformedQuery.SELECT.from.SELECT.from.ref) is not iterable


hakimio:
### Description of erroneous behaviour 

Running the following query:
```
/MyEntities?$apply=filter(<filter>)/groupby((column1,column2),aggregate(amount with sum as totalAmount))
```

Results in an error:
```
[cds] - TypeError: (transformedQuery.SELECT.from.ref || transformedQuery.SELECT.from.SELECT.from.ref) is not iterable
    at expandColumn (node_modules\@cap-js\db-service\lib\cqn4sql.js:805:46)
    at handleExpand (node_modules\@cap-js\db-service\lib\cqn4sql.js:428:40)
    at node_modules\@cap-js\db-service\lib\cqn4sql.js:362:32
    at getTransformedColumns (node_modules\@cap-js\db-service\lib\cqn4sql.js:390:21)
    at transformSelectQuery (node_modules\@cap-js\db-service\lib\cqn4sql.js:191:41)
    at cqn4sql (node_modules\@cap-js\db-service\lib\cqn4sql.js:109:26)
    at PostgresService.cqn4sql (node_modules\@cap-js\db-service\lib\SQLService.js:402:17)
    at PostgresService.cqn2sql (node_modules\@cap-js\db-service\lib\SQLService.js:380:18)
    at PostgresService.onSELECT (node_modules\@cap-js\db-service\lib\SQLService.js:131:39)
    at PostgresService.onSELECT (node_modules\@cap-js\postgres\lib\PostgresService.js:326:18)
```

This is a regression because the same query works fine with `db-service` `v1.18.0`, but throws error with `v1.19.1`.

### Detailed steps to reproduce

1. Run the GET query mentioned above
2. Observe the result

### Details about your project

| Package            | Version                             |
| ---------------------- | ---------------------------------------------------------- |
| @cap-js/asyncapi       | 1.0.3                                                      |
| @cap-js/cds-types      | 0.9.0                                                      |
| @cap-js/change-trackin | 1.0.8                                                      |
| @cap-js/db-service     | 1.19.1                                                     |
| @cap-js/openapi        | 1.2.1                                                      |
| @cap-js/postgres       | 1.13.0                                                     |
| @sap/cds               | 8.9.0                                                      |
| @sap/cds-compiler      | 5.9.0                                                      |
[...]

PR #1114, opened 2025-04-02
fix: Error on missing nested SELECT.from.SELECT.from.ref

danjoa:
- fixes https://github.com/cap-js/cds-dbs/issues/1112#issuecomment-2772189618 ?
- @patricebender pls double check

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -802,7 +802,7 @@ function cqn4sql(originalQuery, model) {
     } else {
       outerAlias = transformedQuery.SELECT.from.as
       subqueryFromRef = [
-        ...(transformedQuery.SELECT.from.ref || /* subq in from */ transformedQuery.SELECT.from.SELECT.from.ref),
+        ...(transformedQuery.SELECT.from.ref || /* subq in from */ transformedQuery.SELECT.from.SELECT.from.ref || []),
         ...ref,
       ]
     }
@@ -1739,7 +1739,7 @@ function cqn4sql(originalQuery, model) {
            * with the main query alias. see @function expandColumn()
            * There is one exception:
            * - if current and next have the same alias, we need to assign a new alias to the next
-           *                          
+           *
            */
           if (!(inferred.SELECT?.expand === true && current.alias.toLowerCase() !== as.toLowerCase())) {
             as = getNextAvailableTableAlias(as)


johannes-vogel:
I think we have to add a test for this, before this release it was
```js
...(transformedQuery.SELECT.from.ref || /* subq in from */ transformedQuery.SELECT.from.target.name),
```

however, i don't know how it worked ...

hakimio:
@johannes-vogel before it was:
```js
...(transformedQuery.SELECT.from.ref || /* subq in from */ [transformedQuery.SELECT.from.target.name]),
```

https://github.com/cap-js/cds-dbs/commit/0828c25798ebe48b720c8be96d5beb4824502918#diff-3406bdae94230d4664b3a51d7efa5f3807f2a43095a9089ff018c5a5faf7ca3aL803-L806

johannes-vogel:
> It worked because of **...[transformedQuery.SELECT.from.target.name]**.

This part I understood. However, a query is not designed to have a `target` in `SELECT.from`.

danjoa:
Exactly I wondered how that would have worked before. We really should have a test

hakimio:
@johannes-vogel here is the answer:
> if we `select from <subquery>` we do not have a `SELECT.from.ref` but the `<subquery>` has a `.target` because it has been inferred properly.
>
> Use the name of the `.target` for building the correlated subquery, instead of dumping.

https://github.com/cap-js/cds-dbs/pull/709

danjoa:
SELECT.from never has a .target. 
Does it?

danjoa:
> > SELECT.from never has a .target. Does it?
> 
> it has a target if we select from another query. I have re-created the issue with the following bookshop query: ...

Either I still don't get it, or you didn't get my question... 

> SELECT.**from** never has a `.target`. Does it?

Given a CQN like that (regardless of inner or outer): 

```js
const q = {SELECT:{from:{ref:[...]}}}
```

Then to all my knowledge (being the author of CQN), a `.target` property might exist: 

1. In `q` (of course)
2. But never in `q.SELECT`
3. And never in `q.SELECT.from`

While the former code checked for `transformedQuery.SELECT.from.target.name`... so I wondered how that could ever have worked?  As you keep insisting that there exists a `.target` in `SELECT.from` pls elaborate when and why, and what I'm missing. 

Note: All this is not about opinions or beliefs but about CQN spec and compliance to it.

patricebender:
This PR was replaced by https://github.com/cap-js/cds-dbs/pull/1126

PR #1115, opened 2025-04-02
fix: forUpdate on complex queries (joins)

johannes-vogel:


David-Kunz (approved):

BobdenOs:
But what does the query with a join actually lock ? [Docs](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20fcf24075191014a89e9dc7b8408b26.html#:~:text=A%20table%20used%20in%20a%20subquery%20cannot%20be%20locked.%20Also%20only%20one%20table/view%20can%20be%20currently%20locked.)

johannes-vogel:
> But what does the query with a join actually lock ? [Docs](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20fcf24075191014a89e9dc7b8408b26.html#:~:text=A%20table%20used%20in%20a%20subquery%20cannot%20be%20locked.%20Also%20only%20one%20table/view%20can%20be%20currently%20locked.)

good question, maybe it locks only the left hand side?! would need to check. according to oliver it works

PR #1117, opened 2025-04-03
test: localized access in quoted mode

johannes-vogel:


db-service/lib/cqn2sql.js
```diff
@@ -23,9 +23,9 @@ class CQN2SQLRenderer {
     this.model = srv?.model
     // Overwrite smart quoting
     if (cds.env.sql.names === 'quoted') {
-      this.class.prototype.name = (name, query) => {
+      this.class.prototype.name = (name) => {
         const e = name.id || name
-        return (query?._target || this.model?.definitions[e])?.['@cds.persistence.name'] || e
+        return this.model?.definitions[e]?.['@cds.persistence.name'] || e
```
BobdenOs, line 28: Was `query._target` wrong ?

db-service/lib/cqn2sql.js
```diff
@@ -366,9 +366,9 @@ class CQN2SQLRenderer {
     if (ref) {
       let z = ref[0]
       if (z.args) {
-        return _aliased(`${this.quote(this.name(z, q))}${this.from_args(z.args)}`)
+        return _aliased(`${this.quote(this.name(z))}${this.from_args(z.args)}`)
```
BobdenOs, line 369: `from(from, q) {` the `q` is now no longer used.

hana/package.json
```diff
@@ -16,7 +16,9 @@
     "CHANGELOG.md"
   ],
   "scripts": {
-    "test": "(([ -z \"${HANA_HOST}\" ] && npm start) || true) && cds-test",
+    "test": "(([ -z \"${HANA_HOST}\" ] && npm start) || true) && npm run test:plain && npm run test:bookshop:quoted",
+    "test:bookshop:quoted":"cds_sql_names=quoted cds-test bookshop",
```
BobdenOs, line 20: Just the bookshop tests ?

patricebender (commented):
once #1140 is in main, this PR can be cleaned up

hana/package.json
johannes-vogel: the other tests are failing due to different reasons. The bookshop tests serve as a regression test as they already use localization and were adaptable with a reasonable amount of time.

PR #1118, opened 2025-04-03
fix(forUpdate): ignoreLocked has precedence over wait

David-Kunz:
Senseless to send such queries, but it can happen when global `sql.lock_acquire_timeout` is set.


David-Kunz:
> Sounds like we get rotten?

Yes, I will also fix the root cause in `@sap/cds`.

David-Kunz:
I think we can close this in favor of the other fix.

PR #1122, opened 2025-04-04
fix: Allow column names that have to be quoted

czindel-sap:
Following statement should be possible with the proposed change
```
await INSERT.into ("ENTITY_COL_BEGIN_WITH_NUMBER") .entries ({
  ID: 100,
  "1VALUE": 1,
});
```
currently this ends up with this error
```
Error inserting into ENTITY_COL_BEGIN_WITH_NUMBER Error: argument type mismatch: not supported path expression: Support only * or <KEY NAME>
-> $.1VALUE
->   ^

    at Socket.<anonymous> (/Users/D070420/SAPDevelop/cap-dev/node_modules/@sap/hana-client/lib/index.js:53:13)
    at Socket.emit (node:events:520:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5)
    at Pipe.onStreamRead (node:internal/stream_base_commons:191:23)
    at Pipe.callbackTrampoline (node:internal/async_hooks:130:17) {
  code: 293,
  sqlState: 'HY000',
  query: false,
  values: { ID: 100, '1VALUE': 1 }
}
```


czindel-sap:
@johannes-vogel It looks like the tests are not executed or stuck, is there something that I need to do?

johannes-vogel:
> @patricebender I see that a lot of the existing tests are failing. Can someone review if this is related to the change and if the change makes sense at all. If everything is good, I can make myself familiar with your test setup and prepare a test

@czindel-sap we already look into it. I don't think that this fix is sufficient/correct.

hana/lib/HANAService.js
```diff
@@ -1444,7 +1444,7 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction}) ERROR ON E
     managed_extract(name, element, converter) {
       // TODO: test property names with single and double quotes
       return {
-        extract: `${this.quote(name)} ${this.insertType4(element)} PATH '$.${name}', ${this.quote('$.' + name)} NVARCHAR(2147483647) FORMAT JSON PATH '$.${name}'`,
+        extract: `${this.quote(name)} ${this.insertType4(element)} PATH '$["${name}"]', ${this.quote('$.' + name)} NVARCHAR(2147483647) FORMAT JSON PATH '$["${name}"]'`,
```
BobdenOs, line 1447:
```suggestion
        extract: `${this.quote(name)} ${this.insertType4(element)} PATH '$."${name}"', ${this.quote('$.' + name)} NVARCHAR(2147483647) FORMAT JSON PATH '$."${name}"'`,
```

hana/lib/HANAService.js
```diff
@@ -1444,7 +1444,7 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction}) ERROR ON E
     managed_extract(name, element, converter) {
       // TODO: test property names with single and double quotes
       return {
-        extract: `${this.quote(name)} ${this.insertType4(element)} PATH '$.${name}', ${this.quote('$.' + name)} NVARCHAR(2147483647) FORMAT JSON PATH '$.${name}'`,
+        extract: `${this.quote(name)} ${this.insertType4(element)} PATH '$."${name}"', ${this.quote('$.' + name)} NVARCHAR(2147483647) FORMAT JSON PATH '$."${name}"'`,
```
BobdenOs, line 1447:
```suggestion
        extract: `${this.quote(name)} ${this.insertType4(element)} PATH ${HANAVERSION <= 2 ? `'$.${name}'` : `'$["${name}"]'`}, ${this.quote('$.' + name)} NVARCHAR(2147483647) FORMAT JSON PATH ${HANAVERSION <= 2 ? `'$.${name}'` : `'$["${name}"]'`}`,

```

BobdenOs:
@czindel-sap The original change was correct, but the feature is not supported on HANA 2.0. Which is the reason why the tests where failing in the pipeline. The branch was updated with the `HANAVERSION` check to switch between the two different syntaxes. It would be good to move the logic out to a variable first so that it doesn't need to compute the escape string twice. For the time being it will just not be supported to have characters which have to be escaped in the column name when using HANA 2.0.

czindel-sap:
@patricebender So far I was not able to execute the tests on my machine. Is there a description how to run the tests?

patricebender:
> @patricebender So far I was not able to execute the tests on my machine. Is there a description how to run the tests?

To execute tests against HANA, I always use the `npm run test:remote` with env variables containing the credentials for the HANA machine:

```zsh
HANA_HOST=… HANA_PORT=443 HANA_USER=… HANA_PASSWORD=… npm run test:remote --file orderBy  -- --unmute
```

with `--file <your test>` only the desired test files should be asserted. :)

czindel-sap:
@johannes-vogel Maybe I do not understand the details of https://github.com/cap-js/cds-dbs/pull/1140, but how does it ensure that column names like "1VALUE" are supported? So far I had no time to prepare a test, but it would help us a lot to have one to ensure that these kind of names will work in future

BobdenOs:
@czindel-sap the PR number was wrong it was supposed to be #1141. Which has the changes of the current PR and has a little [test](https://github.com/cap-js/cds-dbs/pull/1141/files#diff-396b9a08d3da5c5aae85b37c08414fc5ab7995bff1528b23491c203bb5409632R22-R74). The test is currently disabled for the `github.com` pipeline as the required syntax is not supported by HANA 2.0, but these tests can be executed with HANA Cloud. Additionally the test found a gap in the `@cap-js/sqlite` syntax which is now also fixed.

johannes-vogel:
> @johannes-vogel Maybe I do not understand the details of https://github.com/cap-js/cds-dbs/pull/1140,

sorry, my mistake.

czindel-sap:
@BobdenOs @johannes-vogel 👍 Thanks a lot

PR #1113, opened 2025-04-02
chore: package-lock maintenance

johannes-vogel:

PR #1125, opened 2025-04-08
chore: add `@cap-js/cds-test` as devDependency

chgeo:
The old `cds.test` code gets deleted in `@sap/cds` 9, so be prepared and add the devDependency.

### DIFF
--- a/db-service/package.json
+++ b/db-service/package.json
@@ -26,6 +26,9 @@
   "dependencies": {
     "generic-pool": "^3.9.0"
   },
+  "devDependencies": {
+    "@cap-js/cds-test": "^0"
+  },
   "peerDependencies": {
     "@sap/cds": ">=7.9"
   },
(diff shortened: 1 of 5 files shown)


BobdenOs:
@chgeo it is caused by the updated `package-lock.json` where `@sap/cds` is now suddenly having a `peerDependency` to `eslint` ([code](https://github.com/cap-js/cds-dbs/pull/1125/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R126)). Which shouldn't ever be the case as `eslint` is a development library and shouldn't be getting pulled into productive applications. Which will currently happen as it is not flagged as an optional peer dependency.

BobdenOs:
Closing as duplicate of https://github.com/cap-js/cds-dbs/pull/1113

PR #1140, opened 2025-04-14
fix(infer): for localized queries, use `localized.<entity>` as `_target`

patricebender:
factored out some shared code of `infer` and `cqn4sql`, the change itself is quite small

- [x] todo: check if there is a better way of orchestrating `localized` and `getDefinition`


db-service/lib/infer/index.js
```diff
@@ -111,13 +113,13 @@ function infer(originalQuery, model) {
...
-        ? getImplicitAlias(first, useTechnicalAlias)
-        : getImplicitAlias(ref.at(-1).id || ref.at(-1), useTechnicalAlias));    
+        from.uniqueSubqueryAlias ||
+        from.as ||
+        (ref.length === 1
+          ? getImplicitAlias(first, useTechnicalAlias)
+          : getImplicitAlias(ref.at(-1).id || ref.at(-1), useTechnicalAlias))
       if (alias in querySources) throw new Error(`Duplicate alias "${alias}"`)
-      querySources[alias] = { definition: target, args }
+      querySources[alias] = { definition: getDefinition(localized(target)), args }
```
patricebender, line 122: the actual change:

db-service/test/cds-infer/source.test.js
```diff
@@ -151,11 +151,27 @@ describe('multiple sources', () => {
...
+  beforeAll(async () => {
+    model = await cds.load(__dirname + '/../bookshop/db/schema').then(m => cds.compile.for.nodejs(m))
+  })
+
+  it('_target of localized query is the localized version of the entity', () => {
+    const q = SELECT.localized`from bookshop.Books as Books {ID, title}`
+    let query = inferred(q, model)
+    let localizedBooks = model.definitions['localized.bookshop.Books']
+    expect(query).to.have.property('_target').that.equals(localizedBooks)
+  })
```
johannes-vogel, line 176:
```suggestion
  it('_target of localized query is the localized version of the entity', () => {
    const q = SELECT.localized`from bookshop.Books as Books {ID, title}`
    let query = inferred(q, model)
    let localizedBooks = model.definitions['localized.bookshop.Books']
    expect(query).to.have.property('_target').that.equals(localizedBooks)
  })

  it('_target of non localized query is the non localized version of the entity', () => {
    const q = SELECT`from bookshop.Books as Books {ID, title}`
    let query = inferred(q, model)
    let Books = model.definitions['bookshop.Books']
    expect(query).to.have.property('_target').that.equals(Books)
  })
```

db-service/lib/infer/index.js
patricebender: `localized` utility takes care of that:

```js
function localized(definition) {
      if (!isLocalized(definition)) return definition.name
      const view = getDefinition(`localized.${definition.name}`)
      return view?.name || definition.name
    }
```

db-service/lib/utils.js
```diff
@@ -40,32 +40,94 @@ function isCalculatedElement(def) {
...
+  }
+}
+
 // export the function to be used in other modules
 module.exports = {
   prettyPrintRef,
   isCalculatedOnRead,
   isCalculatedElement,
   getImplicitAlias,
+  getModelUtils,
```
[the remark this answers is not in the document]
johannes-vogel: I'd be in favor of clear utils which have one job. it seems to me we overshare information in closures which is often not needed. anyway, we can merge and refactor later if needed.

PR #1141, opened 2025-04-14
fix: Improved support for special characters in column names

BobdenOs:


test/compliance/resources/db/complex/keywords/complex.keywords.~%60!%40%23%24%25%5E%26*()_%2B-%3D'%22%5C%2F.%2C%5B%5D.js
patricebender, line 1: any reason for this particular filename? In #1186 it looks like it causes issues upon cloning on windows machines.

PR #1142, opened 2025-04-15
Prevent `Statement is dropped` Error

czindel-sap:
I can not clearly tell how to reproduce the `Statement is dropped` error, but I receive it sometimes. Maybe it depends on execution times.
For my understanding after the result is read, the result rest should be closed and the result should be returned (https://github.com/cap-js/cds-dbs/blob/main/hana/lib/drivers/hana-client.js#L115). After the result is returned the prepared statement will be dropped. But as there is an additional `executeQuery` call after the `close` it might happen that the statement is already dropped. I am not to deep in the details of the implementation, but I wonder if there is a scenario where this is needed.

### DIFF
--- a/hana/lib/drivers/hana-client.js
+++ b/hana/lib/drivers/hana-client.js
@@ -69,7 +69,10 @@ class HANAClientDriver extends driver {
         const stmt = await ret._prep
         // Create result set
         const reset = async function () {
-          if (this) await prom(this, 'close')()
+          if (this) {
+            await prom(this, 'close')()
+            return
+          }
           const rs = await prom(stmt, 'executeQuery')(values)
           rs.reset = reset
           return rs


czindel-sap:
@BobdenOs I tried to remove all the "noise" from our project and boiled it done to

```cds
service AppService {
    entity ENTITY_WITH_BINARY {
        key ID    : Integer;
            VALUE : cds.Binary;
    }
}
```
and a simple `http://localhost:4004/odata/v4/app/ENTITY_WITH_BINARY` request. I can still see in https://github.com/cap-js/cds-dbs/blob/main/hana/lib/drivers/base.js#L170 that the error occurred. I am not sure about the error handling, but it does not show up in the console.

hana/lib/drivers/hana-client.js
```diff
@@ -69,7 +69,10 @@ class HANAClientDriver extends driver {
         const stmt = await ret._prep
         // Create result set
         const reset = async function () {
-          if (this) await prom(this, 'close')()
+          if (this) {
+            await prom(this, 'close')()
+            return
+          }
```
BobdenOs, line 75:
```suggestion
          if (this) await prom(this, 'close')()
          if (!stmt.isValid()) cds.error`Stream source has been invalidated. Make sure to keep your transaction open until all streams are consumed.`
```

BobdenOs (commented):
Here is a stack trace to explain what is happening.

1. GET('/odata/v4/app/ENTITY_WITH_BINARY')
1. reading starts [all](https://github.com/cap-js/cds-dbs/blob/be025bdd3f11750be506a03b2f24354b5ce46f5e/hana/lib/drivers/hana-client.js#L69)
1. reading finishes and [calls reset](https://github.com/cap-js/cds-dbs/blob/be025bdd3f11750be506a03b2f24354b5ce46f5e/hana/lib/drivers/hana-client.js#L115)
1. reset will [close](https://github.com/cap-js/cds-dbs/blob/be025bdd3f11750be506a03b2f24354b5ce46f5e/hana/lib/drivers/hana-client.js#L72) the result set
1. once `close` detaches from the main thread [commit](https://github.com/cap-js/cds-dbs/blob/be025bdd3f11750be506a03b2f24354b5ce46f5e/hana/lib/HANAService.js#L1602) is called. Which closes all open statements.
1. when `close` is finished the `reset` function tries to open a new [result set](https://github.com/cap-js/cds-dbs/blob/be025bdd3f11750be506a03b2f24354b5ce46f5e/hana/lib/drivers/hana-client.js#L73). Which fails as `commit` has already dropped the statement.
1. The error `Statement is dropped` is thrown by the client and this error is passed along to any future `stream` consumers. If no streams are created the `error` is ignored.

Which works as intended even if the error message `Statement is dropped` is not very nice, but it shouldn't be thrown unless a `stream` is trying to escape its source transaction.

PR #1146, opened 2025-04-17
Hack to prevent busy wait for 404 errors 

swaldmann:
See this related open issue in the `generic-pool` repo: https://github.com/coopernurse/node-pool/issues/273.

PRs to fix this are stale unfortunately:
https://github.com/coopernurse/node-pool/pull/221

I fear there's no non-hacky way to fix this other than #815 or changing our architecture so the `cds.xt.serviceManager` call is not done inside `create` factory handler (probably a good idea anyway) → Let's discuss on Monday


swaldmann:
decided to go with https://github.com/cap-js/cds-dbs/pull/815

PR #1150, opened 2025-04-23
feat: move `compile.to.hdbtabledata` from `@sap/cds`

chgeo:
This is the first step to offload also HANA designtime funtions from `@sap/cds`. 

See cap/cds#5244 for counterpart.

Compatibility:
- When running against cds 8, this new code wins.
- No dependency to cds 9 needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1282,14 +1282,14 @@
       }
     },
     "node_modules/pg": {
-      "version": "8.15.1",
-      "resolved": "https://registry.npmjs.org/pg/-/pg-8.15.1.tgz",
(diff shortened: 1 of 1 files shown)


hana/package.json
```diff
@@ -53,6 +53,13 @@
         }
       },
       "db": "sql"
+    },
+    "hana": {
+      "table_data": {
+        "column_mapping": {
+          "LargeBinary": "decodeBase64"
+        }
+      }
```
danjoa, line 62: Should that be default?

danjoa (approved):

hana/package.json
chgeo: It is, that's why I added it here.  Can be disabled with `LargeBinary: false`.
Other types can be registered, too.

PR #1151, opened 2025-04-23
fix?: allow for outer query column references without alias

BobdenOs:
For some use cases it is not possible to apply and `alias` to the outer query, but it is required to reference columns from the root query inside the disjointed sub select definition.

```cds
using { sap.capire.bookshop as my } from '../db/schema';

@protocol: { odata: '/admin', rest }
service AdminService {
  @restrict: [
    {
      grant: ['UPDATE'],
      where: `exists(
        select 1 from AdminService.Authors as authors
        where authors.ID = author.ID -- author.ID = Books.author.ID
        and authors.name = $user
      )`
    },
    {
      grant: ['DELETE'],
      where: `author.ID = $user` // potential work around
    }
  ]
  entity Books as projection on my.Books;
  entity Authors as projection on my.Authors;
}
```

### DIFF
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -494,6 +494,15 @@ function infer(originalQuery, model) {
             definition: getDefinitionFromSources(sources, id),
             target: getDefinitionFromSources(sources, id),
           })
+        } else if (ref.length === 1 && inferred.outerQueries?.find(outer => id in outer.$combinedElements)) {
+          // outer query accessed without alias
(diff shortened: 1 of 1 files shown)


patricebender:
Reason why this doesnt work explained here (also in line with cds-compiler behavior):
https://github.com/cap-js/cds-dbs/blob/26f74b1b29f62656a03e48259ca16b834f5ad837/db-service/test/cqn4sql/table-alias.test.js#L1183-L1188

PR #1152, opened 2025-04-23
Make  HANAVERSION accessible in derived classes

czindel-sap:
In my specific case I can not use the default implementation for multi tenancy, so I have to override certain parts of the `HANAService`, which leads to the state that `HANAVERSION` remains 0. Therefore the fix from https://github.com/cap-js/cds-dbs/pull/1141 is not accessible


czindel-sap:
Sorry I missed that `CQN2HANA` is a separate class that has no access to the scope to which I moved `HANAVERSION`

czindel-sap:
@johannes-vogel Do you think this could be a feasible solution? Can you also please trigger the test execution

PR #1159, opened 2025-04-25
test: revert "chore(tests): remove compliance test symlink (#1083)"

johannes-vogel:
At the moment, sqlite compliance/scenario tests are only run in internal pipeline. Reason is that we run in our cds-dbs pipeline the tests of each workspace.

Screenshot 2025-04-25 at 11 01 57

This reverts commit cae7b85479c110896f34b91125cb3374c8ab5ec5.
Alternatively, we can also look for another solution to still run tests internally only once.


patricebender (approved):

danjoa:
Ok thanks. Still don't understand why the removal of a nested symlink skips a whole set of tests → that's what I meant with 'voodoo' . Fine with reverting for this reason and the time being ... But please don't fix that in cds-test runner, but here in this strange test setup/layout, as other runners and test tools might likely be equally confused.

PR #1161, opened 2025-04-25
fix: current_utctimestamp as default

etimr:
- HANA doesn't allow functions as default values. 
- `current_utctimestamp()` looks like an exception
- asked by stakeholder


test/compliance/resources/db/basic/common/basic.common.default.js
```diff
@@ -1,4 +1,5 @@
 const dstring = size => ({ d: 'default'.slice(0, size), o: 'not default'.slice(0, size) })
+const currentDate = () => ({ d: new Date().toJSON().slice(0, 10), o: '2000-01-01' })
```
BobdenOs, line 2: This test breaks when run before midnight `utc`

etimr: changed to `UTC`

BobdenOs (approved):

PR #1162, opened 2025-04-25
fix: fixed postgres pool `validate` implementation

### LINKED ISSUE #1158: `Postgres` pool `validate()` function always returns `undefined`

It seems when initially creating the [PostgresService](https://github.com/cap-js/cds-dbs/blob/43b8467816aa0cc60f86a7754ce190e17f9538a9/postgres/lib/PostgresService.js#L9) class, code was copy-pasted from [SQLiteService](https://github.com/cap-js/cds-dbs/blob/43b8467816aa0cc60f86a7754ce190e17f9538a9/sqlite/lib/SQLiteService.js#L26), as a result, pool `validate()` function is defined as `dbc => dbc.open`. The issue is that `open` property exists on `SQLite` client but the [Postgres client](https://node-postgres.com/apis/client) does not have this property and `validate()` always returns `undefined`. This kind of `validate()` function breaks `Postgres` `borrow` and `eviction` pool tests.
@danjoa can you share your thoughts about the issue?

1. Set pool `max` connections to > 1 and enable `testOnBorrow` option.
2. Debug to verify if connections are correctly reused.

### Additional thoughts

- Maybe it's time to consider adopting TypeScript to avoid issues like this or [this](https://github.com/cap-js/cds-dbs/pull/1114#issuecomment-2785497204). TS can also be adopted incrementally without the need to migrate the whole codebase in one go.
- [Generic Pool](https://github.com/coopernurse/node-pool) library seems to be unmaintained for a long time now with the latest version released 3 years ago. Maybe at least for Postgres it would make sense to use [pg-pool](https://node-postgres.com/apis/pool) instead?

| Package           | Version                             |
| ---------------------- | ---------------------------------------------------------- |
| @cap-js/asyncapi       | 1.0.3                                                      |
| @cap-js/cds-types      | 0.10.0                                                     |
| @cap-js/change-tracking | 1.0.8                                                      |
| @cap-js/db-service     | 1.20.0                                                     |
| @cap-js/openapi        | 1.2.1                                                      |
| @cap-js/postgres       | 1.14.0                                                     |
| @sap/cds               | 8.9.2                                                      |
| @sap/cds-compiler      | 5.9.2                                                      |
| @sap/cds-dk            | 8.9.2                                                      |
| @sap/cds-fiori         | 1.4.1                                                      |
| @sap/cds-foss          | 5.0.1                                                      |
| @sap/cds-mtxs          | 2.7.2                                                      |
| @sap/eslint-plugin-cds | 3.2.0                                                      |
| Node.js                | v22.13.1                                                   |

hakimio:
Fixes #1158

### DIFF
--- a/postgres/lib/PostgresService.js
+++ b/postgres/lib/PostgresService.js
@@ -50,7 +50,11 @@ class PostgresService extends SQLService {
             }),
         }
         const dbc = new Client({...credentials, ...clientOptions})
+
+        dbc.on('end', () => dbc.open = false);
+
         await dbc.connect()
+        dbc.open = true
         return dbc
       },
       destroy: dbc => dbc.end(),


postgres/lib/PostgresService.js
```diff
@@ -50,7 +50,12 @@ class PostgresService extends SQLService {
             }),
         }
         const dbc = new Client({...credentials, ...clientOptions})
+
+        dbc.on('error', error => this.emit('error', error))
+        dbc.on('end', () => dbc.open = false);
+
         await dbc.connect()
+        dbc.open = true
```
hakimio, line 58: Can also use a separate local variable to store connection state if setting a property on Postgres client is not ok.

BobdenOs: `this.emit('error')` probably doesn't do what you would expect it to do. All database services rely on the client failing once a query is fired. So unless a closed connection doesn't throw any error this behavior will be unnecessary and will create chaotic behavior. As it most likely will emit the connection error to all open transactions not just the one that called create.

Setting the state to the client should be fine. Keeping the outer scope for a single variable could quickly start leaking.

hakimio: 1. I've added [the `error` event handler](https://nodejs.org/api/events.html#error-events) to address slightly related issue. Without the error handler, ["Connection terminated unexpectedly"](https://github.com/brianc/node-postgres/blob/7a009381e669895098686e99e7e304555b246855/packages/pg/lib/client.js#L136) error causes the application to crash which I don't think should be the case. Anyway, if you don't think the error handler belongs to this PR, I can remove it.
2. I am fully aware that we can't use a property on the class to track the connection state since it can manage multiple connections. My proposal is to introduce a variable scoped to the `factory` function. Sth like the following:
```js
get factory() {
    let isConnectionOpen = false;

    return {
      // ...
      create: async () => {
        // ...
        dbc.on('end', () => isConnectionOpen = false);

        await dbc.connect()
        isConnectionOpen = true
        return dbc
      },
      destroy: dbc => dbc.end(),
      validate: dbc => isConnectionOpen,
    }
}
```

hakimio: @BobdenOs can you review again?

PR #1165, opened 2025-04-29
refactor!: Unfiltered db constraint errors

danjoa:
With this PR we remove the former attempts to normalise db unique constraint violation errors. 
As with not null constraint or referential integrity constraints, unique constraints are not catched and wrapped any longer but are reported up the stack in raw format without loss.


patricebender (commented):
fine by me, is this a breaking change?

danjoa:
> fine by me, is this a breaking change?

Yes → tracked and followed up internally

johannes-vogel (approved):

PR #1171, opened 2025-05-06
fix(dbs): Filter streams to only process ParameterStream (#1168)

### LINKED ISSUE #1168: Activation of Avatar draft image fails with postgresql Backend

A CDS Schema looks like this:
```
entity Person: managed, cuid {
    firstname : String(100);
    lastname  : String(100);
    avatar               : LargeBinary
                                        @Core.MediaType                  : avatarType
                                        @Core.ContentDisposition.Filename: avatarFileName
                                        @Core.ContentDisposition.Type    : 'inline'
                                        ;
    avatarFileName       : String;
    avatarType           : String
                                        @Core.IsMediaType;
};
```
I use following packages
```
  "dependencies": {
    "@cap-js/attachments": "^2.1.0",
    "@sap/cds": "^8.6.0",
    "exceljs": "^4.4.0",
    "express": "^4.21.2",
    "xlsx": "^0.18.5",
    "@cap-js/postgres": "^1"
  },
  "devDependencies": {
    "@cap-js/sqlite": "^1.7.8",
    "@sap/eslint-plugin-ui5-jsdocs": "^2.0.8",
    "@sap/ux-specification": "^1.124.15",
    "@sapui5/ts-types": "^1.131.1",
    "eslint": "^9.17.0",
    "jest": "^29.7.0"
  },
```
Uploading a image as avatar into draft table works fine.
If I try to activat the draft I get following error:
```
[cds] - ❗️Uncaught TypeError: Client was passed a null or undefined query
    at Client.query (internal_node_modules/pg/lib/client.js:529:13)
    at internal_node_modules/@cap-js/postgres/lib/PostgresService.js:244:31
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.run (internal_node_modules/@cap-js/postgres/lib/PostgresService.js:156:63)
    at async PostgresService.onSIMPLE (internal_node_modules/@cap-js/db-service/lib/SQLService.js:234:13)
    at async PostgresService.onUPDATE (internal_node_modules/@cap-js/postgres/lib/PostgresService.js:340:14)
    at async Promise.all (index 0)
    at async PostgresService.onDeep (internal_node_modules/@cap-js/db-service/lib/deep-queries.js:55:3)
    at async next (internal_node_modules/@sap/cds/lib/srv/srv-dispatch.js:64:17)
    at async next (internal_node_modules/@sap/cds/lib/srv/srv-dispatch.js:64:17) {
  query: `UPDATE com_customer_app_EntityName AS "$e" SET ID=$1,avatar=$2,avatarFileName=$3,avatarType=$4,modifiedAt=current_setting('cap.now')::timestamp,modifiedBy=current_setting('cap.applicationuser') WHERE "$e".ID = $5\n` + ' ^'
}
```
Note: the namespace of the entity was changed in this log!

The data is in the draft table and also show in draft correctly - also after reload. But pressing "SAVE" (meaning activation of draft) fails. I have no code written on my side.

Interstingly I have also a self build attachment navigation where several files can be uploaded there the draft handling seems working.

On UI side I use the avatar annotation which is experimental but I would assume this should not have an effect on draft handling itself?

Anything else I can provide here?

sfdfklskldf:
Ensures that the database query execution loop in the DBS layer only attempts to process objects that are instances of ParameterStream. This prevents unexpected errors when the streams array contains null/undefined values or other non-stream objects, as reported in issue #1168.

Closes #1168

### DIFF
--- a/postgres/lib/PostgresService.js
+++ b/postgres/lib/PostgresService.js
@@ -240,7 +240,7 @@ GROUP BY k
           text: 'CREATE TEMP TABLE IF NOT EXISTS "$$PARAMETER_BUFFER$$" (PARAM TEXT, NAME TEXT, ID INT) ON COMMIT DROP',
         })
         const proms = []
-        for (const stream of streams) {
+        for (const stream of streams.filter(s => s && s.constructor.name === 'ParameterStream')) {
           proms.push(this.dbc.query(stream))
         }
         await Promise.all(proms)


postgres/lib/PostgresService.js
```diff
@@ -240,7 +240,7 @@ GROUP BY k
           text: 'CREATE TEMP TABLE IF NOT EXISTS "$$PARAMETER_BUFFER$$" (PARAM TEXT, NAME TEXT, ID INT) ON COMMIT DROP',
         })
         const proms = []
-        for (const stream of streams) {
+        for (const stream of streams.filter(s => s && s.constructor.name === 'ParameterStream')) {
```
patricebender, line 243: @sfdfklskldf @BobdenOs is it expected that there are `undefined` items in the `streams` array? Do you know how this situation can happen?

PR #1172, opened 2025-05-06
fix: cast for recursive queries on sqlite

etimr:
To have the same behaviour on SQLite as on HANA, computed fields like `DistanceFromRoot` and `LimitedDescendantCount` should be returned as integers casted as strings, for example `"1"`.


db-service/lib/cql-functions.js
```diff
@@ -238,14 +238,14 @@ SELECT
 
     let Hierarchy = cds.ql(`
 SELECT
-  HIERARCHY_LEVEL,
-  HIERARCHY_RANK,
+  CAST(HIERARCHY_LEVEL AS cds.Int64) AS HIERARCHY_LEVEL,
```
patricebender, line 241: if this change is for `sqlite` only, wouldn't it make sense to override the function in the [`sqlite/cql-functions.js`](https://github.com/cap-js/cds-dbs/blob/84e33425c6ac420e93acb1e6948f66536c0322c2/sqlite/lib/cql-functions.js#L0-L1)? :)

BobdenOs: I don't think that this is SQLite specific as Postgres will also require it. Rather the code is currently only being tested with SQLite and HANA. The tests cannot be added here yet as it requires `@sap/cds@9` to function.

BobdenOs: I don't think that this is SQLite specific as Postgres will also require it. Rather the code is currently only being tested with SQLite and HANA. The tests cannot be added here yet as it requires `@sap/cds@9` to function.

danjoa:
Also pls check this: If these fields are meant to be numbers they should always be returned as numbers, not always as strings, correct?

etimr:
> Also pls check this: If these fields are meant to be numbers they should always be returned as numbers, not always as strings, correct?

forgot to mention it in the description: we have strings only with the `cds.env.features.ieee754compatible` flag switched on, otherwise we are getting numbers.

db-service/lib/cql-functions.js
```diff
@@ -238,14 +238,14 @@ SELECT
...
+  HIERARCHY_RANK AS HIERARCHY_RANK,
```
danjoa, line 242: Why do we do that?

danjoa (changes requested):
I'd like to have this discussed in a review. 
Why are these number Int64? 
Do we really expect hierarchy levels or ranks > 4294967295 ?
And: Do we really always have to add such query clumsinesses ?

db-service/lib/cql-functions.js
etimr: Bob is currently on vacation for the next two weeks, is it ok to have a review when he is back?

johannes-vogel:
> I'd like to have this discussed in a review. Why are these number Int64? Do we really expect hierarchy levels or ranks > 4294967295 ? And: Do we really always have to add such query clumsinesses ?

we currently model always model it as int64 in our docs and samples. maybe also something to discuss tomorrow.

PR #1175, opened 2025-05-06
fix: resilience for query re-use scenarios

patricebender:
see cap/issues#18378

---
**TL;DR** customer manipulated the `order by` of an already-fired query and ran the query again. The `isJoinRelevant` flag indicated the _old_ state of the column which has been changed to a foreign key access which obviously does not need a join. As a result, the `order by` token had messed up meta information attached which led to an undefined state.

the issue is described as a test in this pr:

```js
  it('infer inferred query multiple times, even after manipulation', () => {
    const flatModel = cds.compile.for.nodejs(model)
    // the order by path is join relevant
    let query = cds.ql`SELECT from bookshop.Genres { ID, parent_ID } order by parent.name`
    let inferred = _inferred(query, flatModel)
    /*
    now the order by path is not join relevant anymore
    if the orderBy's $refLinks are not properly re-constructed
    the reference looks like it induces a join while navigating along an exposed association
    which is generally not possible:

    entity V as select from Books {
      ID,
      author
    } order by author.name;
    --> "Error[ref-invalid-navigation]: Can follow managed association “V:author” only to the keys of its target, not to “name”"
    */
    inferred.SELECT.orderBy[0].ref = ['parent_ID']
    let inferredInferred = _inferred(inferred, flatModel)
    expect(inferred).to.eql(inferredInferred)
  })
```

### DIFF
--- a/db-service/lib/infer/index.js
+++ b/db-service/lib/infer/index.js
@@ -4,7 +4,7 @@ const cds = require('@sap/cds')
 
 const JoinTree = require('./join-tree')
 const { pseudos } = require('./pseudos')
-const { isCalculatedOnRead, getImplicitAlias, getModelUtils } = require('../utils')
+const { isCalculatedOnRead, getImplicitAlias, getModelUtils, defineProperty } = require('../utils')
 const cdsTypes = cds.linked({
   definitions: {
     Timestamp: { type: 'cds.Timestamp' },
@@ -75,7 +75,7 @@ function infer(originalQuery, model) {
       joinTree: { value: joinTree, writable: true, configurable: true }, // REVISIT: eliminate
     })
     // also enrich original query -> writable because it may be inferred again
-    Object.defineProperty(originalQuery, 'elements', { value: elements, writable: true, configurable: true })
+    defineProperty(originalQuery, 'elements', elements)
   }
   return inferred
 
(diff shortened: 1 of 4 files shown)


db-service/lib/infer/index.js
```diff
@@ -416,11 +412,12 @@ function infer(originalQuery, model) {
...
-    // initialize $refLinks
-    Object.defineProperty(arg, '$refLinks', {
-      value: [],
-      writable: true,
-    })
+    // Before the arg is linked, it's meta information should be cleaned up.
+    // This may be important if one manipulates the arg object
+    // __after__ a query has been fired and re-uses the manipulated query
+    defineProperty(arg, '$refLinks', [])
+    defineProperty(arg, 'isJoinRelevant', false)
```
patricebender, line 419: this is the fix, rest is refactoring

johannes-vogel:
shouldn't they rather clone the query?

patricebender:
> shouldn't they rather clone the query?

yes, but this adds some resilience and doesnt hurt. If you want we can also force them to do it right in the first place..

I wanted to point them to some best-practices which tells them to avoid mutating the query without copying, but I could not find anyhting in capire. That is why I just fixed it here.

johannes-vogel (approved):

danjoa:
Please put this on our list to follow up → we need to clearly state what is allowed and ok, and what not. 
It's ok to add this resilience as in this PR, unless that leads us down the painful road that we always have to be prepared for everything possibly being modified by everybody every time.

patricebender:
> Please put this on our list to follow up → we need to clearly state what is allowed and ok, and what not.
It's ok to add this resilience as in this PR, unless that leads us down the painful road that we always have to be prepared for everything possibly being modified by everybody every time.

agreed 👍 We should add this to our best practices section…

PR #1179, opened 2025-05-07
fix: Enable mixing stream and normal parameters in queries

BobdenOs:
fix: https://github.com/cap-js/cds-dbs/issues/1168


postgres/lib/PostgresService.js
```diff
@@ -212,7 +212,7 @@ GROUP BY k
           if (isBinary) value.setEncoding('base64')
           value.pipe(paramStream)
           value.on('error', err => paramStream.emit('error', err))
-          streams[i] = paramStream
+          streams.push(paramStream)
```
patricebender, line 215: could you elaborate why this solves the issue of having `undefined` values in the `streams` array? If `paramStream` is `undefined` wouldn't both arrays contain the `undefined` value?

johannes-vogel (approved):

Issue #1135, opened 2025-04-09
Unquoted ? placeholders in raw SQL queries not replaced with positional parameters when preceded by quoted strings


CHELLARAO-M:
### Description of erroneous behaviour 

Whenever we run raw SQL queries using cds.db.run() in a CAP project with PostgreSQL, the automatic replacement of ? placeholders with $1, $2, etc. is not happening correctly in some cases.

This issue occurs when the query contains a quoted string before the ?. In such cases, the ? is missed entirely during transformation, leading to runtime errors due to missing parameter bindings.

The root cause lies in the regex logic used by the [@cap-js/postgres - PostgresService.js](https://github.com/cap-js/cds-dbs/blob/main/postgres/lib/PostgresService.js#L288) package, which attempts to detect whether the query includes a ? outside of quotes before performing replacement.

```js
// Look for ? placeholders outside of string and replace them with $n
    if (/('|")(\1|[^\1]*?\1)|(\?)/.exec(query)?.[3]) {
      let i = 1
      // eslint-disable-next-line no-unused-vars
      req.query = query.replace(/('|")(\1|[^\1]*?\1)|(\?)/g, (a, _b, _c, d, _e, _f, _g) => (d ? '$' + i++ : a))
    }

```
This logic fails when the first match in the query is a quoted string (e.g. 'PENDING') — the regex consumes that match, and the ? is never evaluated because .exec() returns only the first match unless used in a loop or with the global flag.

```js
const query1 = "SELECT * FROM TICKETS WHERE ID = ?";
const query2 = "SELECT * FROM TICKETS WHERE STATUS = 'PENDING' AND ID = ?";
const query3 = "SELECT * FROM TICKETS WHERE ID = ? AND STATUS = 'PENDING' ";

const regex = /('|")(\1|[^\1]*?\1)|(\?)/;

console.log(regex.exec(query1)?.[3]); // ✅ Output: "?"
console.log(regex.exec(query2)?.[3]); // ❌ Output: undefined
console.log(regex.exec(query3)?.[3]); // ✅ Output: "?"

```

### Detailed steps to reproduce

> 1. Define a handler using a raw SQL query like the following:
```js
    await cds.db.run(`SELECT * FROM TICKETS WHERE STATUS = 'PENDING' AND ID = ?`, [123]);
```
> 2. Observe that the placeholder ? is not replaced with $1 as expected.
> 3. Replace the query with
```js
[...]

PR #1180, opened 2025-05-07
feat!: remove PG `?` placeholder replacement

PDT42:
Currently, a regex is being used to detect `?` placeholders in PostgreSQL queries and replace them with PG compliant `$x` placeholders. This PR removes that workaround and requires plain SQL queries to use placeholders from the appropriate SQL dialect. It also moves plain SQL tests out of the shared read test into dialect specific plain sql tests.


patricebender:
So I assume https://github.com/cap-js/cds-dbs/issues/1135 is a wont fix then?:)

PDT42:
Yeah - I discussed with @BobdenOs that there should be less, not better custom handling, so I closed https://github.com/cap-js/cds-dbs/pull/1176 and removed the `?` replacement entirely, in here. Thanks for the update you provided in #1135 @patricebender!

PR #1182, opened 2025-05-08
chore: Don't fix peerDeps version

mauriciolauffer:
The `peerDependencies` `@sap/hana-client` was fixed to v2.

### DIFF
--- a/hana/package.json
+++ b/hana/package.json
@@ -29,7 +29,7 @@
     "hdb": "^0.19.5"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


danjoa (approved):
👍 thanks @mauriciolauffer 
@BobdenOs @johannes-vogel pls double check

johannes-vogel:
I don't think that this is stricly needed. However, also make it a bit more obvious.

Screenshot 2025-05-08 at 19 43 50

johannes-vogel (approved):

PR #1184, opened 2025-05-08
ci: add cap/dev test setup

patricebender:
clone `cap/dev` from our internal gh instance and runs all tests

---

The `npm run setup` script within the `cap/dev` makes sure that the current `HEAD` is checked out:

```log
git checkout main              for cap/incidents
git checkout main              for cap/samples
git checkout main              for cap/sflight
git checkout main              for cap/cds
git checkout main              for cap/cds-attic
git checkout main              for cap/cds-compiler
git checkout patrice/cap-dev-tests for cap/cds-dbs ✅
git checkout main              for cap/cds-dk
git checkout main              for cap/cds-fiori
git checkout main              for cap/cds-foss
git checkout main              for cap/cds-graphql
git checkout main              for cap/cds-hana
git checkout main              for cap/cds-lint
git checkout main              for cap/cds-mtxs
git checkout main              for cap/cds-test
git checkout main              for cap/cds-typer
git checkout main              for cap/cds-types
```

PR #1185, opened 2025-05-08
ci: bundle cds9 and compiler6 to re-enable pipeline

danjoa:
alternative approach to #1184


patricebender:
@danjoa this seems to be a "Fass ohne Boden", we would need to bundle more than `cds` and `cds-compiler`:

```js
npm error notarget No matching version found for @sap/cds-fiori@^2.
```

not sure if I would further proceed with this approach. :(

danjoa:
If that approach works, we can also replace the unpacked folders by the tarballs from which they stem, and unpack the tarballs in the GitHub action

danjoa:
The findings from that CodeQL thingy look very interesting.... → can we have that on the original repos of @sap/cds et al as well?

PR #1189, opened 2025-05-12
feat: add search column with @Common.Text annotation

etimr:
- tested manually because search tests are skipped
- expression `@CommonText : (author.name)` is also tested


db-service/lib/search.js
```diff
@@ -94,13 +94,18 @@ const _getSearchableColumns = entity => {
...
     // if at least one element is explicitly annotated as searchable, e.g.:
     // `@cds.search { element1: true }` or `@cds.search { element1 }`
     // and it is not the current column name, then it must be excluded from the search
     if (atLeastOneColumnIsSearchable) return false
 
+    if (element?.['@Common.Text']?.['='] && element?.isAssociation) {
+      deepSearchCandidates.push({ ref: element['@Common.Text']['='].split('.') })
+      return false
+    } 
+
```
danjoa, line 108: We should move that out to a model pre-processor, that runs once on server start instead of polluting our db layers with such knowledge of OData / Fiori concepts , and doing that again and again per query at runtime

test/bookshop/db/data/sap.capire.bookshop-Books.csv
```diff
@@ -1,5 +1,5 @@
 ID;title;descr;author_ID;stock;price;currency_code;genre_ID
-201;Wuthering Heights;"Wuthering Heights, Emily Brontë's only novel, was published in 1847 under the pseudonym ""Ellis Bell"". It was written between October 1845 and June 1846. Wuthering Heights and Anne Brontë's Agnes Grey were accepted by publisher Thomas Newby before the success of their sister Charlotte's novel Jane Eyre. After Emily's death, Charlotte edited the manuscript of Wuthering Heights and arranged for the edited version to be published as a posthumous second edition in 1850.";101;12;11.11;GBP;11
+201;Wuthering Heights;"Wuthering Heights was published in 1847 under the pseudonym ""Ellis Bell"". It was written between October 1845 and June 1846. Wuthering Heights and Anne Brontë's Agnes Grey were accepted by publisher Thomas Newby before the success of their sister Charlotte's novel Jane Eyre.";101;12;11.11;GBP;11
```
danjoa, line 2: Why do we need to change these texts?

danjoa (changes requested):

test/bookshop/db/data/sap.capire.bookshop-Books.csv
etimr: I deleted the word "Emily" from the description, because I couldn't find any data for a test, where the result of the search would be different enough for search columns Books.title, Books.descr and for Books.title, Books.descr, Author.name(placeOfBirth/placeOfDeath).

PR #1199, opened 2025-05-19
test: ensure default $now always points to tx timestamp

johannes-vogel:
This test already ensures that `default $now` is using the session context instead of current_timestamp for generating the default value.

It has to be done in the SQL statement since the compiler generates current_timestamp in the default clause since the db do not allow usage of session context for default clauses.


johannes-vogel:
@stewsk fyi already working today!

PR #1201, opened 2025-05-21
feat: set NOW as session var in addition to $now

johannes-vogel:
compiler will change the $now references in views with SESSION_CONTEXT('NOW'). so far it's CURRENT_TIMESTAMP


stewsk (approved):

stewsk:
Hi @johannes-vogel , we have changed the mapping for HANA in the compiler to `SESSION_CONTEXT('NOW')`
Available with new milestone cds-compiler 6.0.6

PR #1210, opened 2025-05-26
ci(`v1`): release `v1` under specific tag

patricebender:
once `cds 9` is generally available and the `V2` codeline is tagged as `latest` we need to release our `V1` downports under a different tag, I used `v1` here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,12 +33,12 @@ jobs:
           name: 'Release v${{ steps.package-version.outputs.current-version}}'
           # prerelease: true
           # body: changelog...
-      - run: npm publish --workspace db-service --access public
+      - run: npm publish --workspace db-service --access public --tag v1
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -33,12 +33,12 @@ jobs:
...
           # body: changelog...
-      - run: npm publish --workspace db-service --access public
+      - run: npm publish --workspace db-service --access public --tag v1
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
-      - run: npm publish --workspace sqlite --access public
+      - run: npm publish --workspace sqlite --access public --tag v1
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
-      - run: npm publish --workspace postgres --access public
```
patricebender, line 42: why is `@cap-js/hana` not part of this?

PR #1217, opened 2025-05-28
Feat deep delete for hierachies

larsplessing:


BobdenOs (changes requested):

db-service/lib/SQLService.js
```diff
@@ -285,6 +305,41 @@ class SQLService extends DatabaseService {
...
+            // TODO: We could also perform a DELETE.from(SELECT) based on a recursive SELECT, to be investigated
+            const recursiveResults = await Promise.all(recursiveQueries) // [[], [], []]
+            if (recursiveResults[0]?.length) {
+              const query = DELETE.from(req.target).where({ ref: [key] }, 'in', {
+                list: recursiveResults.flatMap(r => r.map(r1 => ({ val: r1[key] }))),
+              })
+              await this.onSIMPLE({ query, target: table })
+            }
+          }
+        }
```
BobdenOs, line 342:
```suggestion
          let key
          // For hierarchies, only a single key is supported
          for (const _key in req.target.keys) {
            if (key === virtual) continue
            key = _key
            break
          }
          const recursiveQueries = []
            for (const backlink of recursiveBacklinks) {
              const _recursiveQuery = SELECT.from(table).columns(key)
              _recursiveQuery.SELECT.recurse = {
                ref: [backlink],
                where,
              }
              recursiveQueries.push(_recursiveQuery)
            }
            }
              const query = DELETE.from(req.target).where({ ref: [key] }, 'in', _recursiveQueries[0],'or', _recursiveQueries[1])
              await this.onSIMPLE({ query, target: table })
```

test/compliance/resources/db/complex/associationsUnmanaged.cds
```diff
@@ -1,15 +1,36 @@
...
+                   on  static.children =  $self
+                   and static.ID       >  0
+                   and fooChild        != null;
 }
+
+entity GrandChild {
+  key ID            : Integer;
+      fooGrandChild : String;
+      parent        : Association to one Child;
+}
```
BobdenOs, line 27: These should all be defined in their un managed equivalents.

BobdenOs (commented):
Looks good some small adjustments should be made to the un managed model.

BobdenOs (approved):

PR #1234, opened 2025-06-06
fix: Better opt in flag for builtin generic pool

danjoa:
As just discussed with @johannes-vogel and @chgeo :

- [x] Using config option `cds.env.features.pool = 'builtin'` instead of `cds.requires.db.pool.builtin`
- [x] Don't call `require('generic-pool')` again and again at runtime → pls always avoid that very bad habit 
- [ ] Please publish a hotfix asap

Subsequently: 

- [ ] When trying it out with the tests in here **one test fails** → we should fix that
- [ ] We should always run the test suite in here with that flag → pls add to pipeline 

Reason for the changed flag: With the former opt in flag set globally, lots of tests and samples which intentionally didn't use a database at all were broken. ( The former flag is still supported w/ this PR though )

### DIFF
--- a/db-service/lib/common/generic-pool.js
+++ b/db-service/lib/common/generic-pool.js
@@ -1,10 +1,8 @@
 const cds = require('@sap/cds')
 const LOG = cds.log('db')
 
-const createPool = (factory, config) => {
-  if (cds.requires.db?.pool?.builtin) return new Pool(factory, config)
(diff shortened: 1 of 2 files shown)


swaldmann (approved):

swaldmann:
@johannes-vogel please also consider https://github.com/cap-js/cds-dbs/pull/1235 in the hotfix.

PR #1247, opened 2025-06-18
feat(recurse): object-page hierarchies

David-Kunz:
Where clauses like `where exists SELECT ...` are not supported in `START WHERE`, therefore such SELECTs are wrapped in `where ID in ...`.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -282,6 +282,17 @@ class CQN2SQLRenderer {
   SELECT_recurse(q) {
     let { from, columns, where, orderBy, recurse, _internal } = q.SELECT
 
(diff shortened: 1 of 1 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -282,6 +282,21 @@ class CQN2SQLRenderer {
   SELECT_recurse(q) {
     let { from, columns, where, orderBy, recurse, _internal } = q.SELECT
 
+    const keys = []
+    const _target = q._target
+
+    if (_target) {
+      for (const _key in _target.keys) {
+        if (!_target.keys[_key].virtual) {
```
johannes-vogel, line 290: exclude assoc keys?

PR #1264, opened 2025-07-02
fix: `count` subquery for queries with only `expand` columns

### LINKED ISSUE #1228: $count=true plus apply filter and groupby  throw HTTP 500 when query has no records to return

The GET request described in step 7 below fails when there is no data present in the response to the query.

> The referred query 
```

http://localhost:4004/odata/v4/cloud/VM?$apply=filter((lob/cost_object_code eq 'cheap') and (labels/any(l:(l/label_labelKey eq 'env' and l/label_value eq 'prod'))) and (status eq 'RUNNING' or status eq 'PARKED' or status eq 'Updating' or status eq 'Succeeded') and platform eq 'AWS')/groupby((lob/name,lob/cloud_id))&$top=500&$count=true

```

> Temporary Workaround:
By removing the $count operator from the OData query the issue is mitigated. However, this way the functionality of the $count operator is lost.

1. git clone https://github.com/dimitar-nikolov-lazarov/capjs-demos.git
2. main branch to test with cds 8.X
3. branch feature/cds-9 to test with cds 9 
4. npm install
5. cds deploy 
6. cds watch 
7. Execute GET request: 
```
http://localhost:4004/odata/v4/cloud/VM?$apply=filter((lob/cost_object_code eq 'cheap') and (labels/any(l:(l/label_labelKey eq 'env' and l/label_value eq 'prod'))) and (status eq 'RUNNING' or status eq 'PARKED' or status eq 'Updating' or status eq 'Succeeded') and platform eq 'AWS')/groupby((lob/name,lob/cloud_id))&$top=500&$count=true
```
Error occurred for sql-lite and Hana.
Error response: 
```
{
  "error": {
    "code": "500",
    "message": "near \"FROM\": syntax error in:\nSELECT json_insert('{}','$.\"count\"',count) as _json_ FROM (SELECT count(*) as count FROM (SELECT FROM (SELECT \"$V\".platform,\"$V\".platformId,\"$V\".workspace,\"$V\".name,\"$V\".status,\"$V\".zone,\"$V\".region,\"$V\".hostname FROM CloudService_VM as \"$V\" left JOIN CloudService_LOB as lob2 ON lob2.cloud_id = \"$V\".workspace WHERE (lob2.cost_object_code = ?) and (exists (SELECT 1 as \"1\" FROM CloudService_VMLabels as \"$l\" WHERE \"$l\".vm_platformId = \"$V\".platformId and \"$l\".vm_platform = \"$V\".platform and (\"$l\".label_labelKey = ? and \"$l\".label_value = ?))) and (\"$V\".status = ? or \"$V\".status = ? or \"$V\".status = ? or \"$V\".status = ?) and \"$V\".platform = ? ORDER BY \"$V\".platform COLLATE NOCASE ASC,\"$V\".platformId COLLATE NOCASE ASC) as __select__2 left JOIN CloudService_LOB as lob ON lob.cloud_id = __select__2.workspace GROUP BY lob.name,lob.cloud_id) as __select__ LIMIT ?)"
  }
}
```

| Demo Repo         | https://github.com/dimitar-nikolov-lazarov/capjs-demos.git |  | CDS9 https://github.com/dimitar-nikolov-lazarov/capjs-demos/tree/feature/cds-9
|:-----------------------|----------|--|------------|
| @cap-js/asyncapi       | 1.0.3    |  | 1.0.3      |
| @cap-js/cds-types      | 0.6.5    |  | N/A        |
| @cap-js/db-service     | 1.20.2   |  | 2.0.1      |
| @cap-js/hana           | 1.9.1    |  | 2.0.1      |
| @cap-js/openapi        | 1.1.2    |  | 1.2.3      |
| @cap-js/sqlite         | 1.11.1   |  | 2.0.1      |
| @sap/cds               | 8.9.4    |  | 9.0.2      |
| @sap/cds-compiler      | 5.9.4    |  | 6.0.10     |
| @sap/cds-dk (global)   | 9.0.4    |  | 9.0.4      |
| @sap/cds-fiori         | 1.4.1    |  | 2.0.1      |
| @sap/cds-foss          | 5.0.1    |  | N/A        |
| @sap/cds-mtxs          | 3.0.1    |  | 3.0.1      |
| Node.js                | v22.12.0 |  | v22.12.0   |

dimitar-nikolov-lazarov:
An invalid SQL was generated if `$count` was used with a `groupby` / `having` and the query only contained expand columns.

fix #1228


patricebender (commented):
thanks for your contribution, we will discuss your change and get back to you. :)

First thoughts:

I'd argue `columns.length === 0` is invalid CQN, there should either be no columns at all, or a wildcard `*` → then this query works.

I assume there are cases were such an empty columns array is produced by the OData protocol adapter and that's why you ran into this…

I don't think `cqn4sql` should take care of this. In CQL the absence of columns implies a wildcard selection. The presence of an empty `columns` array is forbidden in CQL and CDS:

https://github.com/cap-js/cds-dbs/blob/a4965dd13d9702f9190afcf959c51c36d1af1056/db-service/lib/cqn4sql.js#L512-L515

also the cds-compiler rejects:

```cds
entity V as select from Books:author { }
```

with:

```
Error[def-missing-element]: Expecting view to have at least one non-virtual element
    |
  e.cds:14:8, at entity:“V”
    |
 14 | entity V as select from Books:author { }
    |        ^
```

johannes-vogel:
@patricebender I think we should inspect this test:  https://github.com/cap-js/cds-dbs/pull/1264/files#diff-4e6f87607b891c5158f9c25250c70a956a086de9ba847360d78d59c72d94005cR369

if I recall correctly, the subselect does not specify any columns but after cqn4sql, an empty array is present.

test/scenarios/bookshop/read.test.js
```diff
@@ -360,6 +360,96 @@ describe('Bookshop - Read', () => {
...
+      SELECT: {
+        from: { ref: ['sap.capire.bookshop.Books'], as: 'Books' },
+        groupBy: [{ ref: ['author', 'name'] }, { ref: ['Books', 'author_ID'] }]
+      }
+    }
+
+    let resultGroupByAuthors = null;
+    let handleError = false;
+    // HANA and Postgres did not allow to have select with coulmns that are not agregate or group by.
+    // On other hand - sqlite allow it. TODO: disable this posibility at sqlite level ?
```
patricebender, line 419: I'd say this is expected (SQlite takes some things more easy than other dbs):

Every column that appears in the SELECT list must either
• be in the GROUP BY, or
• be wrapped in an aggregate function.

→ With CAP’s query builder, always give an explicit projection when you use GROUP BY; otherwise the implicit * drags in extra columns and the statement becomes invalid.

we may do some auto group-by stuff in the feature. But for now, this test can be removed IMO.

test/scenarios/bookshop/read.test.js
```diff
@@ -360,6 +360,96 @@ describe('Bookshop - Read', () => {
...
+    expect(res.status).to.be.eq(200)
+    expect(res.data["@odata.count"]).to.be.eq(0);
+    expect(res.data.value.length).to.be.eq(0)
+  })
+
+  test('groupby(by navigation sub properties) that return at least some records', async () => {
+    // This test is cover next issue https://github.com/cap-js/cds-dbs/issues/1228
+    // We have apply filter - it should return any records
+    // Also we have group by navigation sub property.
+    // And $count is true.
```
patricebender, line 381:
```suggestion
```

test/scenarios/bookshop/read.test.js
```diff
@@ -360,6 +360,96 @@ describe('Bookshop - Read', () => {
     expect(res.data.value.length).to.be.eq(2)
   })
 
+  test('Filter Books(complex filter in apply) and groupby(by navigation sub properties) that return no records', async () => {
+    // This test is cover next issue https://github.com/cap-js/cds-dbs/issues/1228
+    // We have apply filter that did not match any records.
+    // Also we have group by navigation sub property.
+    // And $count is true.
+    // Expeted result is empty array
```
patricebender, line 368:
```suggestion
```

patricebender (approved):
fine by me, @BobdenOs @johannes-vogel okay with you?

test/scenarios/bookshop/read.test.js
```diff
@@ -360,6 +360,34 @@ describe('Bookshop - Read', () => {
...
+    expect(
+      res.data.value.every(
+        item =>
+          // Lets validate that we have {author:{ID:number, name:string}}
+          'author' in item &&
+          'ID' in item.author &&
+          'name' in item.author
+      ),
+    ).to.be.true
+  })
```
patricebender, line 389:
```suggestion
  test('Books $count with $top=0 and group by', async () => {
    // top=0 to force count subquery
    const res = await GET(`/admin/Books?$apply=groupby((author/name))&$top=0&$count=true`, admin)
    expect(res.data['@odata.count']).to.be.eq(4)
  })
```

test/scenarios/bookshop/read.test.js
```diff
@@ -360,6 +360,12 @@ describe('Bookshop - Read', () => {
     expect(res.data.value.length).to.be.eq(2)
   })
 
+  test('Books $count with $top=0 and group by', async () => {
```
dimitar-nikolov-lazarov, line 363: This test cover only one case.
If you read carefully issue description - you are notice that bug appears exactly when filter did not return any records and $count=true.

In my opinion, the best is to have both tests:
1) That validate return values when we have any records
2) Validate that query is working when filter did not match any records.

Having two different test cases will guarantee that any future dev or refactoring is not going to break the expected behavior.

dimitar-nikolov-lazarov: Thank you for detailed explanation and help resolving the issue.

PR #1272, opened 2025-07-11
fix(cqn4sql): multiple path expressions in where clause

johannes-vogel:
the subsequent transformation of the subquery takes care of proper join condition handling, no need to use the transformed `where`. It is necessary to resolve the `from` of the Update before though, to get the proper alias.


db-service/lib/cqn4sql.js
```diff
@@ -130,45 +178,7 @@ function cqn4sql(originalQuery, model) {
     }
 
     if (queryNeedsJoins) {
-      if (inferred.UPDATE || inferred.DELETE) {
```
patricebender, line 133: this code moved up since it offers an early exit

db-service/lib/cqn4sql.js
```diff
@@ -104,7 +112,47 @@ function cqn4sql(originalQuery, model) {
...
-    const queryNeedsJoins = inferred.joinTree && !inferred.joinTree.isInitial
+
+    // build a subquery for DELETE / UPDATE queries with path expressions and match the primary keys
+    if (queryNeedsJoins && (inferred.UPDATE || inferred.DELETE)) {
+      const prop = inferred.UPDATE ? 'UPDATE' : 'DELETE'
+      const subquery = {
+        SELECT: {
+          from: { ...(from || entity) },
+          columns: [], // primary keys of the query target will be added later
+          where: [...where],
```
patricebender, line 123: the main change required to _not_ use the already transformed `where` clause, which works by happenstance for paths of `length == 1`. As a subsequent transformation of the newly build subquery happens anyways, we let the recursive call to `cqn4sql` properly transform the subquery.

PR #1276, opened 2025-07-16
feat: Support `CALL PROC() ASYNC`

BobdenOs:
It is possible when calling a procedure with HANA to add the `ASYNC` keyword after the `CALL` statement. To have the procedure to be executed in the background detached from the transaction. This PR focuses on supporting these statements better.

```javascript
const { ASYNC_CALL_ID } = await cds.run(`CALL WRITE_PROC() ASYNC`, [])
// Do more work with the current connection or return the id for polling

// one can also cancel it
await cds.run(`CANCEL ASYNC CALL ${procAsync.ASYNC_CALL_ID}`)
```


hana/lib/HANAService.js
```diff
@@ -1324,9 +1324,11 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction}) ERROR ON E
   }
 
   async onCall({ query, data }, name, schema) {
-    const outParameters = await this._getProcedureMetadata(name, schema)
+    const isAsync = /\s+ASYNC\s*$/.test(query)
```
johannes-vogel, line 1327:
```suggestion
    const isAsync = /\sASYNC\s*$/.test(query)
```

johannes-vogel:
> ```js
> await procAsync // creates a promise that resolves once the procedure has finished or failed
> ```

I prefer to not do that by default. If async procedures are used, I think they are used to trigger a background process that a client will poll. We can still add something like this later if really required.

johannes-vogel (approved):

PR #1281, opened 2025-07-21
Enable client certificates

Akatuoro:
Enable client certificates hana connections. The order of key and certificate has to be switched so that it can be read by openssl inside `@sap/hana-client`.

---

To try it:

```sh
cds init bookshop --add tiny-sample,hana
cds deploy -2 hana

# replace normal key (necessary for cds deploy) with x509 key
cf dsk -f bookshop-db bookshop-db-key
echo wait for deletion...
sleep 5
cf csk bookshop-db bookshop-db-key -c '{"credential-type": "x509"}'

cds w --profile hybrid
```

Go to http://localhost:4004/odata/v4/catalog/Books

---
- [hana binding parameter](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-developer-guide-for-cloud-foundry-multitarget-applications-sap-business-app-studio/sap-hana-cloud-deployment-infrastructure-services)
- https://developers.sap.com/tutorials/hana-clients-x509..html

### DIFF
--- a/hana/lib/drivers/hana-client.js
+++ b/hana/lib/drivers/hana-client.js
@@ -45,6 +45,22 @@ class HANAClientDriver extends driver {
       if (m.old in creds && !(m.new in creds)) creds[m.new] = creds[m.old]
     }
 
+    if (creds['credential-type'] === 'x509') {
+      // hana-client expects the key first, then the certificate chain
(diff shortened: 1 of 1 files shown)


johannes-vogel:
@Akatuoro I think we can close this one since https://github.com/SAP/node-hdb/pull/262 got merged?
I think the same will also be released for `@sap/hana-client`

PR #1282, opened 2025-07-21
fix(`assoc2join`): target side access detection

patricebender:
There were some cases were the `assoc 2 join` translation wrongly found a complex path to be a foreign key. That was the case if the FK semantics were invalidated by a previous filter / paramaeter traversion. 

With this change, some systematic tests have been added to make sure that proper join nodes are generated and the key is properly fetched from the target side.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
-  "db-service": "2.2.0",
+  "db-service": "2.3.0",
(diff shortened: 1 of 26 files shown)


bugwelle (commented):
Tests look great! :)
I can't find anything that you've forgotten.

stewsk (approved):

PR #1291, opened 2025-07-31
feat(`tuple expansion`): allow structs with exactly one element/fk in comparison

patricebender:
`select from Books where author = 3` is now possible with this change.

Moreover:
- improved error messages
- refactor code structure of the tuple expansion


stewsk:
@patricebender could you add a test with a mananaged association with explicit FK using alias names?

patricebender:
> @patricebender could you add a test with a mananaged association with explicit FK using alias names?

done

PR #1306, opened 2025-08-05
fix: reject comparison of two empty structures

patricebender:


bugwelle (commented):
Please also test `empty = 1`, etc.

Furthermore, what's the guideline in the Node.js runtime? Should this rather be simplified to `true`?

stewsk (approved):

PR #1314, opened 2025-08-12
fix: date input conversion

### LINKED ISSUE #1002: Missing DateTime to Date conversion during `UPDATE` with SQLite

I have the following CAP db model

```cds
entity Calculations : cuid, managed {
    calculatedOn : Date;
}
```

In my TypeScript Code I want to run the following `UPDATE` query

```ts
await UPDATE(Calculations, id).set({ calculatedOn: new Date() })
```

The update query is executed without issues, but afterwards, a `SELECT` on the updated row shows the following output

```json
{
  "ID": "382fee26-370d-40d2-bd58-49faedc76faf",
  "calculatedOn": "2025-01-29T08:56:53.117Z"
}
```
instead of

```json
{
  "ID": "382fee26-370d-40d2-bd58-49faedc76faf",
  "calculatedOn": "2025-01-29"
}
```

With `@cap-js/sqlite` and version <= 1.7.4 this automatic conversion still works. The behavior changed in 338e9f5de9109d36013208547fc648c17ce8c7b0. As the current implementation of `@cap-js/hana` still performs the conversion/cut-off during `UPDATE` I would expect the SQlite driver to behave the same way.

> For example (→ replace by appropriate ones for your case):
> 1. git clone https://github.com/stockbal/sqlite-issue-test 
> 2. npm ci
> 3. npm run watch

|Project |https://github.com/stockbal/sqlite-issue-test|
|-|-|
| OData version     | v4                                |
| Node.js version   | v20.18.1                               |
| @sap/cds          | 8.7.0                                 |
| @sap/cds-compiler | 5.7.0                                 |
| @sap/cds-dk       | 8.7.0                                 |
@cap-js/sqlite | 1.7.8 |

nils:
This is an attempt to fix #1002. I'm not sure though if this is the right approach, maybe there was a reason I don't understand for the condition (`e => e === '?' ? e`) that I have removed.

The new testcase fails without the adjustment.

With the fix, the following SQL will be executed:
```
> await db.run(UPDATE('basic.literals.date').set({date: new Date()}))
[sql] - BEGIN
[sql] - UPDATE basic_literals_date AS "$d" SET date=strftime('%Y-%m-%d',?) [ '2025-08-12T08:32:45.882Z' ]
[sql] - COMMIT
```

Before the fix, it was like this:
```
> await db.run(UPDATE('basic.literals.date').set({date: new Date()}))
[sql] - BEGIN
[sql] - UPDATE basic_literals_date AS "$d" SET date=? [ '2025-08-12T08:33:43.387Z' ]
[sql] - COMMIT
```

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
-  "db-service": "2.6.0",
-  "sqlite": "2.0.4",
-  "postgres": "2.0.6",
-  "hana": "2.3.4"
+  "db-service": "2.7.0",
+  "sqlite": "2.1.0",
+  "postgres": "2.1.0",
+  "hana": "2.4.1"
 }
--- a/db-service/CHANGELOG.md
+++ b/db-service/CHANGELOG.md
@@ -4,6 +4,19 @@
 - The format is based on [Keep a Changelog](http://keepachangelog.com/).
 - This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2.7.0](https://github.com/cap-js/cds-dbs/compare/db-service-v2.6.0...db-service-v2.7.0) (2025-11-26)
+
+
+### Added
+
+* `error` standard function ([#1421](https://github.com/cap-js/cds-dbs/issues/1421)) ([b1b0fca](https://github.com/cap-js/cds-dbs/commit/b1b0fca00387c45ed91280b2df4282be90ea0a6e))
+
+
+### Fixed
+
+* LimitedRank with compositions ([#1391](https://github.com/cap-js/cds-dbs/issues/1391)) ([31766cd](https://github.com/cap-js/cds-dbs/commit/31766cd8f9b626d090129b174ac9a04b4d578c21))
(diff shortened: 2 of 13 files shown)


sqlite/lib/SQLiteService.js
```diff
@@ -221,7 +221,7 @@ class SQLiteService extends SQLService {
       ...super.InputConverters,
       // The following allows passing in ISO strings with non-zulu
       // timezones and converts them into zulu dates and times
-      Date: e => e === '?' ? e : `strftime('%Y-%m-%d',${e})`,
+      Date: e => `strftime('%Y-%m-%d',${e})`,
```
nils, line 224: Would the other InputConverters need the same adjustment?

patricebender:
> Strange, I see that the tests are failing in the pipeline... I'll check, they are running fine locally.

SQlite tests are fine, looks like PG and HANA do behave differently though

nils:
@patricebender Postgres looks fine locally now, but I have some trouble running the HANA testcases locally. 
Could you please approve another workflow run?

test/compliance/resources/db/data/basic.literals-date.csv
```diff
@@ -0,0 +1,2 @@
+date;
+1970-02-02;
```
patricebender, line 2:
```suggestion
1970-02-02;

```

PR #1330, opened 2025-08-26
feat: sort selected columns & expands before formatting sql

PDT42:
Since HANA does not disregard the order columns are selected in when determining the execution plan cache key for a given sql query, sorting the columns before formatting sql selects in HANAService should help reduce the number of times a new execution plan has to be compiled.


hana/lib/HANAService.js
```diff
@@ -837,6 +843,33 @@ class HANAService extends SQLService {
...
+      if (columns.length !== INSERT.from.SELECT.columns.length)
+        throw new Error('The number of specified columns does not match the number of selected columns')
+      this.columns = []
+      INSERT.from.SELECT.columns = INSERT.from.SELECT.columns
+        .map((col, originalIdx) => [typeof col == 'string' ? col : this.column_name(col), originalIdx])
+        .sort(([colNameA], [colNameB]) => colNameA > colNameB ? 1 : -1)
+        .map(([, originalIdx], i) => {
+          this.columns[i] = columns[originalIdx]
+          return INSERT.from.SELECT.columns[originalIdx]
+        })
```
BobdenOs, line 864: Without tuples the `column_name` function would could get called more often, but it would always return the same pointer. Could look a bit cleaner by having a variable for `INSERT.from.SELECT.columns`.

```suggestion
      INSERT.from.SELECT.columns = INSERT.from.SELECT.columns
        .map((_, index) => index)
        .sort((a, b) => this.column_name(INSERT.from.SELECT.columns[a]) > this.column_name(INSERT.from.SELECT.columns[b]) ? 1 : -1)
        .map((index, i) => {
          this.columns[i] = columns[index]
          return INSERT.from.SELECT.columns[index]
        })
```

hana/lib/HANAService.js
```diff
@@ -847,6 +853,37 @@ class HANAService extends SQLService {
...
+      const entity = this.name(q._target.name, q)
+      const alias = INSERT.into.as
+      const elements = q.elements || q._target?.elements || {}
+      const columns = (INSERT.columns || ObjectKeys(elements)).filter(
+        c => c in elements && !elements[c].virtual && !elements[c].isAssociation,
+      )
+      
+      const selectedColumn = INSERT.from?.SELECT?.columns || INSERT.as?.SELECT?.columns
+      if (columns.length !== selectedColumn?.length)
+        throw new Error('The number of specified columns does not match the number of selected columns')
```
BobdenOs, line 867:
```suggestion
        cds.error`The number of specified columns does not match the number of selected columns`
```

hana/lib/HANAService.js
```diff
@@ -847,6 +853,37 @@ class HANAService extends SQLService {
...
+        .map((index, i) => {
+          this.columns[i] = columns[index]
+          return selectedColumn[index]
+        })
+      if (INSERT.from?.SELECT) INSERT.from.SELECT.columns = sortedColumns
+      if (INSERT.as?.SELECT) INSERT.as.SELECT.columns = sortedColumns
+
+      this.sql = `INSERT INTO ${this.quote(entity)}${alias ? ' as ' + this.quote(alias) : ''} (${this.columns.map(c =>
+        this.quote(c),
+      )}) ${this.SELECT(this.cqn4sql(INSERT.from || INSERT.as))}`
```
BobdenOs, line 882: `this.cqn4sql` will modify columns. Potentially adding more columns. I am not sure whether it should be called before the sorting or whether we pre sort for simplicity. But when `cqn4sql` is not called before the column sort it is important to reject `'*'`.

PR #1335, opened 2025-09-03
fix(`@cds.search`): no duplicates for search along `to-many` paths

patricebender:
Push down the search function into a subquery and match the subquery results with the primary keys of the `query._target`:

```cds
select from Authors where ID in (
 select ID from Authors where search(books.title, 'Harry Potter')
)
```

with this approach, we only get one result for `Authors` even if they have multiple `books` matching the search criteria.

---

this change needs #1337

### DIFF
--- a/db-service/test/cqn4sql/search.test.js
+++ b/db-service/test/cqn4sql/search.test.js
@@ -368,7 +368,26 @@ describe('search w/ path expressions', () => {
       } where search((authorWithAddress.note, address.city), 'x')
     )`
   
-  expect(JSON.parse(JSON.stringify(res))).to.deep.equal(expected)
+    expect(JSON.parse(JSON.stringify(res))).to.deep.equal(expected)
+  })
+
+  it('search along to-many path', () => {
+    // @cds.search: {books, books.genre.name}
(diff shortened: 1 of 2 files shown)


db-service/test/cqn4sql/search.test.js
```diff
@@ -419,4 +503,24 @@ describe('caching searchable fields', () => {
     let secondRun = cqn4sql(query, model)
     expect(JSON.parse(JSON.stringify(secondRun))).to.deep.equal(expected)
   })
+
+  it('to-many associations lead to exists subquery search', () => {
```
patricebender, line 507: todo

db-service/lib/infer/index.js
```diff
@@ -791,13 +791,30 @@ function infer(originalQuery, model) {
...
+    
+      const parts = [];
+      if (entitySources.length) {
+        parts.push(`not found in the elements of ${entitySources.join(', ')}`);
+      }
+      if (subquerySources.length) {
+        parts.push(`not exposed in the columns of ${subquerySources.join(', ')}`);
+      }
+    
+      throw new Error(`"${step}" ${parts.join('; ')}`);
```
patricebender, line 814: only some changes to format of error message

PR #1344, opened 2025-09-09
fix: set proper element link for path into fk

patricebender:
→ in the changed `if` branch, we (re) set the leaf link again to the last association. However, the path in the column already drilled down to the respective element. We just need to propagate the already calculated element and attach it to the flat column.

Before this change, in this particular case the association (`parent` in the test) was erroneously set as `element` link for the column, which led to a faulty `cqn2sql` rendering down the line.

### DIFF
--- a/db-service/test/cqn4sql/column.element.test.js
+++ b/db-service/test/cqn4sql/column.element.test.js
@@ -162,4 +162,16 @@ describe('assign element onto columns with flat model', () => {
     // foreign key is part of flat model
     expect(query.SELECT.columns[1]).to.have.property('element').that.eqls(TestPublisher.elements.publisherRenamedKey_notID)
   })
+
+  it('attaches foreign key and not the association as element', () => {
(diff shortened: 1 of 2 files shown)


db-service/test/cqn4sql/column.element.test.js
```diff
@@ -162,4 +162,16 @@ describe('assign element onto columns with flat model', () => {
...
+
+  it('attaches foreign key and not the association as element', () => {
+    let query = cqn4sql(cds.ql`
+      SELECT from bookshop.Books as Books {
+        genre.parent.ID,
+        genre.parent as assoc
+      }
+    `, model)
+    const { Genres } = model.entities
+    expect(query.SELECT.columns[0]).to.have.property('element').that.eqls(Genres.elements.ID)
```
patricebender, line 174: could also be `Genres:parent_ID` but the path navigates to the target → fine IMO. Semantically this should not make any difference.

PR #1351, opened 2025-09-11
chore: only own columns do not induce a semijoin in search

johannes-vogel:
follow up of #1346

@patricebender This solves the unnecessary semijoin if only own elements are searched. Please double check once you're back from vacation.


patricebender (commented):
lgtm

PR #1358, opened 2025-09-24
feat: sort selected columns & expands before formatting sql

PDT42:
Reopens https://github.com/cap-js/cds-dbs/pull/1330 after it was reverted due to it causing failing tests in the base cds package.


patricebender (changes requested):
I was always under the impression, that the order of columns matters for users of our services. Why else do we have mechanism in place, which guarantee stable column order, like the [smart wildcard](https://cap.cloud.sap/docs/cds/cql#smart-selector)?

The `cds-compiler` and `cqn4sql` have built-in logic to adhere to this principle. The rules can be summarized with the following case:

```cds
entity E {
  a : String;  b : String;
  c : String;  d : String;
};

entity After  as select from E { *, 1 as b, 2 as c    }; // a, b, c, d
entity Before as select from E {    1 as b, 2 as c, * }; // b, c, a, d
entity Around as select from E { 1 as b,  *,  2 as c  }; // b, a, c, d

entity Excluding as select from E { *, 2 as c  } excluding { c }; // a, b, d, c
```

→ this also goes for nested projections

please also see the sophisticated test suite, which makes sure that the `cap-js` database services adhere to those rules:

https://github.com/cap-js/cds-dbs/blob/d7f9094b67b4048e1feede554018dc08c8077cc0/db-service/test/cqn4sql/wildcards.test.js#L99

PR #1373, opened 2025-10-06
chore: re-write filter at `from.at(-1)` into queries `where`

patricebender:
this is a preparation for #1361 and it also makes our lifes easier as the infix filter at this place is semantically equivalent to a regular `where` clause.


db-service/test/cqn4sql/path-in-from.test.js
```diff
@@ -65,7 +65,7 @@ describe('infix filter on entities', () => {
       ] as Books {ID} where price > 5 group by price having price order by price limit 5`, model)
     expect(query).to.deep.equal(
       cds.ql`SELECT from bookshop.Books as Books {Books.ID}
-            WHERE (Books.price > 5) and (Books.price < 12.13)
+            WHERE Books.price < 12.13 and Books.price > 5
```
patricebender, line 68: only wrap in parens if really needed

PR #1391, opened 2025-10-28
fix: LimitedRank with compositions

larsplessing:
When accessed through the composition path, the LimitedRank value was incorrectly calculated due to an issue with the GROUP BY clause.


larsplessing:
blocked by cds version containing https://github.tools.sap/cap/cds/pull/5821

PR #1397, opened 2025-11-03
chore: `$main`

patricebender:
for internal usage only

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1041,9 +1041,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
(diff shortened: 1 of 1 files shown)


db-service/test/cqn4sql/filter-breakout.test/dollar-main.spec.js
patricebender: @ChristophWedler could you review the tests and let me know what you think? :)

patricebender: > For `union` the runtime explicitly doesn't support it. There is no logic currently to convert the `SET` into the corresponding SQL. So at least for that I would be confident that no one is using it right now.

https://github.com/cap-js/cds-dbs/blob/d8a26f84aa68aadca40f2f4e74c450f09b1a469f/db-service/lib/infer/index.js#L34

→ I will still check what is happening for stuff like `select from Authors, Books …`

sjvans:
@BobdenOs ok for you?

PR #1399, opened 2025-11-04
chore(cqn4sql): always use cqn list for search columns

johannes-vogel:
The current implementation for `cds.hana.fuzzy: false` always expects a cqn list of columns which is only the case if multiple columns are searched.

Imo, it makes it easier to implement against the stable representation as cqn list no matter how many columns are searched. Also (future use case) if app developers want to oversteer the default behavior, the structure is easier to use/document.

@patricebender I'll also adapt the tests if you agree with the design

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -2248,7 +2248,7 @@ function cqn4sql(originalQuery, model) {
     const searchFunc = {
       func: 'search',
       args: [
-        searchIn.length === 1 ? searchIn[0] : { list: searchIn },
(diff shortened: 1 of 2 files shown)


patricebender (commented):
I would prefer to add this to the `search` function. I did it for the other DBs:

https://github.com/cap-js/cds-dbs/blob/d7f9094b67b4048e1feede554018dc08c8077cc0/db-service/lib/cql-functions.js#L24

but I have no hard feelings

danjoa (dismissed):
👍 fully agreed

PR #1421, opened 2025-11-18
feat: `error` standard function

schwma:
This PR adds a standard function `error` that generates SQL statement that produces a runtime compatible error object, useful for i18n of error messages.

Example:

```cql
SELECT error('MESSAGE',(arg),(target1,target2)) FROM ...
```

Column `error` contains the following JSON:

```json
{
  "message": "MESSAGE",
  "args": ["valueOfArg"],
  "targets": ["target1", "target2"]
}
```

### DIFF
--- a/test/compliance/SELECT.test.js
+++ b/test/compliance/SELECT.test.js
@@ -1005,7 +1005,7 @@ describe('SELECT', () => {
       }, 'Ensure that the function reads correct parameters and returns the right values')
     })
 
-    test('nulls in lists are preserved', async () => {
+    test.skip('nulls in lists are preserved', async () => {
       const { string } = cds.entities('basic.projection')
       const cqn = cds.ql`SELECT error('MESSAGE',(null,null),(null)) FROM ${string}`
       cqn.SELECT.one = true


test/compliance/SELECT.test.js
```diff
@@ -920,6 +920,92 @@ describe('SELECT', () => {
...
+      const { string } = cds.entities('basic.projection')
+      const cqn = cds.ql`SELECT error('MESSAGE',('Arg1'),('Target1', 'Target2')) FROM ${string}`
+      cqn.SELECT.one = true
+      const res = await cds.run(cqn)
+      assert.ok(res[0].error, 'Ensure that the function is applied')
+      const funcRes = JSON.parse(res[0].error)
+      assert.deepStrictEqual(funcRes, {
+        message: 'MESSAGE',
+        args: ['Arg1'],
+        targets: ['Target1', 'Target2'],
```
BobdenOs: When calling `req.error` with the `{targets; [...]}` does it correctly set the OData targets in the response ?

BobdenOs (approved):

PR #1430, opened 2025-11-26
fix: introduce repository.url in package.json

johannes-vogel:
required for provenance publication, see https://github.com/cap-js/cds-dbs/actions/runs/19700654725/job/56435842449

### DIFF
--- a/hana/package.json
+++ b/hana/package.json
@@ -3,6 +3,10 @@
   "version": "2.4.0",
   "description": "CDS database service for SAP HANA",
   "homepage": "https://cap.cloud.sap/",
+  "repository": {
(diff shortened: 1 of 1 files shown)


patricebender (approved):
👍 does this need to be a `fix` (and hence part of the `CHANGELOG`?)

→ we need this also for `release/v1`

patricebender:
> > 👍 does this need to be a `fix` (and hence part of the `CHANGELOG`?)
> > → we need this also for `release/v1`
> 
> yes we also need that for v1. if not a fix, we need another way to retrigger the release process :)

ahh sure. I agree that this is the easiest way.

PR #1496, opened 2026-02-05
feat: cache prepare statement for reuse

BobdenOs:
Currently `@cap-js/hana` will always call `prepare` for queries that have placeholders `?`. With this PR the `SQL` statement is used as a cache key preventing `prepare` from being called. Allowing already prepared statements to be run directly. Reducing the amount of round trips. It is not allowed to run `SELECT` statements in `statement.runBatch`. Therefor it is not possible to use the event loop as an arguments collection mechanism to further reduce network overhead. As the HANA connection is (currently) only capable of having a single packed in flight there is no possibility that performance will be worsened by having statement executions be queued on the cached prepared statement.

There is a special case for streaming queries (incl. blob streams for `@sap/hana-client`). It is not possible to determine when the prepared `statement` is no longer required. Therefor the prepared statements for streaming queries are detached from the cache once initiated.


kasimos:
This will also try to store the preparations for any call statements, right? When trying out a different way to deal with that, we encountered issues with that, as some built-in procedures like setting application locks will only be executable once.
So one way would be to add the procedure calls with unique key (which I did with adding a comment that the sql preprocessor should ideally strip off before passing it to the plan cache, but no guarantees that it will...) or by adding them with a random uuid or the size of the Map as key, that would anyhow not be a valid sql and therefore shouldn't clash.

BobdenOs:
Good to know that there are these special procedure. There are similar problems with queries that want to stream their results. They would block the prepared statement for an extended period of time. Therefor in this PR I had added the ability to "detach" a prepared statement ([code](https://github.com/cap-js/cds-dbs/pull/1496/changes#diff-29c876f026b6f328fe0f0886d9fb9a325255e4e4f600612b5d2bf057c4ae3a96R15)). Which means that the `CALL` logic simply has to set the flag to `true` and it will get its own prepared statement.

hana/lib/drivers/base.js
```diff
@@ -9,47 +9,73 @@ class HANADriver {
...
+  _prepare(sql, detached) {
+    let prep = (!detached && this.statements[sql]) || module.exports.prom(
       this._native,
       'prepare',
     )(sql).then(stmt => {
       stmt._parentConnection = this._native
       return stmt
     })
+    if (!detached) {
+      const release = {} // TODO: for node@22 use Promise.withResolvers()
```
sjvans, line 34: @BobdenOs what about this todo?

BobdenOs: we can merge. I would still wait a bit before hard switching to `Promise.withResolvers()`

hana/lib/drivers/base.js
```diff
@@ -9,47 +9,73 @@ class HANADriver {
...
 
-  /**
-   * Generic prepare implementation for the native HANA drivers
-   * @param {string} sql The SQL string to be prepared
-   * @returns {import('@cap-js/db-service/lib/SQLService').PreparedStatement}
-   */
-  async prepare(sql) {
-    const prep = module.exports.prom(
+  _prepare(sql, detached) {
+    let prep = (!detached && this.statements[sql]) || module.exports.prom(
```
sjvans:
```suggestion
    detached ??= cds.env.requires.db.hana_statements_cache === false

    let prep = (!detached && this.statements[sql]) || module.exports.prom(
```

sjvans: → https://github.com/cap-js/cds-dbs/pull/1713

sjvans (approved):

patricebender (approved):

PR #1501, opened 2026-02-12
fix: sqlite generated key is named lastInsertRowid

johannes-vogel:


db-service/lib/InsertResults.js
```diff
@@ -71,7 +71,7 @@ module.exports = class InsertResult {
     // If no generated keys in entries/rows/values we might have database-generated keys
     const rows = this.results.slice(0, this.affectedRows) // only up to # of root entries
```
BobdenOs: `this.results` should always be of length `1`, because of the `json` insert SQLite only executes a single operation. For this operation it will return the number of affected rows and the last row ID. There for the `for` loop needs to include an additional loop that walks from `lastInsertRowid - affectedRows4` till `lastInsertRowid`.

BobdenOs (approved):

PR #1506, opened 2026-02-17
feat(hana): destroy connections that have exceeded the max statement count

sjvans:
instead of dropping all prepared statements on commit/ rollback, we count the number of prepared statements per connection and destroy instead of release that connection in case it accumulated more than the max allowed number of statements over its lifetime.


hana/lib/HANAService.js
```diff
@@ -37,6 +37,20 @@ class HANAService extends SQLService {
     this.on(['COMMIT'], this.onCOMMIT)
     this.on(['ROLLBACK'], this.onROLLBACK)
     this.on(['SELECT', 'INSERT', 'UPSERT', 'UPDATE', 'DELETE'], this.onNOTFOUND)
+
+    // TODO: remove
+    this.options._max_statement_count = 10
```
sjvans:
```suggestion
```

sjvans:
invalid. https://github.com/SAP/node-hdb/pull/293 instead.

PR #1521, opened 2026-03-02
feat: Add support for `cds.Vector` on SQLite and Postgres

schiwekM:
Hi colleagues,

this PR adds support for `cds.Vector` on SQLite & PG.
For HANA Cloud I've added a compatibility function to simplify the usage of the vector function by allowing a default mechanism for the HANA remote source.


vkozyura:
switched to https://github.com/cap-js/cds-dbs/pull/1676

PR #1538, opened 2026-03-19
fix: output converters for static input values

schiwekM:
While testing the new change tracking plugin with PG & HANA we came across SQL issues related to the output converters for date & time fields.


postgres/lib/PostgresService.js
```diff
@@ -569,10 +569,10 @@ GROUP BY k
       ...super.OutputConverters,
       Binary: e => `ENCODE(${e},'base64')`,
       LargeBinary: e => `ENCODE(${e},'base64')`,
-      Date: e => `to_char(${e}, 'YYYY-MM-DD')`,
-      Time: e => `to_char(${e}, 'HH24:MI:SS')`,
-      DateTime: e => `to_char(${e}, 'YYYY-MM-DD"T"HH24:MI:SS"Z"')`,
-      Timestamp: e => `to_char(${e}, 'YYYY-MM-DD"T"HH24:MI:SS.FF3"Z"')`,
+      Date: e => `to_char(${e}::timestamp, 'YYYY-MM-DD')`,
+      Time: e => `to_timestamp(${e}::text, 'HH24:MI:SS')::TIME`,
```
BobdenOs: This is still wrong. It is required to provide the expected time format inside the `to_char` function to ensure that CAP always returns the same time signature. It is possible to globally configure custom time formats and localization preferences. Therefor even if the default time format matches our expectation it is still required to ensure that we always return the exact format.

sqlite/lib/SQLiteService.js
```diff
@@ -214,7 +215,7 @@ class SQLiteService extends SQLService {
     }
 
     val(v) {
-      if (typeof v.val === 'boolean') v.val = v.val ? 1 : 0
+      if (typeof v.val === 'boolean') v.val = v.val && v.val !== 'false' ? 1 : 0
```
BobdenOs: how could `v.val` be `false` if it is inside the `if` block that checks that the type of `v.val` is `boolean` ?

johannes-vogel:
@schiwekM Is this PR still required? I don't see that this is needed

schiwekM:
@johannes-vogel with the remaining change it would only add consistency for the timestamp on sqlite as currently the formatting for timestamp diverges there.
But can be closed as the SQLite behaviour does not need to match HANA/PG I guess.

PR #1547, opened 2026-03-23
feat: decimal affinity in sqlite

PDT42:
Currently, when a field type is specified as `Decimal` for sqlite, we will accept that type as is. The cds-compiler plans to enable manually setting an "affinity" that allows to have decimal type fields be configured as `REAL_DECIMAL` in sqlite, which [according to the sqlite docs](https://www.sqlite.org/datatype3.html#:~:text=class%20into%20another.-,3.1.%20Determination%20Of%20Column%20Affinity,-For%20tables%20not), will cause the column to be interpreted as a `REAL`. 

We may want to wait for the `cdsc` and `db` flags to be aligned, before merging this.


PDT42:
Using `CAST(${expr} AS TEXT)` works for me locally, but will cause failing tests in the pipeline:

```bash
AssertionError [ERR_ASSERTION]: Ensure that the Database echos correct data back, property float does not match expected result.
  at dataTest (/home/runner/work/cds-dbs/cds-dbs/test/compliance/CREATE.test.js:155:16)
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: '-9.00719925474099e+15',
  expected: /-9007199254740991/,
  operator: 'match',
  diff: 'simple'
}

AssertionError [ERR_ASSERTION]: Ensure that the Database echos correct data back, property float does not match expected result.
  at dataTest (/home/runner/work/cds-dbs/cds-dbs/test/compliance/CREATE.test.js:155:16)
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: '9.00719925474099e+15',
  expected: /^9007199254740991/,
  operator: 'match',
  diff: 'simple'
}

AssertionError [ERR_ASSERTION]: Ensure that the Database echos correct data back, property float does not match expected result.
  at dataTest (/home/runner/work/cds-dbs/cds-dbs/test/compliance/CREATE.test.js:155:16)
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: '-9.00719925474099e+15',
  expected: /-9007199254740991/,
  operator: 'match',
  diff: 'simple'
}

AssertionError [ERR_ASSERTION]: Ensure that the Database echos correct data back, property float does not match expected result.
  at dataTest (/home/runner/work/cds-dbs/cds-dbs/test/compliance/CREATE.test.js:155:16)
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: '9.00719925474099e+15',
  expected: /^9007199254740991/,
  operator: 'match',
  diff: 'simple'
}
```

I found out (again) that rounding behavior in [sqlite changed to 17 significant digits with version 3.53.0](https://www.sqlite.org/releaselog/3_53_0.html#:~:text=Rounding%20is%20now%20done%20by%20default%20to%2017%20significant%20digits%2C%20instead%20of%2015), which is the version I am running locally, while CI uses 3.52.0. 

Our test data uses 16 digits, which (just so as it happens) is more than 15 and less than 17; and numbers that without affinity will default to `NUMERIC` ... For `NUMERIC` no floating point evaluation is involved in `CAST`. Now that the decimal always uses `REAL` affinity, it will be "rounded" during `CAST` and be put into scientific notation - in the pipeline - with sqlite v3.52.0, but not in my local environment  - with sqlite v3.53.0. 

Now ... I could: 

- Update the actual or expected test data
- Bump the sqlite version we use in the pipeline
- Work around the version issue by using `FORMAT` instead of `CAST`
   `CASE WHEN ${expr} IS NULL THEN NULL ELSE format('%.17g', ${expr}) END`
   Where `'%.17g'` will correctly handle the largest possible IEEE754 compatible number and strip trailing zeros

BobdenOs (approved):

PR #1558, opened 2026-03-30
docs: cqn4sql path expression transformations

patricebender:
## Summary
- Adds a document covering all join-related path expression transformations in cqn4sql
- Covers association-to-join, exists subqueries, expand/inline, infix filters, inner joins in exists, FROM navigation, and UPDATE/DELETE rewrites
- Documents the current limitation around join-relevant paths in infix filters and the planned concept from #1361


BobdenOs (approved):

db-service/doc/cqn4sql-transformations.md
```diff
@@ -0,0 +1,494 @@
...
+  left outer join Authors as author on author.ID = Books.author_ID
+{ Books.ID }
+where author.name = 'Brontë'
+```
+
+If a column and a WHERE clause both traverse the same association, they share the join.
+
+### FK shortcut — no join needed
+
+If the path accesses only a foreign key field (or the association itself without further navigation), no join is generated:
```
stewsk, line 75: clarify: what if also another field is accessed via the assoc?

db-service/doc/cqn4sql-transformations.md
```diff
@@ -0,0 +1,494 @@
...
+
+-- Output
+SELECT from Books
+  left outer join Authors as author
+    on author.ID = Books.author_ID
+    and author.placeOfBirth = 'Marbach'
+{ author.name as author_name }
+```
+
+The same association with **different filters** (or with/without a filter) produces **separate joins** with unique aliases:
```
stewsk, line 108: why "with/without"?

db-service/doc/cqn4sql-transformations.md
```diff
@@ -0,0 +1,494 @@
...
+    on author.ID = Books.author_ID and author.placeOfBirth = 'Marbach'
+  left outer join Authors as author2
+    on author2.ID = Books.author_ID
+{
+  author.name  as n1,
+  author2.name as n2
+}
+```
+
+Infix filters also apply in `exists` and `FROM` navigation, where they become conditions in the `WHERE EXISTS` subquery.
```
stewsk, line 129:
```suggestion
Infix filters also can occur in `exists` and `FROM` navigation, where they become conditions in the `WHERE EXISTS` subquery.
```

db-service/doc/cqn4sql-transformations.md
```diff
@@ -0,0 +1,494 @@
...
+
+---
+
+## 3. WHERE EXISTS Subqueries
+
+Both association paths in `FROM` and the `exists` predicate produce `WHERE EXISTS` correlated subqueries. The mechanics are the same — a `FROM` path like `Authors:books` is semantically equivalent to `SELECT from Books where exists author`.
```
stewsk, line 135:
```suggestion
Association paths both in `FROM` and in an `exists` predicate produce `WHERE EXISTS` correlated subqueries. The mechanics are the same — a `FROM` path like `Authors:books` is semantically equivalent to `SELECT from Books where exists author`.
```

db-service/doc/cqn4sql-transformations.md
```diff
@@ -0,0 +1,494 @@
...
+
+### Association in FROM
+
+An association step in the `FROM` clause rewrites the query target to the association's target entity and adds a `WHERE EXISTS` subquery correlating back to the source:
```
stewsk, line 139:
```suggestion
An association step in the `FROM` clause moves the query target to the association's target entity and adds a `WHERE EXISTS` subquery correlating back to the source:
```

stewsk (approved):

db-service/doc/cqn4sql-transformations.md
patricebender: → the same association once with a filter and once without a filter produce separate joins

PR #1564, opened 2026-04-09
feat: apply search result to orderby clause

BobdenOs:


db-service/lib/cqn4sql.js
```diff
@@ -69,6 +69,17 @@ function cqn4sql(originalQuery, model, useTechnicalAlias = true) {
...
       if (where) inferred.SELECT.where = where
       else if (having) inferred.SELECT.having = having
+      if (searchTerm.func) (inferred.SELECT.orderBy ??= []).unshift({ func: searchTerm.func, args: [...searchTerm.args, true], sort: 'desc' })
+      else if (searchTerm.xpr) {
+        const searchSelect = searchTerm.xpr[2]
+        const searchFunc = searchSelect.SELECT.where[0]
+          ; (inferred.SELECT.orderBy ??= []).unshift({
+            __proto__: SELECT.from(searchSelect.SELECT.from)
+              .columns({ func: searchFunc.func, args: [...searchFunc.args, true] })
+              .where([searchTerm.xpr[0], 'in', { list: searchSelect.SELECT.columns }]), // TODO: <-- ensure that the sub select in the order by is bound to the original query result row
```
patricebender, line 92: Could you add a test which needs the TODO to be resolved? I will then have a look how i could fix the correlation

BobdenOs: There are currently no very complex `@cds.search` models in the database tests. So the best reference I have found is the following test:

https://github.com/cap-js/cds-dbs/blob/7a7688f580701872ba8510f5709915e0805cbea0/db-service/test/cqn4sql/search.test.js#L376-L393

My understanding is that this is one of the tests that will end up inside this branch where the search is an expression. The problem is that when the same search query is used for the `order by` that it is not connected to the row of the original query. This is solved in the `where` clause as it is related to the original query through the `ID in (select)` pattern. So for the `order by` it would be required that the `(select)` is already related to the row of the original query. I hope this makes sense as I won't be able to write a reproduction test for this.

johannes-vogel:
consumed in https://github.com/cap-js/cds-dbs/pull/1725

PR #1565, opened 2026-04-09
fix: INSERT.from(SELECT) used non-existent `value` 

patricebender:
referencing undefined `value` column

`managed_default` unconditionally emitted `json_type(value, ...)` to check whether a column value was provided. This only works in `json_each()` contexts used by `INSERT.entries()`/`INSERT.rows()`. For `INSERT.from(SELECT)`, there is no `json_each()` because the source is a subquery aliased as `NEW`.

Now `managed_default` checks the INSERT context: for `INSERT.from(SELECT)`, it uses `src IS NULL` instead of `json_type(value, ...)`.

Closes #1559


patricebender:
@BobdenOs is this PR ready?

BobdenOs:
@patricebender it is ready, but it now requires an additional reviewer.

BobdenOs (dismissed):

sjvans (approved):

PR #1567, opened 2026-04-10
chore: add forSql alias

BobdenOs:


patricebender (approved):
imo `chore` is fine. Thanks 👍

PR #1572, opened 2026-04-13
test: filtered select one from entity with composite-key

PDT42:
Currently, the HANA Service will first query locked columns and then, in case `ignoreLocked` is set, filter a subsequent query for the rows it found. When an entity has a composite key, that filter is a check for whether the key tuple is `IN` a list of key tuples returned by the first query. In case no unlocked rows can be found, the way we format the filter will cause a HANA error, because an empty list parameter `{ list: [ ] }` will be interpreted as `{ val: null }` during `cqn4sql`, resulting in a HANA filter like `WHERE (keyField1, keyField2, ...) IS null`. 

This PR prevents the problem by exiting early in case no un-locked rows can be found.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -176,18 +176,20 @@ class HANAService extends SQLService {
       const resultQuery = query.clone()
       resultQuery.SELECT.forUpdate = undefined
       resultQuery.SELECT.forShareLock = undefined
+      
       const keys = Object.keys(req.target?.keys || {})
+      
       if (keys.length && query.SELECT.forUpdate?.ignoreLocked) {
-        // Exit early when ALL existing rows are locked
+        // Exit early when no row was found in the inital query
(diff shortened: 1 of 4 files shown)


hana/lib/HANAService.js
```diff
@@ -179,8 +178,11 @@ class HANAService extends SQLService {
       resultQuery.SELECT.forShareLock = undefined
       const keys = Object.keys(req.target?.keys || {})
       if (keys.length && query.SELECT.forUpdate?.ignoreLocked) {
-        // REVISIT: No support for count
-        // where [keys] in [values]
+        // Exit early when ALL existing rows are locked
+        if (rows.length === 0) return isOne ? undefined : []
+        
+        // REVISIT: No Support for count
+        // Filter for those rows that are not locked
```
BobdenOs:
```suggestion
        // Filter for rows that are locked by previous query
```

BobdenOs (approved):

PR #1575, opened 2026-04-14
chore: replace endless `forSQL` recursion

danjoa:
This reverts commit b82e094c4c6e4a64d837e92301b2d090a3d0ca8f.

### DIFF
--- a/db-service/lib/SQLService.js
+++ b/db-service/lib/SQLService.js
@@ -538,6 +538,7 @@ if (DEBUG_PQL._debug || cds.repl) {
           const cqn = db.srv.cqn4sql(this)
           return this.flat(cqn)
         }
+        forSql() { return this.forSQL() }
         toSQL() {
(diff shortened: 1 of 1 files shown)


db-service/lib/SQLService.js
```diff
@@ -538,7 +538,6 @@ if (DEBUG_PQL._debug || cds.repl) {
           const cqn = db.srv.cqn4sql(this)
           return this.flat(cqn)
         }
-        forSql() { return this.forSql() }
```
danjoa, line 541: What was that meant for? 
Calling `this.forSql()` in the implementation of `forSql()` looks like and endless recursion, doesn't it?

patricebender: this should've been `forSql() { return this.forSQL() }` → so that we can call both `forSQL` and `forSql`

patricebender (approved):

PR #1598, opened 2026-05-05
feat(sqlite): use node:sqlite as default driver

danjoa:


patricebender (commented):
this is a breaking change isn't it?

BobdenOs:
> this is a breaking change isn't it?

It shouldn't be a breaking change. As it would fallback to `node:sqlite`. If you know of things that don't work on `node:sqlite` let me know.

BobdenOs (approved):

johannes-vogel:
I think we should still ensure in some pipeline that everything works with better-sqlite3 and sql.js.

BobdenOs (approved):

PR #1604, opened 2026-05-08
fix: resolve `$self` in infix filters with path expressions and nested exists

patricebender:
When an infix filter contains a path expression or nested exists,
`$self` resolution failed because the subquery generated by
`transformSubquery` lacks the outer query's elements in its projection.

- Guard `$baseLink` branch so `$self` refs skip it and reach proper resolution
- Fall back to outermost query elements when local `queryElements` don't contain the referenced element
- Resolve `$self` directly to the outer query alias in `getTransformedTokenStream` for subquery contexts

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
       packages: write
 
     env:
-      cds_features_pool: true # TODO: make it work with Postgres test setup
+      cds_features_pool: builtin
       FORCE_COLOR: true
 
     strategy:
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -1823,6 +1823,17 @@ function cqn4sql(originalQuery, model, useTechnicalAlias = true) {
               continue
             }
             if (token.ref.length > 1 && token.ref[0] === '$self' && !token.$refLinks[0].definition.kind) {
+              if (inferred.outerQueries) {
+                const outerQuery = inferred.outerQueries[0]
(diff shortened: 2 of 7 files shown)


BobdenOs (commented):
Is it correct that this `$self`  behaves differently then the `@sap/cds-compiler` usage of `$self` ?

patricebender:
> Is it correct that this `$self` behaves differently then the `@sap/cds-compiler` usage of `$self` ?

do you have a concrete sample?

```cds
entity V as select from Parent {
  *,
  1+1 as foo,
  child[$self.foo = 2].ID as bar
}  where exists (SELECT from Child { 2 + 2 as foo, $self.foo as bar } where parent[$self.foo = 4].ID = 42);
```

into

```sql
CREATE VIEW V AS SELECT
  Parent_0.ID,
  1 + 1 AS calc,
  1 + 1 AS foo,
  child_2.ID AS bar
FROM (Parent AS Parent_0 LEFT JOIN Child AS child_2 ON ((child_2.parent_ID = Parent_0.ID)) AND ((1 + 1) = 2))
WHERE EXISTS (SELECT
    2 + 2 AS foo,
    2 + 2 AS bar
  FROM (Child AS Child_1 LEFT JOIN Parent AS parent_3 ON (Child_1.parent_ID = parent_3.ID) AND ((2 + 2) = 4))
  WHERE parent_3.ID = 42);
```

→ `exists` within infix filter are not possible with the `cds-compiler`

BobdenOs (approved):

PR #1605, opened 2026-05-08
fix: transient retries before invalidating credentials in built-in connection pool

swaldmann:
While the built-in pool uses `cds-mtxs` back-off mechanisms for fetching Service Manager credentials we currently invalidate credentials after the first rejection. We should also do some retries involving `dbc.connect()`  beforehand.


hana/lib/HANAService.js
```diff
@@ -50,9 +50,10 @@ class HANAService extends SQLService {
     }
     const isMultitenant = !!service.options.credentials.sm_url || !!service.options.credentials.baseurl || ('multiTenant' in this.options ? this.options.multiTenant : cds.env.requires.multitenancy)
     const acquireTimeoutMillis = this.options.pool?.acquireTimeoutMillis || (cds.env.profiles.includes('production') ? 1000 : 10000)
+    const maxRetries = 5
```
johannes-vogel: in the old implementation we had two retries.


```suggestion
    const maxRetries = 3
```

johannes-vogel:
we should also introduce this for the other dbs. With that we can remove again the global lock introduced in #1619

hana/lib/HANAService.js
```diff
@@ -50,9 +50,10 @@ class HANAService extends SQLService {
     }
     const isMultitenant = !!service.options.credentials.sm_url || !!service.options.credentials.baseurl || ('multiTenant' in this.options ? this.options.multiTenant : cds.env.requires.multitenancy)
     const acquireTimeoutMillis = this.options.pool?.acquireTimeoutMillis || (cds.env.profiles.includes('production') ? 1000 : 10000)
+    const maxRetries = 3
     return {
       options: this.options.pool || {},
-      create: async function create(tenant, start = Date.now()) {
+      create: async function create(tenant, start = Date.now(), attempt = 1) {
```
johannes-vogel, line 56: Shouldn't this be part of the builtin pool itself?

swaldmann: There are some specialties in the Service Manager retry handling, e.g. reacting to `retry-after` headers. Therefore I limited the retries here for the cached SM credential retrieval + `db.connect` part while the Service Manager retries continue using the MTX implementation.

PR #1622, opened 2026-05-22
fix: we can't return keys for `INSERT.from(SELECT.from(...))`

sjvans:
see https://github.com/cap-js/cds-dbs/commit/6400bca6dc74bebdfa104d1b36822397920800ee#diff-ee4a5323ba1f414d7ac44d28829146d94aeec457fa138f595b7edd68c6f5dfc1L29

### DIFF
--- a/db-service/lib/InsertResults.js
+++ b/db-service/lib/InsertResults.js
@@ -59,15 +59,14 @@ module.exports = class InsertResult extends Array {
     }
 
     const { INSERT } = this.query
+
+    // For INSERT.from(SELECT.from(...)), we can't return keys
+    if (INSERT.from || INSERT.as) return this
+
     const k0 = keys[0]
 
-    // For INSERT.from(SELECT.from(...)) return empty objects
-    if (INSERT.from || INSERT.as) {
-      for (let i = 0; i < this.affected; i++) this.push({})
-    }
-
     // For INSERT.entries() with generated keys in there return these keys
-    else if (INSERT.entries && k0 in INSERT.entries[0]) {
+    if (INSERT.entries && k0 in INSERT.entries[0]) {
       for (const d of INSERT.entries) {
         this.push (keys.reduce((p,k) => (p[k] = d[k], p), {}))
       }
--- a/db-service/lib/InsertResults.js
+++ b/db-service/lib/InsertResults.js
@@ -61,9 +61,9 @@ module.exports = class InsertResult extends Array {
     const { INSERT } = this.query
     const k0 = keys[0]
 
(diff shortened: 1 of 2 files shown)


db-service/lib/InsertResults.js
```diff
@@ -61,8 +61,13 @@ module.exports = class InsertResult extends Array {
     const { INSERT } = this.query
     const k0 = keys[0]
 
+    // For INSERT.from(SELECT.from(...)) return a dummy iterator with correct length
```
sjvans:
```suggestion
    // For INSERT.from(SELECT.from(...)) return empty objects
```

db-service/lib/InsertResults.js
```diff
@@ -61,8 +61,13 @@ module.exports = class InsertResult extends Array {
...
+    if (INSERT.from || INSERT.as) {
+      for (let i = 0; i < this.affectedRows; i++) this.push({})
```
sjvans:
```suggestion
      for (let i = 0; i < this.affected; i++) this.push({})
```

db-service/lib/InsertResults.js
```diff
@@ -61,8 +61,13 @@ module.exports = class InsertResult extends Array {
     const { INSERT } = this.query
     const k0 = keys[0]
 
+    // For INSERT.from(SELECT.from(...)) return empty objects
```
BobdenOs: At this point of the code this comment is not necessarily true. The target of the query has been identified. It has at least a key. This key could be a autogenerated key. Which we can reproduce with the `lastInsertRowid` logic at the end.

Which error is this PR trying to fix ? It looks like it should fall through to the last implementation that relies souly on the `this.results` data.

sjvans: `await INSERT.into(Authors).columns('ID').entries(SELECT.from('BookshopModel.Books').columns('ID'))` results in `d` being `undefined` in line 84 (because `[INSERT.values] → [undefined]` in line 83)

johannes-vogel: Shouldn't this return also an empty array until there's a "returning clause" which potentially returns something? For the time-being the loop does not contain any additional information.

chgeo:
Can we merge this soon?

Alternatively, please disable test `cds/tests/_runtime/cds-services/__tests__/integration/with-ql/api/run.test.js` that blocks all PRs atm.

BobdenOs (dismissed):

johannes-vogel:
@BobdenOs Please re-review. I added tests for the InsertResult variations.

BobdenOs (approved):

johannes-vogel (approved):

PR #1635, opened 2026-06-02
fix: calculated elements in drafts

johannes-vogel:


db-service/lib/cqn4sql.js
```diff
@@ -1812,7 +1812,9 @@ function cqn4sql(originalQuery, model, useTechnicalAlias = true) {
           // reject virtual elements in expressions as they will lead to a sql error down the line
           if (lhsDef?.virtual) throw new Error(`Virtual elements are not allowed in expressions`)
 
-          let result = is_regexp(token?.val) ? token : copy(token) // REVISIT: too expensive! //
+          let result = is_regexp(token?.val) ? token : copy(token) // REVISIT: too expensive!
+          // REVISIT: required because we copy the token above and lose the not enumerable "param: false"
```
danjoa, line 1816: We should avoid copying the token at all, imho. → see my old, ignored REVISIT comment above

danjoa (dismissed):

db-service/test/bookshop/srv/calc-elem-service.cds
```diff
@@ -0,0 +1,8 @@
+service CalcService {
+  @odata.draft.enabled
+  entity Orders {
+    key ID: Integer;
+    amount: Integer;
+    expensive: Integer = amount > 10 ? 1: 0;
+  }
+}
```
patricebender:
```suggestion
}

```

patricebender (approved):

db-service/lib/cqn4sql.js
johannes-vogel: Agreed. @patricebender could you check why that is required?

PR #1638, opened 2026-06-09
test: rm constructs which have multiple `$self` in one on-condition

patricebender:
cds-compiler has dropped support of those in PR #14798 and now emits an error with all versions >= 7. Removing the tests and the model to make the integration tests green.


ChristophWedler:
See compiler issue #14764 for a reason why `on` conditions with multiple backlink comparisons are not supported anymore.

johannes-vogel (approved):

PR #1641, opened 2026-06-10
chore(cqn4sql): replace deep copies with shallow spreads

patricebender:
`copy()` in `getTransformedTokenStream`, `getTransformedColumn`, the ORDER BY / GROUP BY transformer, and `getTransformedFrom` all only mutate top-level properties on the result, making a full recursive deep copy unnecessary. Shallow spread suffices.

`calculateOnCondition` still uses `copy()` — it mutates nested token objects in-place during iteration, so a shallow copy would corrupt the original token stream.


patricebender:
@johannes-vogel pls review

danjoa (approved):
We might also optimize the remaining deep clone by a lazy copy-on-change approach

PR #1644, opened 2026-06-11
chore(v2): prepare downport branch

patricebender:
Prepare `release/v2` to act as the v2 maintenance line, mirroring the setup we did for `release/v1` last year.

- Retarget `release-please.yml` to `release/v2`; publish all four packages under the `maintenance` npm dist-tag
- Close `@sap/cds` peer-dep ranges to `^9.8` (`@sap/cds-dk` to `^9`)
- Add `pull-request-title-pattern: chore(v2): release downport` so v2 release PRs are distinguishable
- Align `conventional-commits.yml` trigger with `main` (`pull_request_target` → `pull_request`)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,11 +2,11 @@ name: Tests
 
 on:
   push:
-    branches: [main]
+    branches: [release/v2]
   pull_request:
     types: [opened, synchronize, reopened, auto_merge_enabled]
 
-# Allow parallel jobs on `main`, so that each commit is tested. For PRs, run only the latest commit.
+# Allow parallel jobs on `release/v2`, so that each commit is tested. For PRs, run only the latest commit.
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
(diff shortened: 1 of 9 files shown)


johannes-vogel (approved):

johannes-vogel (changes requested):
please also check that the tests are triggered

.github/workflows/test.yml
```diff
@@ -47,10 +45,10 @@ jobs:
         env:
           TAG: ${{ steps.hxe.outputs.TAG }}
           IMAGE_ID: ${{ steps.hxe.outputs.IMAGE_ID }}
-      - name: sqlite driver (node:sqlite)
+      - name: sqlite driver (better-sqlite3))
         run: npm test -w sqlite
         env:
-          CDS_REQUIRES_DB_DRIVER: 'node'
+          CDS_REQUIRES_DB_DRIVER: 'better-sqlite3'
```
patricebender: is it correct that the default test driver should be `better-sqlite`?

johannes-vogel: yes. in v2 code line the default is better-sqlite3 which is already run before as part of `npm t -ws`. These test runs ensure compatibility with node native sqlite and sql.js.

→ this change should be reverted

johannes-vogel (approved):

PR #1646, opened 2026-06-12
fix: transient retries before invalidating credentials in built-in connection pool

swaldmann:
While the built-in pool uses `cds-mtxs` back-off mechanisms for fetching Service Manager credentials we currently invalidate credentials after the first rejection. We should also do some retries involving `dbc.connect()`  beforehand. 

Also implements the same simple retry mechanism for Postgres. 

We'll consider later if this should become a generic part of the pool implementation, including the HTTP retry handling currently used in the Service Manager client.


hana/lib/HANAService.js
```diff
@@ -70,8 +71,13 @@ class HANAService extends SQLService {
                 throw new Error(`Pool failed connecting to '${tenant}'`, { cause: err })
               }
               const deadline = start + acquireTimeoutMillis
+              if (attempt <= maxRetries && Date.now() < deadline) {
+                // Retry transient connection failures before invalidating credentials
+                LOG.error('connection attempt', attempt, 'of', maxRetries, 'failed - retrying')
```
johannes-vogel: I think this is rather a debug/info level. If the hickup is solved after retry, there's nothing to do.

hana/lib/HANAService.js
```diff
@@ -50,9 +50,10 @@ class HANAService extends SQLService {
     }
     const isMultitenant = !!service.options.credentials.sm_url || !!service.options.credentials.baseurl || ('multiTenant' in this.options ? this.options.multiTenant : cds.env.requires.multitenancy)
     const acquireTimeoutMillis = this.options.pool?.acquireTimeoutMillis || (cds.env.profiles.includes('production') ? 1000 : 10000)
+    const maxRetries = 3
     return {
       options: this.options.pool || {},
-      create: async function create(tenant, start = Date.now()) {
+      create: async function create(tenant, start = Date.now(), attempt = 1) {
```
johannes-vogel: I think we should have the same retries for single tenant hana. Maybe it makes sense to have two separate create functions for st and mt. This would also improve readability.

johannes-vogel (approved):

PR #1653, opened 2026-06-16
fix: consolidated flags to cds.features.use_generic_pool

danjoa:
We should consolidate and harmonize our kill switches

### DIFF
--- a/db-service/lib/common/generic-pool.js
+++ b/db-service/lib/common/generic-pool.js
@@ -1,7 +1,7 @@
 const cds = require('@sap/cds')
 const LOG = cds.log('db')
 
-const use_new_pool = cds.requires.db?.pool?.builtin === false || cds.env.features.pool === 'generic-pool' ? false : true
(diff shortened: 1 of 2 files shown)


BobdenOs (dismissed):

db-service/lib/common/generic-pool.js
```diff
@@ -1,7 +1,7 @@
 const cds = require('@sap/cds')
 const LOG = cds.log('db')
 
-const use_new_pool = cds.requires.db?.pool?.builtin === false || cds.env.features.pool === 'generic-pool' ? false : true
+const use_new_pool = cds.env.features.use_generic_pool
```
BobdenOs:
```suggestion
const use_new_pool = !cds.env.features.use_generic_pool
```

I assume that `use_new_pool` should be on when `use_generic_pool` is not truthy.

johannes-vogel (approved):

Issue #1669, opened 2026-06-29
Integer literals in `case` statements not working


stockbal:
### Description of erroneous behaviour

We have built a draft-enabled odata service with the following definition:

```cds
    @odata.draft.enabled
    @Capabilities.Insertable: false
    @Capabilities.Deletable : false
    @Capabilities.Updatable : true
    entity ManualSyncs as
        projection on demo.ManualSyncs {
            *,
            (
                case
                    when matchFound = #NoMatch
                         then 1
                    when matchFound = #FullMatch
                         then 3
                    when matchFound = #PartialMatch
                         then 2
                    else 0
                end
            ) as matchFoundIndicator : Integer
        };
```

When exposing this entity in a Fiori Elements List Report application and triggering a GET request with the standard filters, the `hdb` module will trigger the following error:

```
[odata] - > GET /ManualSyncs {
  '$count': 'true',
  '$select': 'HasActiveEntity,HasDraftEntity,ID,IsActiveEntity,comment,createdAt,matchFound,matchFoundIndicator,modifiedAt,orderID',
  '$expand': 'DraftAdministrativeData($select=DraftUUID,InProcessByUser)',
  '$filter': '(IsActiveEntity eq false or SiblingEntity/IsActiveEntity eq null)',
  '$skip': '0',
  '$top': '30'
}
[error] - 500 - Error: Cannot set parameter at row: 1. Argument must be a string
    at next (/draft-calc-elements/node_modules/hdb/lib/protocol/ExecuteTask.js:184:19)
    at ExecuteTask.getParameters (/draft-calc-elements/node_modules/hdb/lib/protocol/ExecuteTask.js:204:3)
    at ExecuteTask.sendExecute (/draft-calc-elements/node_modules/hdb/lib/protocol/ExecuteTask.js:211:8)
    at execute (/draft-calc-elements/node_modules/hdb/lib/protocol/ExecuteTask.js:82:10)
    at ExecuteTask.run (/draft-calc-elements/node_modules/hdb/lib/protocol/ExecuteTask.js:124:3)
    at run (/draft-calc-elements/node_modules/hdb/lib/util/Queue.js:97:12)
    at Queue.dequeue (/draft-calc-elements/node_modules/hdb/lib/util/Queue.js:100:3)
    at Queue.push (/draft-calc-elements/node_modules/hdb/lib/util/Queue.js:48:10)
[...]

PR #1670, opened 2026-06-30
fix: use integer literals instead of placeholders in xpr

patricebender:
reported in #1669 → not an issue with draft nor calculated element but with integer literals in case statements in general


johannes-vogel:
I think this is fixed already with cds10. in the original issue, the case when statement becomes a calculated element through the draft enablement. Otherwise, the runtime would rely on the db view and not add the case when stmt manually to the statement.

PR #1674, opened 2026-07-02
refactor: simplify vector embedding implementation

vkozyura:
- Replace manual ONNX implementation with @xenova/transformers
- Add deterministic hash-based fallback (port of Java's HashEmbeddingService)
- Move embedding computation to CAP db.before handlers (async-friendly)
- Keep synchronous pure-math functions in SQLite (COSINE_SIMILARITY, L2DISTANCE, L2NORMALIZE)

This approach:
- Reduces code from ~619 lines to ~275 lines
- Uses maintained library instead of copied ONNX internals
- Provides deterministic fallback for testing (same input = same output)
- Solves sync/async problem by pre-computing embeddings at CAP level

The VECTOR_EMBEDDING SQL function now throws an error if called directly, as embeddings should be computed via CAP handlers using @cds.vectorSource annotation.


vkozyura:
switched to https://github.com/cap-js/cds-dbs/pull/1676

PR #1676, opened 2026-07-03
feat: Add support for `cds.Vector` on SQLite and Postgres

vkozyura:
## Summary
- Add SQLite vector functions: `COSINE_SIMILARITY`, `L2DISTANCE`, `L2NORMALIZE`, `VECTOR_EMBEDDING`
- Use synchronous hash-based embedding for `VECTOR_EMBEDDING` (deterministic, suitable for testing/development)
- Vector functions work directly in SQL queries, including with column references

## Implementation
- Vector functions defined inline in SQLiteService.js following existing patterns
- Helper functions in `sqlite/lib/vector_handling/index.js`
- Hash-based embedding produces 384-dimensional normalized vectors

## Test plan
- [x] COSINE_SIMILARITY: identical, orthogonal, opposite vectors, null handling
- [x] L2DISTANCE: identical vectors, unit vectors, known distance
- [x] L2NORMALIZE: normalizes to unit length, already normalized unchanged
- [x] VECTOR_EMBEDDING: computes embedding, deterministic, different inputs, null handling
- [x] Semantic search: ORDER BY similarity, dynamic embedding on columns


schiwekM (approved):

hana/lib/cql-functions.js
```diff
@@ -267,6 +267,15 @@ const HANAFunctions = {
   HIERARCHY: undefined,
   HIERARCHY_DESCENDANTS: undefined,
   HIERARCHY_ANCESTORS: undefined,
+
+  // Ease of use for stakeholders for auto-filling the remote source
+  vector_embedding(text, text_type, model_and_version, remote_source = cds.env.ai?.embeddings?.remoteSource) {
```
BobdenOs: What is the source of `cds.env.ai?.embeddings?.remoteSource` ? If this comes from the previous PR or the AI plugin we should probably communicate this with java as well as they are looking into a global option as well.

I would expect also an global setting for `model_and_version` so that it doesn't need to be hardcoded into models.

postgres/lib/PostgresService.js
```diff
@@ -49,6 +50,14 @@ class PostgresService extends SQLService {
...
           const dbc = new Client({ ...credentials, ...clientOptions })
           await dbc.connect()
+          // cds.Vector support for PG
+          try {
+            await dbc.query('CREATE EXTENSION IF NOT EXISTS vector')
+            await pgvector.registerTypes(dbc)
+          } catch (e) {
+            const LOG = cds.log('postgres')
+            LOG.debug('pgvector extension not available, skipping vector support:', e.message)
+          }
```
BobdenOs: This should not be ran every `create`. This might not even need to run at all depending on how the postgres build task is deploying the model.

postgres/lib/PostgresService.js
```diff
@@ -4,6 +4,7 @@ const cds = require('@sap/cds')
 const crypto = require('crypto')
 const { Writable, Readable } = require('stream')
 const sessionVariableMap = require('./session.json')
+const pgvector = require('pgvector/pg');
```
BobdenOs: Should be optional

postgres/package.json
```diff
@@ -28,7 +28,8 @@
   },
   "dependencies": {
     "@cap-js/db-service": "^3.0.1",
-    "pg": "^8"
+    "pg": "^8",
+    "pgvector": "^0.2.1"
```
BobdenOs: should be optional

postgres/pg-stack.yml
```diff
@@ -3,10 +3,13 @@ version: '3.1'
...
 services:
   db:
-    image: postgres:16-alpine
+    image: pgvector/pgvector:pg16
     restart: always
     environment:
       POSTGRES_PASSWORD: postgres
+      LANG: en_US.utf8
+      LC_COLLATE: en_US.utf8
+      LC_CTYPE: en_US.utf8
```
BobdenOs, line 12: why?

sqlite/lib/vector_handling/index.js
```diff
@@ -0,0 +1,92 @@
...
+  if (typeof vector === 'string') return JSON.parse(vector)
+  if (Array.isArray(vector)) return vector
+  throw new Error(`Unsupported vector type: ${typeof vector}`)
+}
+
+function fromFloatArray(arr, original) {
+  return original instanceof Float32Array ? new Float32Array(arr) : JSON.stringify(arr)
+}
+
+module.exports = { cosineSimilarity, l2Distance, l2Normalize, hashEmbedding, toFloatArray, fromFloatArray }
```
BobdenOs: As discussed just include it in the SQLiteService file directly. No need for reference confusion.

test/compliance/functions.test.js
```diff
@@ -1,5 +1,7 @@
 const cds = require('../cds.js')
 
+const isHana = () => cds.db?.options?.impl === '@cap-js/hana'
```
BobdenOs: This check is not allowed to exist in `cds-dbs`.

postgres/pg-stack.yml
vkozyura: was an ai attempt to solve sorting/collation issues -> removed

postgres/package.json
vkozyura: removed

postgres/lib/PostgresService.js
vkozyura: removed

postgres/lib/PostgresService.js
vkozyura: commented out with a revisit containing possible options - pg tests are currently disabled

sqlite/lib/vector_handling/index.js
vkozyura: moved

test/compliance/functions.test.js
vkozyura: moved to sqlite/test/vector.test.js

hana/lib/cql-functions.js
vkozyura: The cds.env.ai.embeddings.remoteSource configuration is introduced by the @cap-js/ai. 
Java stack has no ai property in CdsProperties.java. The Java test hardcodes the model.

The Parameter model_and_version is called embeddingModel in Java.

Suggestion: Sync with Java as follow-up

postgres/test/vector.test.js
```diff
@@ -0,0 +1,132 @@
...
+      CREATE OR REPLACE FUNCTION public.vector_embedding(model text, input text)
+      RETURNS text AS $$
+        SELECT CASE WHEN input IS NULL THEN NULL
+          ELSE (SELECT json_agg(sin(i * hashtext(input)::float8 / 1000))::text
+                FROM generate_series(1, 384) i)
+        END;
+      $$ LANGUAGE SQL IMMUTABLE;
+    `)
+    await client.end()
+  })
```
BobdenOs: you should run this in a `before` handler as that will allow `cds.test` to create the database information. You can access the correct credentials using the `cds.db.options.credentials.__system__` for the root connection which you can create using `cds.connect.to('sys', {...cds.db.options, credentials: cds.db.options.credentials.__system__})`

test/scenarios/bookshop/read.test.js
```diff
@@ -376,12 +376,15 @@ describe('Bookshop - Read', () => {
 
     const q = cds.ql`SELECT title FROM sap.capire.bookshop.Books ORDER BY title`
     const res3 = await cds.run(q)
-    expect(res3.at(-1).title).to.be.eq('dracula')
+    // pgvector image: 'Wuthering Heights' sorts last (W > d in Unicode order)
+    // sqlite/alpine: 'dracula' sorts last (different collation behavior)
```
BobdenOs: This comment means that there is something broken in regards to `localized` vs `binary` sorting. Which is determined by the query. This means that this change to the test is unacceptable and the root cause has to be found and addressed.

test/scenarios/bookshop/runtime-views.test.js
```diff
@@ -159,7 +159,11 @@ describe('Runtime Views', () => {
           .orderBy('title')
 
         const authors = res.map(b => b.author.name)
-        expect(authors).to.deep.equal(['Richard Carpenter', 'Edgar Allen Poe', 'Charlotte Brontë', 'Edgar Allen Poe', "Emily Brontë"])
+        // pgvector image: 'Emily Brontë' (Wuthering Heights) comes second
+        // sqlite/alpine: 'Emily Brontë' sorted last (different collation behavior)
```
BobdenOs: Same as mentioned before. This change to the test is not allowed.

BobdenOs (changes requested):
The test changes are not allowed.

hana/lib/cql-functions.js
vkozyura: [DONE] Sqlite, Postgres: provide only hash vector_embedding by default
[FOLLOW-UP] HANA: Configuration based on Java profiles

test/scenarios/bookshop/read.test.js
vkozyura: solved

test/scenarios/bookshop/runtime-views.test.js
vkozyura: solved

postgres/lib/PostgresService.js
vkozyura: follow-up to be created

postgres/test/vector.test.js
vkozyura: deleted

anselm94:
I always thought, we have this support in SQLite for local development, but only just stumbled upon this small note in https://cap.cloud.sap/docs/releases/2026/apr26#vector-embeddings

> ... CAP Node.js will follow soon.

I'm interested in the release of this feature. A tentative timeline would help me :)

postgres/lib/PostgresService.js
```diff
@@ -49,6 +55,30 @@ class PostgresService extends SQLService {
...
+            try {
+              await dbc.query(`
+                CREATE OR REPLACE FUNCTION public.vector_embedding(model text, input text)
+                RETURNS text AS $$
+                  SELECT CASE WHEN input IS NULL THEN NULL
+                    ELSE (SELECT json_agg(sin(i * hashtext(input)::float8 / 1000))::text
+                          FROM generate_series(1, 384) i)
+                  END;
+                $$ LANGUAGE SQL IMMUTABLE;
+              `)
```
BobdenOs: Move this to the `tenant` function or the `npm start` script just like the `create extension` setup step.

postgres/lib/PostgresService.js
```diff
@@ -89,7 +119,7 @@ class PostgresService extends SQLService {
         JSON.stringify(env),
       ]),
       ...(this.options?.credentials?.schema
-        ? [this.exec(`SET search_path TO "${this.options?.credentials?.schema}";`)]
+        ? [this.exec(`SET search_path TO "${this.options?.credentials?.schema}"${pgvector ? ', public' : ''};`)] // include public for extensions like pgvector when available
```
BobdenOs: Just always include `public`

postgres/test/vector.test.js
```diff
@@ -0,0 +1,122 @@
...
+      host: 'localhost',
+      port: 5432,
+      user: 'postgres',
+      password: 'postgres',
+      database: testDb
+    })
+    await client.connect()
+    await client.query('CREATE EXTENSION IF NOT EXISTS vector')
+    await client.end()
+  })
```
BobdenOs: This should all be removed. This setup should happen already for all `cds-dbs` tests.

postgres/package.json
```diff
@@ -24,12 +24,15 @@
...
     "test": "npm start && cds-test",
-    "start": "docker compose -f pg-stack.yml up -d"
+    "start": "docker compose -f pg-stack.yml up -d --wait && docker compose -f pg-stack.yml exec db psql -U postgres -c 'CREATE EXTENSION IF NOT EXISTS vector;' || true"
   },
   "dependencies": {
     "@cap-js/db-service": "^3.0.1",
     "pg": "^8.22.0"
   },
+  "devDependencies": {
+    "pgvector": "^0.3.0"
```
BobdenOs: this should be an optional `peerDependency`.

BobdenOs (changes requested):

postgres/lib/PostgresService.js
vkozyura: moved to database()

postgres/test/vector.test.js
vkozyura: done

postgres/package.json
vkozyura: done. pgvector npm package was only needed for type parsing, which is optional.

postgres/lib/PostgresService.js
vkozyura: done

postgres/lib/PostgresService.js
```diff
@@ -47,8 +50,14 @@ class PostgresService extends SQLService {
...
         }
         try {
-          const dbc = new Client({ ...credentials, ...clientOptions })
+          const dbc = new Client({ ...credentials, ...clientOptions, ...tenant })
           await dbc.connect()
+
+          // Register vector type parsers after connection (skip if connect is mocked in tests)
+          if (pgvector && dbc._connected !== false) {
+            try { await pgvector.registerTypes(dbc) } catch { /* not available or mocked */ }
+          }
```
BobdenOs: We do not have mock `pg` clients. After `dbs.connect` checkking `_connected` is next level.

postgres/lib/PostgresService.js
```diff
@@ -89,7 +98,7 @@ class PostgresService extends SQLService {
         JSON.stringify(env),
       ]),
       ...(this.options?.credentials?.schema
-        ? [this.exec(`SET search_path TO "${this.options?.credentials?.schema}";`)]
+        ? [this.exec(`SET search_path TO "${this.options?.credentials?.schema}", public;`)]
```
BobdenOs, line 96: if you do not create the function in the `public` schema it is not required to add the `public` schema here as well.

postgres/lib/PostgresService.js
```diff
@@ -619,15 +645,31 @@ GROUP BY k
...
-      if (exists.rowCount) return
-      await this.exec(`
-        DROP GROUP IF EXISTS "${creds.usergroup}";
-        DROP USER IF EXISTS "${creds.user}";
-        CREATE GROUP "${creds.usergroup}";
-        CREATE USER "${creds.user}" WITH CREATEROLE IN GROUP "${creds.usergroup}" PASSWORD '${creds.user}';
-        GRANT "${creds.usergroup}" TO "${creds.user}" WITH ADMIN OPTION;
-      `)
-      await this.exec(`CREATE DATABASE "${creds.database}" OWNER="${creds.user}" TEMPLATE=template0`)
+      if (!exists.rowCount) {
```
BobdenOs: If the database already exists there is nothing to do and the `return` is correct.

postgres/lib/PostgresService.js
```diff
@@ -619,15 +645,31 @@ GROUP BY k
...
+        try {
+          await dbCon.query('CREATE EXTENSION IF NOT EXISTS vector')
+        } finally {
+          await dbCon.end()
+        }
+        // Create vector_embedding function in public schema
+        await this._createVectorEmbeddingFunction(creds.database, system)
+      } catch {
+        /* ignore if extension/function creation fails */
+      }
```
BobdenOs: This is a lot of useless code for calling a single SQL statement. Just move it to the `tenant` function where this user is already connected and doing work. Especially as the current code is not even skipping this query when the tenant already exists.

postgres/lib/PostgresService.js
```diff
@@ -681,7 +723,11 @@ GROUP BY k
       // Create new schema using schema owner
       await this.tx(async tx => {
         await tx.run(`DROP SCHEMA IF EXISTS "${creds.schema}" CASCADE`)
-        if (!clean) await tx.run(`CREATE SCHEMA "${creds.schema}" AUTHORIZATION "${creds.user}"`)
+        if (!clean) {
+          await tx.run(`CREATE SCHEMA "${creds.schema}" AUTHORIZATION "${creds.user}"`)
+          // Create vector_embedding function in public schema
+          await this._createVectorEmbeddingFunction(creds.database, this.options.credentials)
```
BobdenOs: Just move the SQL statement here, because you are doing a bunch of connection logic again inside the function. Just put the function in the current schema instead of public. Resolves also other cascading changing in the PR.

postgres/lib/PostgresService.js
```diff
@@ -602,6 +612,22 @@ GROUP BY k
     }
   }
 
+  async _createVectorEmbeddingFunction(database, credentials) {
+    try {
+      const con = await this.factory.create({ ...credentials, database })
+      await con.query(`
+        CREATE OR REPLACE FUNCTION public.vector_embedding(model text, input text)
```
BobdenOs: This is not the signature of the [`vector_embedding`](https://help.sap.com/docs/hana-cloud-database/sap-hana-cloud-sap-hana-database-vector-engine-guide/vector-embedding-function-vector) function.

It is supposed to be `VECTOR_EMBEDDING(<text>, <text-type>, <model-and-version>[, [<remote_source_schema_name>.]<remote_source_name>] [, <options>])`

The same for the `sqlite` function is using the wrong signature and all the tests are also poisoned with the wrong signature.

BobdenOs (dismissed):

postgres/lib/PostgresService.js
```diff
@@ -47,8 +50,14 @@ class PostgresService extends SQLService {
             }),
         }
         try {
-          const dbc = new Client({ ...credentials, ...clientOptions })
+          const dbc = new Client({ ...credentials, ...clientOptions, ...tenant })
```
BobdenOs: `tenant` is a string and the `pg` client does not know about `tenants`.

postgres/lib/PostgresService.js
vkozyura: reverted to return

postgres/lib/PostgresService.js
vkozyura: simplified :)

postgres/lib/PostgresService.js
vkozyura: in database() extension is created in public schema, that's why we need public in search_path, without it we would obtain; error: type "vector" does not exist

postgres/lib/PostgresService.js
vkozyura: create function is moved to tenant, but we need to keep create extension here and do it with superuser

postgres/lib/PostgresService.js
vkozyura: done

postgres/lib/PostgresService.js
vkozyura: fixed in code and in tests

postgres/lib/PostgresService.js
vkozyura: FOLLOW-UP: how to handle try-catch? Try-catch is needed for creating connection which will create an extension, which does not yet exists during connection creation.

postgres/lib/PostgresService.js
vkozyura: FOLLOW-UP: How to avoid creating an additional connection? Using con in database() doesn't work because it points to the default db postgres. with this con we do this.exec(`CREATE DATABASE') and then we need a superuser connection to this db.

postgres/lib/cql-functions.js
```diff
@@ -191,6 +191,33 @@ const HANAFunctions = {
...
   },
+
+  /**
+   * Computes the cosine similarity of two vectors
+   * @param {*} v1 - Vector 1
+   * @param {*} v2 - Vector 2
+   * @returns {string} - SQL statement
+   */
+  cosine_similarity(v1, v2) {
+    return `1 - cosine_distance(${this.expr(v1)},${this.expr(v2)})`
```
PDT42, line 202: i.e. distance is small -> similarity is great. Can people also use `cosine_distance` directly?

postgres/lib/PostgresService.js
```diff
@@ -535,6 +539,7 @@ GROUP BY k
       DateTime: () => 'TIMESTAMP',
       Timestamp: () => 'TIMESTAMP',
       Map: () => 'JSONB',
+      Vector: () => 'vector',
```
PDT42: Any reason not to use `VECTOR` and align with the other types? Irrelevant?

postgres/lib/PostgresService.js
```diff
@@ -614,26 +619,24 @@ GROUP BY k
     const system = cds.requires.db.credentials
 
     try {
-      const con = await this.factory.create(system)
+      const con = await this.factory.create()
```
PDT42, line 622: If we don't use `system` in `factory.create()` anymore - what creds do we use?

postgres/lib/PostgresService.js
PDT42, line 648: Is `this.options.credentials` different from `cds.requires.db.credentials`? Shouldn't these be ... the same?

postgres/lib/PostgresService.js
```diff
@@ -681,7 +684,19 @@ GROUP BY k
...
         await tx.run(`DROP SCHEMA IF EXISTS "${creds.schema}" CASCADE`)
-        if (!clean) await tx.run(`CREATE SCHEMA "${creds.schema}" AUTHORIZATION "${creds.user}"`)
+        if (!clean) {
+          await tx.run(`CREATE SCHEMA "${creds.schema}" AUTHORIZATION "${creds.user}"`)
+          // Create vector_embedding function in tenant schema
+          await tx.run(`
+            CREATE OR REPLACE FUNCTION "${creds.schema}".vector_embedding(input text, text_type text, model_and_version text)
+            RETURNS text AS $$
+              SELECT CASE WHEN input IS NULL THEN NULL
+                ELSE (SELECT json_agg(sin(i * hashtext(input)::float8 / 1000))::text
```
PDT42, line 694: This is the "embedding", correct? We simply spoof real embedding and return a hash as text, right?

sqlite/lib/SQLiteService.js
```diff
@@ -321,4 +315,106 @@ function loadSQLite(driver) {
   }
 }
 
+// define date and time functions in js to allow for throwing errors
+const isTime = /^\d{1,2}:\d{1,2}:\d{1,2}$/
+const hasTimezone = /([+-]\d{1,2}:?\d{0,2}|Z)$/
```
PDT42: Why not just ... define functions? Why regex' with function like names?

sqlite/lib/SQLiteService.js
```diff
@@ -321,4 +315,106 @@ function loadSQLite(driver) {
...
+const toDate = (d, allowTime = false) => {
+  const date = new Date(allowTime && isTime.test(d) ? `1970-01-01T${d}Z` : hasTimezone.test(d) ? d : d + 'Z')
+  if (Number.isNaN(date.getTime())) throw new Error(`Value does not contain a valid ${allowTime ? 'time' : 'date'} "${d}"`)
```
PDT42, line 323: I don't understand this error and it should probably be using `cds.error`?

sqlite/lib/SQLiteService.js
```diff
@@ -321,4 +315,106 @@ function loadSQLite(driver) {
...
+  for (let i = 0; i < v.length; i++) v[i] /= norm
+  return v
+}
+
+/**
+ * Deterministic synchronous hash-based embedding function for SQLite.
+ * NOTE: If a synchronous embedding library becomes available for Node.js,
+ * it can be integrated here to replace the hash-based implementation.
+ */
+function hashEmbedding(text, dimensions = 384) {
```
PDT42, line 365: Why use 384? Seems to be the particular dimensionality of [Xenova/all-MiniLM-L6-v2](https://huggingface.co/Xenova/all-MiniLM-L6-v2) ...

PDT42 (dismissed):

postgres/lib/PostgresService.js
```diff
@@ -614,26 +619,24 @@ GROUP BY k
...
       await this.exec(`
         DROP GROUP IF EXISTS "${creds.usergroup}";
         DROP USER IF EXISTS "${creds.user}";
         CREATE GROUP "${creds.usergroup}";
         CREATE USER "${creds.user}" WITH CREATEROLE IN GROUP "${creds.usergroup}" PASSWORD '${creds.user}';
         GRANT "${creds.usergroup}" TO "${creds.user}" WITH ADMIN OPTION;
       `)
-      await this.exec(`CREATE DATABASE "${creds.database}" OWNER="${creds.user}" TEMPLATE=template0`)
-    } catch {
+      await this.exec(`CREATE DATABASE "${creds.database}" OWNER="${creds.user}" TEMPLATE=template1`)
```
PDT42, line 636: Where do people get the template?

postgres/lib/cql-functions.js
BobdenOs: People can also use the native SQL function directly.

postgres/lib/PostgresService.js
BobdenOs: Does not matter, but I agree keep it consistent.

postgres/lib/PostgresService.js
BobdenOs: system

postgres/lib/PostgresService.js
BobdenOs: cds.requires.db.credentials feeds the options credentials. By keeping it in the service it does not polute the test environment.

postgres/lib/PostgresService.js
BobdenOs: The template0 and template1 are shipped with postgres. Template0 stays clean and template1 is the default for creating new database in postgres.

postgres/lib/cql-functions.js
vkozyura: we support functions, which are supported by HANA: https://help.sap.com/docs/HANA_CLOUD_DATABASE_CN/29ae8839a4dc492788d0749181dc4df2/177f759fa7ab499a88dad2c7d92df9f9.html?locale=en-US&version=2025_1_QRC

postgres/lib/PostgresService.js
vkozyura: This is actually the PostgreSQL pgvector type name which is lowercase in SQL

postgres/lib/PostgresService.js
vkozyura: This was part of the clean-up @BobdenOs and me did in order to eliminate providing tenant parameter in credentials

postgres/lib/PostgresService.js
vkozyura: My understanding: This is intentional for multi-tenancy. @BobdenOs: correct me, if am wrong here.

postgres/lib/PostgresService.js
vkozyura: It's deterministic, hash-based, but not real ML embeddings - mostly for tests and for not crashing. I added usage of ai plugin for a real onnx embedding function - please check.

sqlite/lib/SQLiteService.js
vkozyura: old code, just moved

sqlite/lib/SQLiteService.js
vkozyura: old code, just moved

sqlite/lib/SQLiteService.js
vkozyura: 384 dimensions is hardcoded because it matches the real onnx model that's optionally loaded from ai plugin

postgres/lib/PostgresService.js
vkozyura: ai says: template0 and template1 are template databases (not just "system databases") that exist by default in every PostgreSQL installation.

sqlite/lib/SQLiteService.js
PDT42: Okay ... but: Why not just ... define functions? Why regex' with function like names? 😬

sqlite/lib/SQLiteService.js
PDT42: But, we won't be using this model forever ... but for the `hashEmbedding` placeholder, let's just ignore it for now.

postgres/lib/cql-functions.js
PDT42: Okay! So HANA only has `cosine_similarity`, but sqlite has `cosine_distance`, but we only expose the functions that HANA has.

sqlite/lib/SQLiteService.js

sqlite/lib/SQLiteService.js
```diff
@@ -36,6 +26,14 @@ class SQLiteService extends SQLService {
           const dbc = new sqlite(database, this.options.client || {})
           await dbc.ready
 
+          if (!SQLiteService._aiEmbeddingChecked) {
+            SQLiteService._aiEmbeddingChecked = true
+            try {
+              const aiPlugin = await import('@cap-js/ai/lib/vector_embedding/index.js')
```
BobdenOs: This is not an ESM module.

sqlite/lib/SQLiteService.js
```diff
@@ -46,6 +44,17 @@ class SQLiteService extends SQLService {
           dbc.function('hour', deterministic, d => d === null ? null : toDate(d, true).getUTCHours())
           dbc.function('minute', deterministic, d => d === null ? null : toDate(d, true).getUTCMinutes())
           dbc.function('second', deterministic, d => d === null ? null : toDate(d, true).getUTCSeconds())
+          dbc.function('COSINE_SIMILARITY', deterministic, (a, b) => cosineSimilarity(toFloatArray(a), toFloatArray(b)))
+          dbc.function('L2DISTANCE', deterministic, (a, b) => l2Distance(toFloatArray(a), toFloatArray(b)))
+          dbc.function('L2NORMALIZE', deterministic, v => v == null ? null : fromFloatArray(l2Normalize(toFloatArray(v)), v))
+          dbc.function('VECTOR_EMBEDDING', deterministic, (input, text_type, model_and_version) => {
```
BobdenOs: The function should simply be replaced `1:1` with the exposed embedding function from `@cap-js/ai`.

sqlite/test/vector.test.js
```diff
@@ -0,0 +1,146 @@
...
+    })
+
+    test('null returns null', async () => {
+      const res = await SELECT.from('complex.associations.Books')
+        .columns`vector_embedding(null, 'text', 'model') as embedding`
+      expect(res[0].embedding).to.eq(null)
+    })
+
+    // AI plugin specific tests
+    if (hasAIPlugin) {
```
BobdenOs: the vector function should be tested in the `@cap-js/ai` package.

postgres/lib/PostgresService.js
BobdenOs: Why not just align this with the other data types ?

sqlite/lib/SQLiteService.js
vkozyura: @BobdenOs: This is dynamic import available regardless of module system. Adv: no additional await init() is needed. I would keep it. [link](https://nodejs.org/api/esm.html#import-expressions):  It is supported in both CommonJS and ES modules, and can be used to load both CommonJS and ES modules.

sqlite/test/vector.test.js
vkozyura: functionallity is also tested in  @cap-js/ai  - here I want to see that "await import" worked. for this i temporary added devdependency to git branch of ai, which worked fine. i deleted it - we can add it again after ai feature is released.

sqlite/lib/SQLiteService.js
vkozyura: we agreed on fallback to hash impl - i still find it reasonable for those who don't want to install ai plugin.

postgres/lib/PostgresService.js
vkozyura: outdated

postgres/lib/PostgresService.js
vkozyura: done

BobdenOs (approved):

johannes-vogel (approved):

Issue #1679, opened 2026-07-14
Hana - wrong single tenant credentials auth error loop


kasimos:
### Description of erroneous behaviour

> For Cap10: When the credentials for singleTenant mode provided to the HanaService are faulty, it will now retry an infinite number of times to the db using this wrong credentials set and thereby most likely locks the db user.

> Before Cap10: When the credentials for singleTenant mode provided to the HanaService are faulty, it will try to connect once with this credentials set and then exit with "[error] - 500 - TimeoutError: ResourceRequest timed out". 

### Detailed steps to reproduce

> Use any Cap Project you have with hana profile hybrid
> 1. Change the credentials read from your env file to a user called "invalid"
> 2. cds watch --profile hybrid
> 3. it will directly run into a loop with errors

### Details about your project
| Package              | Version | Location                                                        |
 | -------------------- | ------- | --------------------------------------------------------------- |
 | @sap/cds-dk (global) | 10.0.3  | ~/.node_modules_global/lib/node_modules/@sap/cds-dk             |
 | @sap/cds-dk          | 10.0.3  | ./node_modules/@sap/cds-dk                                      |
 | @sap/cds             | 10.0.3  | ./node_modules/@sap/cds                                         |
 | @sap/cds             | 10.0.3  | ./node_modules/@sap/cds-dk/node_modules/@sap/cds                |
 | @sap/cds-compiler    | 7.0.1   | ./node_modules/@sap/cds-dk/node_modules/@sap/cds-compiler       |
 | @sap/cds-compiler    | 7.0.1   | ./node_modules/@sap/cds-compiler                                |
 | @sap/cds-fiori       | 2.3.0   | ./node_modules/@sap/cds-dk/node_modules/@sap/cds-fiori          |
 | @sap/cds-fiori       | 2.3.0   | ./node_modules/@sap/cds-fiori                                   |
 | @sap/cds-mtxs        | 3.9.5   | ./node_modules/@sap/cds-dk/node_modules/@sap/cds-mtxs           |
 | @cap-js/asyncapi     | 1.1.0   | ./node_modules/@sap/cds-dk/node_modules/@cap-js/asyncapi        |
 | @cap-js/db-service   | 3.0.1   | ./node_modules/@cap-js/hana/node_modules/@cap-js/db-service     |
 | @cap-js/db-service   | 2.11.0  | ./node_modules/@cap-js/db-service                               |
[...]

PR #1680, opened 2026-07-14
fix: wrong hana single tenant credential connection loop

kasimos:
Addresses Ticket #1679 

Now auth errors in single tenant hana connections will be logged in the debug log and throw an appropriate error, thereby no longer running into a loop and locking the db user

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -59,7 +59,11 @@ class HANAService extends SQLService {
         service.server.major = dbc.server.major || service.server.major
         return dbc
       } catch (err) {
-        if (err.code === 10) return // authentication error, see error handler below
+        if (err.code === 10) {
+          // authentication error
+          LOG.debug('connection failed:', err)
+          throw new Error(`Pool failed connecting`, { cause: err })
+        } 
         if (attempt < maxRetries) {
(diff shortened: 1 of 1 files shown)


hana/lib/HANAService.js
```diff
@@ -59,7 +59,11 @@ class HANAService extends SQLService {
         service.server.major = dbc.server.major || service.server.major
         return dbc
       } catch (err) {
-        if (err.code === 10) return // authentication error, see error handler below
+        if (err.code === 10) {
+          // authentication error
+          LOG.debug('connection failed:', err)
+          throw new Error(`Pool failed connecting`, { cause: err })
+        } 
         if (attempt < maxRetries) {
```
johannes-vogel, line 67:
```suggestion
        if (err.code !== 10 && attempt < maxRetries) {
```

I think the error just needs to bubble up. There's no need to log or wrap it.

kasimos: This works except for the initial connection when starting a CAP app. There the proposed change will throw:
"[queue] - Failed to fetch targets for flush Error: Pool resource could not be acquired within 10s"
For requests that are after that hitting the service it will correctly show: "[odata] - 500 - SqlError {message: "authentication failed..."

johannes-vogel: I've created a new PR that has a better behavior: #1684. This should include the db error in the response and log.

Issue #1681, opened 2026-07-14
Hana - Connection slowly creeps up memory and closes super slow with many statements prepared on it


kasimos:
### Description of erroneous behaviour

When using cds queries in services a new statement with the generated sql will be passed to the prepare method and stored in an array. This does not reuse any previous preparations and will eventually run into Hana sql error: "ERROR [SQL-608]: "the number of prepared statements per connection cannot exceed 100000."
see Sap Note 3031816 

Even when it doesn't reach it, the array size and therefore the memory usage will slowly creep up and when cleaning up the connection to put it back to the pool this will close all prepared statements one after the other and thereby block the main event loop of nodejs.

The suggested fix will use a Map instead of the Array and make sure Procedure calls are exempt from this, as some built-in procedures (e.g. for Application locks) do not work with a previous prepared statement.
The execution time also goes down greatly.
For a simple 10k selects it in my case from around 10s to 5s and a nearly instant return to the pool once finished. Previously the closing also took around 10s and was blocking. 

Please note that a similar fix would be possible for sqlite as well. For Postgres it depends on the used driver as some are already doing this preparation caching for you.

### Detailed steps to reproduce

> Use a Hana CAP Project
> 1. Add a service handler calling a sql select with a where clause in a loop of 100k times
> 2. Start with cds watch --profile hybrid
> 3. Call your service endpoint 

### Details about your project
| Package              | Version | Location                                                        |
 | -------------------- | ------- | --------------------------------------------------------------- |
 | @sap/cds-dk (global) | 10.0.3  | ~/.node_modules_global/lib/node_modules/@sap/cds-dk             |
 | @sap/cds-dk          | 10.0.3  | ./node_modules/@sap/cds-dk                                      |
 | @sap/cds             | 10.0.3  | ./node_modules/@sap/cds                                         |
 | @sap/cds             | 10.0.3  | ./node_modules/@sap/cds-dk/node_modules/@sap/cds                |
 | @sap/cds-compiler    | 7.0.1   | ./node_modules/@sap/cds-dk/node_modules/@sap/cds-compiler       |
 | @sap/cds-compiler    | 7.0.1   | ./node_modules/@sap/cds-compiler                                |
 | @sap/cds-fiori       | 2.3.0   | ./node_modules/@sap/cds-dk/node_modules/@sap/cds-fiori          |
 | @sap/cds-fiori       | 2.3.0   | ./node_modules/@sap/cds-fiori                                   |
[...]

PR #1682, opened 2026-07-14
fix: hana preparations use map

kasimos:
Addresses #1681 

Use a Map to store prepared statements and reuse them when possible.
Add a unique comment for procedure calls to prevent reuse for those, as some builtin procedures like acquiring application locks do not work with already prepared statements.

Please note that a similar fix would be possible for sqlite as well. For Postgres it depends on the used driver as some are already doing this preparation caching for you.

### DIFF
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -342,7 +342,7 @@ class HANAService extends SQLService {
       const newPS = this.ensureDBC().prepare(sql, hasBlobs)
       this.ensureDBC().statements.set(sql, newPS)
       stmt = newPS
-      }
+    }
     return stmt
   }
 
--- a/hana/lib/HANAService.js
+++ b/hana/lib/HANAService.js
@@ -335,9 +335,14 @@ class HANAService extends SQLService {
 
   // prepare and exec are both implemented inside the drivers
   prepare(sql, hasBlobs) {
-    const stmt = this.ensureDBC().prepare(sql, hasBlobs)
-    // we store the statements, to release them on commit/rollback all at once
-    this.dbc.statements.push(stmt)
+    // we store the prepared statements for caching
+    // and to release them on commit/rollback all at once
+    let stmt = this.ensureDBC().statements.get(sql);
+    if (!stmt) {
+      const newPS = this.ensureDBC().prepare(sql, hasBlobs)
+      this.ensureDBC().statements.set(sql, newPS)
+      stmt = newPS
+      }
     return stmt
   }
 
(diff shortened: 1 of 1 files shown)


BobdenOs:
Duplicate from: https://github.com/cap-js/cds-dbs/pull/1496

hana/lib/HANAService.js
```diff
@@ -335,9 +335,14 @@ class HANAService extends SQLService {
 
   // prepare and exec are both implemented inside the drivers
   prepare(sql, hasBlobs) {
-    const stmt = this.ensureDBC().prepare(sql, hasBlobs)
-    // we store the statements, to release them on commit/rollback all at once
-    this.dbc.statements.push(stmt)
+    // we store the prepared statements for caching
+    // and to release them on commit/rollback all at once
+    let stmt = this.ensureDBC().statements.get(sql);
```
BobdenOs, line 340: This is fundamentally flawed, because it is possible to run the same `sql` statement in "parallel" at this level which will start throwing errors when the same statement instance is called with `exec` / `execute` as the statement has state. Therefor it is required to create a queue for each statement until it is finished with an `exec` / `execute`.

hana/lib/HANAService.js
```diff
@@ -1358,7 +1363,8 @@ SELECT ${mixing} FROM JSON_TABLE(SRC.JSON, '$' COLUMNS(${extraction}) ERROR ON E
   async onCall({ query, data }, name, schema) {
     const isAsync = /\sASYNC\s*$/.test(query)
     const outParameters = isAsync ? [{ PARAMETER_NAME: 'ASYNC_CALL_ID' }] : await this._getProcedureMetadata(name, schema)
-    const ps = await this.prepare(query)
+    // make procedure preparations unique to avoid caching issues with internal procedures
+    const ps = await this.prepare(query+'--'+this.ensureDBC().statements.length)
```
BobdenOs, line 1367: By changing the `SQL` it does not just miss the local statement cache, but it will also miss the HANA execution plan cache. Which means that this can cause a dramatic overall response time impact as this force HANA to fully parse through the statement and all its dependencies which depends upon the complexity of the procedure definition.

BobdenOs (changes requested):
This approach is insufficient. We already have a PR which solves the problem https://github.com/cap-js/cds-dbs/pull/1496

currently there is a bug in the `hdb` driver that prevents a connection to re use a prepared statement after an session variable is updated. Once that is resolved it will be possible to merge the proper fix.

kasimos:
Thanks, somehow I didn't find your PR when we encountered this. Sorry for that. closing the PR and moving discussion over to #1496

PR #1693, opened 2026-07-28
fix: default fields not filled when not exposed

larsPlessing:
On an INSERT via a projection that doesn't expose them, fields with default values (e.g. managed fields) were not populated.

### DIFF
--- a/db-service/lib/cqn2sql.js
+++ b/db-service/lib/cqn2sql.js
@@ -919,7 +919,7 @@ class CQN2SQLRenderer {
       this.entries = [[...this.values, stream]]
     }
 
-    const extractions = this._managed = this.managed(columns.map(c => ({ name: c })), { ...transitions.target.elements, ...elements, })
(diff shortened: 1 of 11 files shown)


db-service/lib/cqn2sql.js
```diff
@@ -919,9 +919,9 @@ class CQN2SQLRenderer {
       this.entries = [[...this.values, stream]]
     }
 
-    const extractions = this._managed = this.managed(columns.map(c => ({ name: c })), elements)
-    return (this.sql = `INSERT INTO ${this.quote(entity)}${alias ? ' as ' + this.quote(alias) : ''} (${this.columns.map(c => this.quote(transitions.mapping.get(c)?.ref?.[0] || c))
-      }) SELECT ${extractions.slice(0, columns.length).map(c => c.insert)} FROM json_each(?)`)
+    const extractions = this._managed = this.managed(columns.map(c => ({ name: c })), { ...transitions.target.elements, ...elements }).filter(e => !transitions.mapping.get(e.name)?.ref[1])
```
BobdenOs: What is this filter supposed to be doing ?

PR #1695, opened 2026-07-31
fix(hana): propagate error codes

swaldmann:


hana/lib/HANAService.js
```diff
@@ -83,8 +83,11 @@ class HANAService extends SQLService {
         return dbc
       } catch (err) {
         if (!cds.env.features.use_generic_pool) {
-          if (err.status === 404 || err.status === 429) {
-            throw new Error(`Pool failed connecting to '${tenant}'`, { cause: err })
+          if (err.status === 404) {
+            cds.error(404, `Pool failed connecting: Tenant '${tenant}' not found (${err.message})`, { cause: err })
```
johannes-vogel: Is there a need to wrap the errors?

swaldmann: Nope, changed it in https://github.com/cap-js/cds-dbs/pull/1695/commits/c1657821c6aa2a28282f0a94a5aac984cae210c7

johannes-vogel (approved):

PR #1699, opened 2026-08-03
fix(hana): fix two object-mode expand streaming crashes

nils:
Fixes the two bugs introduced as failing tests in the base PR #1698.

  **Bug 1** — `lib/drivers/base.js`

  `handleLevel` called `delete level.expands[property]` then read the same key
  back at the `result` field of the pushed expand level — yielding `undefined`.
  `rsNextObjectMode` then called `undefined.push(row)`.

  Fix: capture the value before deleting the key, use it for `result`.

  **Bug 2** — `lib/drivers/hdb.js`

  `rsIterator`'s `next()` eagerly fires `this._prefetch = raw.next()`. When the
  consumer pauses between rows the HANA ResultSet can close mid-stream. The
  stashed prefetch promise rejects with `ERR_STREAM_PREMATURE_CLOSE` and, because
  it is never awaited, crashes the process as an unhandledRejection.

  Fix: attach a no-op `.catch()` to `this._prefetch` so the rejection is handled
  when the promise is never consumed. The value is still re-fetched normally via
  `ret` on the next call.

Confirmed against a live HANA instance/data model. 
The tests themselves have not been executed against a live HANA yet, I'm waiting for the pipeline to do that.


test/compliance/SELECT.test.js
```diff
@@ -1221,6 +1221,60 @@ describe('SELECT', () => {
...
+      const cqn = SELECT([{ ref: ['ID'] }, { ref: ['name'] }, { ref: ['books'], expand: ['*'] }]).from(Authors)
+
+      const expected = await cqn.clone()
+
+      const rows = []
+      for await (const row of cqn.clone()) rows.push(row)
+
+      expect(rows).deep.eq(expected)
+      expect(rows[0].books).to.be.an('array')
+      expect(rows[0].books[0].title).to.equal('Wuthering Heights')
```
BobdenOs, line 1236: Order problem in the expect as the query does use `order by` or equivalent. (applies to all tests)

S3lected:
Hi colleagues,
If you need any further information, please feel free to reach out to us on Teams. This is very important for our RTC.

Thanks!

Best regards,
Matthias Kiefer

test/compliance/SELECT.test.js
```diff
@@ -1221,6 +1221,60 @@ describe('SELECT', () => {
...
+      const cqn = SELECT([{ ref: ['ID'] }, { ref: ['name'] }, { ref: ['books'], expand: ['*'] }]).from(Authors)
+
+      const expected = await cqn.clone()
+
+      const rows = []
+      for await (const row of cqn.clone()) {
+        // Simulate a slow per-row async side-effect (e.g. writing to object storage).
+        // This pauses the async iterator while the HANA ResultSet is still open,
+        // triggering the prefetch orphan on the next read cycle.
+        await new Promise((resolve) => setTimeout(resolve, 50))
```
BobdenOs, line 1272: If `setTimeout(0)` is enough then please use that. Using `sleep` inside a test is bad practice.

test/compliance/SELECT.test.js
```diff
@@ -1221,6 +1221,60 @@ describe('SELECT', () => {
...
+      const { Authors } = cds.entities('complex.associations')
+      const cqn = SELECT([{ ref: ['ID'] }, { ref: ['name'] }, { ref: ['books'], expand: ['*'] }]).from(Authors)
+
+      const expected = await cqn.clone()
+
+      const rows = []
+      for await (const row of cqn.clone()) {
+        // Simulate a slow per-row async side-effect (e.g. writing to object storage).
+        // This pauses the async iterator while the HANA ResultSet is still open,
+        // triggering the prefetch orphan on the next read cycle.
```
BobdenOs, line 1271: The prefetch promise would have to still throw an error. What would be the trigger for the error of the prefetch ?

test/compliance/SELECT.test.js
```diff
@@ -1221,6 +1221,60 @@ describe('SELECT', () => {
...
+    test('async iterator with a paused consumer', () => cds.tx(async () => {
+      // Regression: the hdb driver eagerly prefetches the next chunk on every
+      // `next()`. If the consumer pauses between rows (e.g. awaiting an async
+      // side-effect like writing each row to storage), the HANA ResultSet can
+      // close mid-stream and the orphaned prefetch promise rejects with
+      // ERR_STREAM_PREMATURE_CLOSE — previously surfacing as an unhandled
+      // rejection. Use an expand query so the hdb rsIterator object-mode path
+      // is active (the affected code path), and introduce a real async pause
+      // between rows to let the underlying ResultSet settle while the consumer
+      // is suspended.
```
BobdenOs, line 1261: `test/compliance/SELECT.test.js` is not just for `hana`. So if you need to make a regression test for HANA specific limitations it is probably best to move them to the `hana` folder. Or if you do not see the need to keep this specific regression test it would also be fine for me to remove it once it was confirmed that the fix works.

nils:
I think @johannes-vogel via #1705 has a better and more complete approach for testing and fixing these issues.

PR #1704, opened 2026-08-05
feat: use ai plugin optionally for onnx vector embedding

vkozyura:


sqlite/lib/SQLiteService.js
```diff
@@ -26,6 +26,14 @@ class SQLiteService extends SQLService {
           const dbc = new sqlite(database, this.options.client || {})
           await dbc.ready
 
+          if (!SQLiteService._aiEmbeddingChecked) {
+            SQLiteService._aiEmbeddingChecked = true
+            try {
+              const aiPlugin = await import('@cap-js/ai/lib/vector_embedding/index.js')
```
sjvans, line 32: don't require files

PDT42: With the updated path, this check fails in  know-cap PR#14 with "Unsupported Directory Import". Does `await import(...)` resolve `index.js` automatically?

vkozyura: the part "// AI plugin specific tests" in vector.test.js runs and shows that onnx model was loaded

sqlite/lib/SQLiteService.js
```diff
@@ -39,7 +47,14 @@ class SQLiteService extends SQLService {
...
           dbc.function('L2NORMALIZE', deterministic, v => v == null ? null : fromFloatArray(l2Normalize(toFloatArray(v)), v))
-          dbc.function('VECTOR_EMBEDDING', deterministic, (input, text_type, model_and_version) => input == null ? null : JSON.stringify(hashEmbedding(String(input))))
+          dbc.function('VECTOR_EMBEDDING', deterministic, (input, text_type, model_and_version) => {
+            if (input == null) return null
+            if (SQLiteService._aiEmbedding) {
+              try { return SQLiteService._aiEmbedding(input, text_type, model_and_version) }
+              catch { /* fallback to hash */ }
+            }
+            return JSON.stringify(hashEmbedding(String(input)))
+          })
```
BobdenOs, line 57: Having a silent fallback to `hash` is bad. Either the `_aiEmbedding` function is available and will always be used or the hash function is used. Having a wrapper function with `try / catch` and checking on each call is going to cause a lot of issues that a simple `throw` will solve.

BobdenOs (changes requested):

sjvans:
Superseded by cap-js/ai#46 and the explicit ai-sqlite service kind.

With this approach, @cap-js/sqlite keeps its deterministic hash VECTOR_EMBEDDING implementation unchanged. @cap-js/ai owns ONNX initialization and overrides VECTOR_EMBEDDING only when consumers select kind: ai-sqlite. This avoids optional package discovery and silent fallback from semantic inference to hash output.

The ai-sqlite integration tests in cap-js/ai cover the three-argument form, the optional fourth argument, and SQL NULL semantics, so no cds-dbs production or cross-package test change remains for this PR.

PR #1707, opened 2026-08-10
fix: include reuse-module initial data in @cap-js/postgres build

### LINKED ISSUE #1694: @cap-js/postgres build omits reuse-module (dependency) initial data from gen/pg

The `@cap-js/postgres` build task drops **initial data shipped by reuse packages** (dependencies) from the generated `gen/pg` deployer artifact. It copies **only** the app-local `db/data` folder, so any `db/data/*.csv` provided by a dependency is missing after `cds build`.

This diverges from two other code paths that *do* include reuse-module data:

- **Runtime deploy** (`cds deploy` from source) reads initial-data CSVs from **all** model sources (app + reuse modules) and concatenates them per entity.
- **The HANA build task** reproduces that in its artifact via `csvFileDetection` (default on), which enumerates resources through `cds.deploy.resources(model)`.

The Postgres build task has no equivalent, so its artifact is inconsistent with both.

This is especially harmful for the recommended **“extend a reuse code list with only the new keys”** pattern. Given a reuse package that ships a code list:

```csv
# node_modules/@some/reuse/db/data/my.CodeList.csv
code,name
A,Alpha
B,Beta
C,Gamma
```

and a consumer that adds only the new key (as documented, to avoid `UNIQUE constraint` on the shared PK):

```csv
# db/data/my.CodeList.csv
code,name
D,Delta
```

`cds deploy` from source yields **A, B, C, D** (correct). But the Postgres build artifact contains only the consumer’s file:

```csv
# gen/pg/db/data/my.CodeList.csv
code,name
D,Delta
```

so a fresh Postgres deployment is missing the base rows **A, B, C**.

**Root cause** — `@cap-js/postgres/cds-plugin.js`, `PostgresBuildPlugin.build()` copies a single local folder:

```js
let data
if (fs.existsSync(path.join(this.task.src, 'data'))) {
  data = 'data'
} else if (fs.existsSync(path.join(this.task.src, 'csv'))) {
  data = 'csv'
}
if (data) {
  promises.push(this.copy(data).to(path.join('db', 'data')))
}
```

There is no reuse-module discovery. `cds.build.tasks[*].options.data` is also ignored by this plugin, so it can’t be used as a workaround.

**Suggested fix** — collect initial data the same way the runtime/HANA path does, e.g. iterate `await cds.deploy.resources(model)` (which already returns every seed CSV across app + reuse modules) and write the merged union per entity into `gen/pg/db/data`, ideally gated by the same `csvFileDetection` option for parity with HANA.

1. Create a reuse package `@some/reuse` shipping `db/data/my.CodeList.csv` with rows `A,B,C`.
2. In a consumer CAP app, add the dependency and `db/data/my.CodeList.csv` containing only the new key `D`.
3. `cds deploy --to sqlite:test.db` → `SELECT code FROM my_CodeList;` returns **A, B, C, D** (correct).
4. `cds build --production` → inspect `gen/pg/db/data/my.CodeList.csv` → contains **only `D`**.
5. Deploy `gen/pg` to a fresh Postgres DB → `my.CodeList` is missing base rows **A, B, C**.

| Your Project Name | (internal) |
|:------------------|------------|
| OData version     | v4         |
| Node.js version   | v22.23.2   |
| @sap/cds          | 9.9.3      |
| @sap/cds-compiler | 6.9.4      |
| @sap/cds-dk       | 9.9.4      |
| @cap-js/db-service | 2.11.0    |
| @cap-js/postgres  | 2.3.0      |
| @cap-js/sqlite    | 2.4.0      |

alex-bkmr-cbs:
The postgres build task copied only the app-local db/data folder, dropping
initial-data CSVs shipped by reuse packages from the gen/pg artifact. This made
a fresh Postgres deployment inconsistent with `cds deploy` from source and with
the HANA build task.

Collect initial data across app + reuse modules via cds.deploy.resources() and
write the merged per-entity union into gen/pg/db/data. Gated by a csvFileDetection
option (default on) for parity with HANA; setting it to false restores the legacy
local-only copy.

Closes #1694

Signed-off-by: Alex <alexander.buchkremer@cbs-consulting.de>

### DIFF
--- a/postgres/cds-plugin.js
+++ b/postgres/cds-plugin.js
@@ -65,15 +65,80 @@ cds.build?.register?.('postgres', class PostgresBuildPlugin extends cds.build.Pl
 
     promises.push(this.write(cds.compile.to.json(model)).to(path.join('db', 'csn.json')))
 
-    let data
-    if (fs.existsSync(path.join(this.task.src, 'data'))) {
-      data = 'data'
-    } else if (fs.existsSync(path.join(this.task.src, 'csv'))) {
-      data = 'csv'
-    }
-    if (data) {
-      promises.push(this.copy(data).to(path.join('db', 'data')))
+    // csvFileDetection (default on, for parity with the HANA build task) collects initial
+    // data from all model sources - app-local db/data AND every reuse module's db/data - so
+    // the deployer artifact matches `cds deploy` from source. When disabled, only the
+    // app-local data/csv folder is copied (legacy behaviour).
+    const csvFileDetection = this.task.options?.csvFileDetection ?? true
+    if (csvFileDetection && typeof cds.deploy?.resources === 'function') {
+      for (const [dest, sources] of await collectInitialData(model)) {
+        const to = path.join('db', 'data', dest)
+        if (sources.length === 1) {
+          promises.push(this.copy(sources[0]).to(to))
+        } else {
+          promises.push(this.write(mergeCsvFiles(sources)).to(to))
+        }
+      }
+    } else {
+      let data
+      if (fs.existsSync(path.join(this.task.src, 'data'))) {
+        data = 'data'
+      } else if (fs.existsSync(path.join(this.task.src, 'csv'))) {
+        data = 'csv'
+      }
+      if (data) {
+        promises.push(this.copy(data).to(path.join('db', 'data')))
+      }
     }
     return Promise.all(promises)
   }
 })
+
+// Discover initial-data resources across app + reuse modules, grouped by their
+// destination file name. Files that resolve to the same destination (e.g. a code list
+// extended by a reuse module and the consumer) are merged; init.js/ts is skipped as it
+// cannot be reproduced from the artifact's db/data folder.
+async function collectInitialData (model) {
+  const resources = await cds.deploy.resources(model)
+  const byDest = new Map()
+  for (const [file, entity] of Object.entries(resources)) {
+    if (entity === '*') continue // init.js/ts
+    const dest = path.basename(file)
+    const group = byDest.get(dest)
+    if (group) group.push(file)
+    else byDest.set(dest, [file])
+  }
+  return byDest
+}
+
+// Merge several CSV files for the same entity into the union of their rows under a unified
+// header. Files are discovered reuse-module-first, so base rows precede consumer rows.
+function mergeCsvFiles (files) {
+  const columns = []
+  const seen = new Set()
+  const records = []
+  for (const file of files) {
+    const [header, ...rows] = cds.parse.csv(fs.readFileSync(file, 'utf8'))
+    if (!header) continue
+    for (const column of header) if (!seen.has(column)) { seen.add(column); columns.push(column) }
+    for (const row of rows) {
+      const record = {}
+      header.forEach((column, i) => { record[column] = row[i] })
+      records.push(record)
+    }
+  }
+  const lines = [columns.map(csvEscape).join(',')]
+  const emitted = new Set()
+  for (const record of records) {
+    const line = columns.map(column => csvEscape(record[column] ?? '')).join(',')
+    if (emitted.has(line)) continue
+    emitted.add(line)
+    lines.push(line)
+  }
+  return lines.join('\n') + '\n'
+}
+
+function csvEscape (value) {
+  value = String(value)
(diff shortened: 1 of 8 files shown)


postgres/cds-plugin.js
patricebender, line 1: @chgeo / @swaldmann I assume that you have the same problem for the HANA build task. How do you merge the data there, given that initial data could be provided via `yaml`, `json` and not only `csv`. Could we somehow re-use your work in this build plugin for postgres rather than to re-invent the wheel?

chgeo: > How do you merge the data there, given that initial data could be provided via `yaml`, `json` and not only `csv`

Only csv files and handled by the HANA build.  No support for other data formats so far.
The csv files themselves are not merged but just handed over to the HDI deployer.  We have never wanted to merge data on a per-record basis, for build performance reasons, and to avoid confusion on which data gets effectively deployed.

patricebender (commented):
@alex-bkmr-cbs thanks for your contriubtion, as Christian stated, the CAP build tasks for other DBs do not merge initial data per-record, for build performance reasons and to avoid confusion over what actually gets deployed. The PR should therefore drop the per-record CSV-merge code and instead follow how the HANA build task already does it in `@sap/cds-dk` (`lib/build/plugins/hana/index.js`, `_copyResourcesExt`, lines 134–163): discover via `cds.deploy.resources(model)`, key by `path.basename` in a Map, `.reverse()` so app resources win over reuse-module files, and copy each file as-is.

<details>

<summary> @sap/cds-dk coding </summary>



```js
    /**
     * Copies all files located at <src> (except HANA artifacts not contained in <db>/src/**) to the folder <dest>.
     * '*.csv' files are read based on the corresponding CDS model file location and copied as flat list into folder '<dest>/src/gen>'.
     *
     * @param {string} src
     * @param {string} dest
     * @param {Object} model
     */
    async _copyResourcesExt(src, dest, model) {
        const resources = Object.keys(await cds.deploy.resources(model)).reverse() // reverse to get reuse resources first and app resources last
        const dbSrc = path.join(src, 'src')

        // 1. copy csv files into 'src/gen/data' or 'src/gen/csv' subfolder
        if (resources.length) {
            // determine the CSV folder from the location of the CSVs defined in this build task db folder, use subfolder data by default
            const dbCSVs = resources.filter(res => res.startsWith(src) && !res.startsWith(dbSrc))
            let csvFolder = this.getBuildOption(CSV_FILE_TARGET)
            if (csvFolder && !["data", "csv"].includes(csvFolder)) {
                throw new BuildError("Invalid value for option 'csvFileTarget'. Valid values are 'data' or 'csv'.")
            }
            if (!csvFolder) {
                const csvPath = path.join(src, "data")
                // determine subfolder name used by the application for backward compatibility
                csvFolder = !dbCSVs.length || dbCSVs.some(res => res && res.startsWith(csvPath)) ? "data" : "csv"
            }

            // preserve last-write-wins semantics for duplicate basenames while parallelizing distinct file copies
            const csvResourcesByName = new Map()
            for (const res of resources) {
                // do not duplicate resources that are already contained in db/src/**
                if (res && /\.csv$/.test(res) && !res.startsWith(dbSrc)) {
                    csvResourcesByName.set(path.basename(res), res)
                }
            }
            await Promise.all(Array.from(csvResourcesByName.entries()).map(([fileName, res]) => {
                return this.copy(res).to(path.join(this.task.options.compileDest, csvFolder, fileName))
            }))
        }

        if (this.isStagingBuild()) {
            let blockList = "\\.cds$|\\.csv$|\\.hdbtabledata$"
            blockList += !this.hasBuildOption(CONTENT_ENV, true) ? "|\\.env($|\\..*$)" : ""
            blockList += !this.hasBuildOption(CONTENT_DEFAULT_ENV_JSON, true) ? "|default-env\\.json$" : ""
            blockList = new RegExp(blockList)

            // 2. staging build: copy files except *.cds, .env, default-env.json, ./node_modules/**
            await this.copyNativeContent(src, dest, (entry) => {
                if (entry.startsWith(dbSrc)) {
                    return true
                }
                if (isdir(entry)) {
                    return !/(\/|\\)node_modules(\/|\\)?$/.test(entry) ||
                        (/(\/|\\)node_modules(\/|\\)?$/.test(entry) && this.hasBuildOption(CONTENT_NODE_MODULES, true))
                }
                return !blockList.test(entry)
            })
        }
    }
```

</details>

PR #1712, opened 2026-08-24
fix(cqn4sql): reject overly nested query expressions instead of crashing

patricebender:
Deeply nested input (subqueries, xpr, association paths, inline/expand,   functions/lists) recursed until the stack overflowed, crashing with a 500.

Add a shared depth budget across infer and cqn4sql, raising a 400 instead.

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -33,6 +33,23 @@ const notSupportedOps = [['>'], ['<'], ['>='], ['<='], ['*'], ['+'], ['-'], ['/'
 
 const allOps = eqOps.concat(eqOps).concat(notEqOps).concat(notSupportedOps)
 
+
+// Wraps a recursive fn so its live depth is bounded, throwing a catchable 400
+// at the limit. Depth is per wrapper instance, i.e. per `cqn4sql` call.
+const MAX_RECURSION_DEPTH = 300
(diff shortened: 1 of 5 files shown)


db-service/lib/cqn4sql.js
```diff
@@ -56,6 +73,16 @@ const { pseudos } = require('./infer/pseudos')
...
+  // guard the input-driven recursive transforms against stack overflow from
+  // nested input
+  // eslint-disable-next-line no-func-assign
+  ;[getTransformedTokenStream, getFlatColumnsFor, nestedProjectionOnStructure, expandColumn, transformSubquery] = [
+    getTransformedTokenStream,
+    getFlatColumnsFor,
+    nestedProjectionOnStructure,
+    expandColumn,
+    transformSubquery,
+  ].map(depthGuarded)
```
BobdenOs: Is my understanding correct that each one of these functions can be called `MAX_RECURSION_DEPTH ` / `300` times before they return. So when there are combinations between these functions calling each other they will still be able to hit an `300` / `600` / `900` stacks size ?

patricebender: Yeah I think your suspicion is right. I work on some more test cases first and then think about a better way to catch this.

BobdenOs (approved):

PR #1716, opened 2026-08-26
fix: exclude cds.Vector from wildcard select expansion

sjvans:
## Motivation

CAP Java excludes `Vector`-typed elements from the default select list (i.e. a `SELECT *` or a SELECT with no explicit column list). This PR aligns the Node.js runtime to the same behaviour by treating `cds.Vector` identically to `cds.LargeBinary` at the four wildcard-expansion sites in `db-service`.

Vector columns can be large (high-dimensional float arrays) and are rarely wanted in default reads, just like binary blobs.

## What changed

Four wildcard-expansion sites were updated:

| File | ~Line | Change |
|---|---|---|
| `db-service/lib/cqn4sql.js` | 1401 | `getColumnsForWildcard`: skip `cds.Vector` alongside `cds.LargeBinary` |
| `db-service/lib/cqn4sql.js` | 1479 | `getFlatColumnsFor`: return `[]` for `cds.Vector` when `isWildcard` |
| `db-service/lib/infer/index.js` | 1098 | single-source wildcard: filter out `cds.Vector` |
| `db-service/lib/infer/index.js` | 1118 | multi-source wildcard: same filter |

## Behaviour

- **Top-level `{ * }` or column-less SELECT**: Vector elements are **excluded** (same as LargeBinary).
- **Association expand `assoc { * }`**: Vector elements are **excluded** in the correlated expand subquery too — the expand routes through the infer wildcard path, so no separate change was needed at the expand-to-join site (`~L962`).
- **Inline struct wildcard `struct.{ * }`**: Vector elements are **kept** — this mirrors the pre-existing LargeBinary behaviour for nested wildcards (see existing test `inline wildcard does not ignore large binaries`).
- **Explicit column selection** (`SELECT { embedding }`): unchanged — still works.

## Tests

- New entity `WithVector` (including a self-referencing association `self`) added to `db-service/test/bookshop/db/schema.cds`.
- New tests in `db-service/test/cqn4sql/wildcards.test.js`:
  - `top-level wildcard excludes Vector and LargeBinary` — asserts `{ * }` and implicit-star both expand to only `ID, name, self_ID`.
  - `inline wildcard does not ignore vectors` — asserts `struct.{ * }` keeps `struct_embedding` and `struct_deepImage`.
  - expand coverage — asserts `self { * }` expands to only `$s.ID, $s.name, $s.self_ID` (Vector/LargeBinary excluded in the correlated subquery).
- All 27 tests in `wildcards.test.js` pass; 0 regressions.

## Open question for reviewers

- **Deep-op blob re-add** (`cqn4sql.js` ~L227): blobs are re-added in certain deep-operation paths. Should Vector get the same treatment? Deliberately left unchanged in this draft.

---

_Draft — opened for discussion before merging._


db-service/test/cqn4sql/wildcards.test.js
```diff
@@ -356,6 +356,38 @@ describe('wildcard expansion and exclude clause', () => {
...
+    )
+    const expected = cds.ql`SELECT from bookshop.WithVector as WithVector {
+      WithVector.ID,
+      WithVector.name
+    }`
+    expect(starExpansion).to.deep.equal(expected)
+    expect(noColumnsExpansion).to.deep.equal(expected)
+  })
+
+  it('inline wildcard does not ignore vectors', () => {
```
patricebender, line 377: we could also extend the entity with a self referencing association `self` which is then in the test using an expand `self { * }` to make sure this case is also covered in the correlated expand subquery

sjvans: Good call — added `self : Association to WithVector` to the test entity and a `self { * }` expand test. Turns out the correlated expand subquery already routes through the infer wildcard path, so `cds.Vector` (and `LargeBinary`) are excluded there too — the expand subquery expands to `$s.ID, $s.name, $s.self_ID`, no change needed at the expand-to-join site (`~L962`). Pushed in `bf407fbb`.

patricebender (approved):

PR #1717, opened 2026-08-27
fix: correlate $search ranking ORDER BY to the outer row

patricebender:
The deep-search ranking ORDER BY reused the search expression without binding it to the outer query row, so the sub-select produced a `key IN (key)` tautology instead of a correlation. By defering the ranking `ORDER BY` to after `infer()`, where the outer alias is known, the score sub-select can be correlated to the outer row post-transform (mirroring expand's _correlate). A deep search may lead to one outer row yielding many joined child rows, so wrap the score in MAX() to keep the scalar sub-select single-valued and rank by the best-matching child.

### DIFF
--- a/db-service/lib/cqn4sql.js
+++ b/db-service/lib/cqn4sql.js
@@ -82,17 +82,9 @@ function _cqn4sql(originalQuery, model, useTechnicalAlias = true) {
       const { where, having } = transformSearch(searchTerm)
       if (where) inferred.SELECT.where = where
       else if (having) inferred.SELECT.having = having
-      if (searchTerm.func) (inferred.SELECT.orderBy ??= []).unshift({ func: searchTerm.func, args: [...searchTerm.args, true], sort: 'desc' })
-      else if (searchTerm.xpr) {
-        const searchSelect = searchTerm.xpr[2]
-        const searchFunc = searchSelect.SELECT.where[0]
-          ; (inferred.SELECT.orderBy ??= []).unshift({
-            __proto__: SELECT.from(searchSelect.SELECT.from)
-              .columns({ func: searchFunc.func, args: [...searchFunc.args, true] })
-              .where([searchTerm.xpr[0], 'in', { list: searchSelect.SELECT.columns }]), // TODO: <-- ensure that the sub select in the order by is bound to the original query result row
-            sort: 'desc'
-          })
-      }
+      // Defer the ranking ORDER BY to after infer(), where the outer table alias is known and the
+      // deep-search sub-select can be correlated to the outer row (see buildSearchRankOrderBy).
+      defineProperty(inferred, '$searchRank', searchTerm)
     }
   }
   // query modifiers can also be defined in from ref leaf infix filter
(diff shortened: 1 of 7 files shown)


hana/test/fuzzy.test.js
```diff
@@ -74,8 +75,8 @@ describe('search', () => {
       const { Books } = cds.entities('sap.capire.bookshop')
       const cqn = SELECT.from(Books).search('"autobio"', '"Jane"').columns('1')
       const { sql, values } = cds.db.cqn2sql(cqn)
-      // 5 columns to be searched createdBy, modifiedBy, title, descr, currency_code
-      expect(sql.match(/(like)/g).length).to.eq(10)
+      // 5 searched columns × 2 terms, in WHERE and again in the ranking ORDER BY
```
johannes-vogel: The rank doesn't make sense for fuzzy=false

db-service/lib/cqn4sql.js
```diff
@@ -2608,6 +2608,72 @@ function _cqn4sql(originalQuery, model, useTechnicalAlias = true) {
...
+      if (i) where.push('and')
+      // seeded unqualified on both sides; correlateSearchRank redirects the rhs to the outer row
+      where.push({ ref: [...innerKeys[i].ref] }, '=', { ref: [...innerKeys[i].ref] })
+    }
+
+    const entry = {
+      __proto__: SELECT.from(searchSelect.SELECT.from)
+        // one correlated outer may lead to many joined child rows -> MAX collapses them to a
+        // single value so the scalar sub-select is valid in the context of order by: rank by the best-matching score
+        .columns({ func: 'max', args: [{ func: searchFunc.func, args: [...searchFunc.args, { val: true }] }] })
```
johannes-vogel, line 2651: How do you solve that in where clauses atm?

johannes-vogel (changes requested):

db-service/lib/cqn4sql.js
patricebender: The `WHERE` doesn't need to collapse the to-many: the subquery `key IN (SELECT
key FROM …joins… WHERE search(…)`) yields one row per matching child, so the same key repeats, but `IN` checks set membership so the duplicates don't matter (`A.ID IN (10,10,10)` ≡ `A.ID IN (10)`).

hana/test/fuzzy.test.js
patricebender: true, I added a flag in cqn4sql for this.. even though I dislike the HANA specific stuff in the agnostic transformer, I think this is the easiest way. Do you have a better idea?

db-service/test/cqn4sql/search.test.js
patricebender, line 1: - opt out with `cds.hana.fuzzy.ranked_search = false`
- `implicit: true` in order by indicates runtime order by key mechanism → rank search takes precedence, runtime order by key directly after.

hana/lib/cql-functions.js
```diff
@@ -156,8 +156,9 @@ const StandardFunctions = {
       return `(CASE WHEN (${toString({ xpr })}) THEN TRUE ELSE FALSE END)`
     }
 
-    // fuzziness config
-    const fuzzyIndex = cds.env.hana?.fuzzy || 0.7
+    // fuzziness config; `fuzzy` may be a number (threshold) or an object (e.g. { ranked_search }),
+    // so only take it as the minimal score when it is actually a number
+    const fuzzyIndex = typeof cds.env.hana?.fuzzy === 'number' ? cds.env.hana.fuzzy : 0.7
```
johannes-vogel: I think we then should have a `score` in the object to still configure the score.

patricebender: - [x] additional to a plain `fuzzy: <score>` a `fuzzy: { score: <score>, ranked_search: … }`

hana/test/search-ranking.test.js
```diff
@@ -0,0 +1,80 @@
...
+  })
+
+  test('user-provided order by takes precedence over the search rank', async () => {
+    const { SearchAuthors } = cds.entities('search.ranking')
+    // by name desc: 'Weak' (20) before 'Strong' (10) — the OPPOSITE of the relevance order,
+    // so this only holds if user ordering wins and the rank is applied after it.
+    const res = await SELECT.from(SearchAuthors).columns('ID').search('Cat').orderBy('name desc')
+    expect(res.map(r => r.ID)).to.eql([20, 10])
+  })
+
```
patricebender, line 61: - [x] add test which queries against App service, e.g. catalogue service to see runtime implicit order by as secondary sort criterium after ranked search kicks ins

patricebender:
Superseded by #1725, which retargets this work directly onto `main` (GitHub blocked changing this PR's base while it was stacked on #1564). Same branch, rebased on main.

Issue #1721, opened 2026-09-03
Connection leak in generic-pool.js during peak load


samibouge:
**Title:** Builtin pool leaks a connection when a released resource is dispensed to an already timed-out `acquire()` — pool drains to 0 and never recovers

### Description of erroneous behaviour

In the builtin pool (`db-service/lib/common/generic-pool.js`, default since 3.0.0), `#dispense()` takes a resource out of `_available`, shifts the head of `_queue`, and if that request is no longer `PENDING` (its `acquireTimeoutMillis` timer already rejected it) it recurses and returns **without putting the resource back into `_available` or into `_loans`**:

```js
// db-service/lib/common/generic-pool.js @ main (identical in 3.0.1)
const dispense = async resource => {
  const request = this._queue.shift()
  if (!request) {
    resource.idle()
    this._available.add(resource)
    return false
  }
  if (request.state !== RequestState.PENDING) {
    this.#dispense()          // <- `resource` is dropped here
    return false
  }
  ...
```

Every `release()` that lands on a stale queue head therefore loses one connection permanently: `size` stays at `max`, so no replacement is ever created, and the connection stays open on the database. Once acquire timeouts start (any burst that outruns the pool for longer than `acquireTimeoutMillis`, which is 1 s in the production defaults), most releases hit a stale head, the pool empties itself within seconds and only a process restart recovers. Rejected requests also stay in `_queue` until a resource would be handed to them, so `pending` grows without bound.

The npm `generic-pool` library handles exactly this case by putting the resource back ([`_dispatchPooledResourceToNextWaitingClient`](https://github.com/coopernurse/node-pool/blob/v3.9.0/lib/Pool.js#L273-L283): *"While we were away either all the waiting clients timed out or were somehow fulfilled. put our pooledResource back."*). PR #1614 (§3, "`#dispense()` leaked a resource when the shifted request was no longer pending") described and fixed this, but was closed unmerged.

Observed in production shape (HANA Cloud, `pool.max = 10`, `acquireTimeoutMillis = 1000`): a 30-second database overload produced a few acquire timeouts, after which every request failed with `Pool resource could not be acquired within 1s` for the rest of the process lifetime while the database sat idle with all connections still open. Pool state sampled every second during a load test (`max: 40`):

| t | size | available | borrowed | pending |
|---|---|---|---|---|
| 22 s | 40 | 0 | 40 | 0 |
| 28 s | 40 | 0 | 26 | 33 |
[...]

PR #1724, opened 2026-09-04
fix(pool): recover lost connection after stale acquire timeout

swaldmann:
Similar to https://github.com/cap-js/cds-dbs/issues/1721, but avoiding the potential stack overflow via recursive calls by draining stale queue heads iteratively.


johannes-vogel (dismissed):

swaldmann:
Thanks for the test fixes @johannes-vogel! Should be RFR now.

johannes-vogel (approved):

sjvans (approved):

PR #1738, opened 2026-09-15
fix(cqn2pql): handle parameterized views in from_args

### LINKED ISSUE #1723: CQN2PQLRenderer.from_args throws 500 for parameterized views when DEBUG=all is set
**Describe the bug:**
When DEBUG=all (or any value that enables pql debug logging) is set as an environment variable, SQLService.cqn2pql is mounted and used as a middleware renderer. CQN2PQLRenderer extends CQN2SQL but does not override from_args, causing it to fall through to the base class which throws:

Error: Parameterized views are not supported by CQN2PQLRenderer
The normal cqn2sql path works correctly (e.g. @cap-js/hana's CQN2HANA properly overrides from_args).

**To reproduce:**

Define a parameterized view in CDS: define view Foo(p: String) as select from Bar where x = :p
Set env var DEBUG=all
Call GET /Foo(p='test')/Set
→ 500 Internal Server Error

Expected: Either the debug renderer delegates to the actual db-adapter's renderer for unsupported operations, or logs a warning and skips PQL logging for this query.

**Versions:**

@sap/cds: 10.0.6
@cap-js/db-service: 3.0.1
@cap-js/hana: 3.0.2

hm23:
## Problem

When `DEBUG=pql` (or `DEBUG=all`) is set, `SQLService.cqn2sql` is monkey-patched to first call `cqn2pql` for pretty-printing before delegating to the real renderer. `CQN2PQLRenderer` extends `CQN2SQL` but did not override `from_args`, causing it to fall through to the base-class stub which unconditionally throws:

```
Error: Parameterized views are not supported by CQN2PQLRenderer
```

This produced a 500 for any request against a parameterized view whenever debug logging was active — even though the actual SQL adapter (e.g. HANA) would have handled the query correctly.

## Fix

Override `from_args` in `CQN2PQLRenderer` to render named arguments as a human-readable inline list:

```js
from_args(args) {
  return `(${Object.keys(args).map(k => `${k}: ${this.expr(args[k])}`).join(', ')})`
}
```

Example output:
```sql
SELECT
  ID, title
FROM sap_capire_bookshop_ParamBooks(available: 100)
```

Fixes #1723

### DIFF
--- a/db-service/test/param-views-debug.test.js
+++ b/db-service/test/param-views-debug.test.js
@@ -1,0 +1,43 @@
+// Reproduction for: https://github.com/cap-js/cds-dbs/issues/1723
+// CQN2PQLRenderer.from_args throws for parameterized views when DEBUG=pql is set.
+//
+// The bug: when DEBUG=pql, SQLService.cqn2sql is wrapped to first call cqn2pql for
+// pretty-printing. cqn2pql uses CQN2PQLRenderer which extends CQN2SQL but does not
+// override from_args. The base-class from_args throws:
+//   "Parameterized views are not supported by CQN2PQLRenderer"
+
+const CQN2PQLRenderer = require('../lib/cqn2pql')
+const { expect } = require('@sap/cds').test
+
+describe('CQN2PQLRenderer - parameterized views (issue #1723)', () => {
+  // CQN that cqn4sql produces for: SELECT from ParamBooks(available: {val:100}) { ID, title }
+  const cqn = {
+    SELECT: {
+      from: {
+        ref: [
+          {
+            id: 'sap.capire.bookshop.ParamBooks',
+            args: { available: { val: 100 } },
+          },
+        ],
+      },
+      columns: [
+        { ref: ['ID'] },
+        { ref: ['title'] },
+      ],
+    },
+  }
+
+  test('render() should not throw for parameterized views', () => {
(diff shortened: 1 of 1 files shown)


db-service/test/param-views-debug.test.js
```diff
@@ -0,0 +1,43 @@
...
+// pretty-printing. cqn2pql uses CQN2PQLRenderer which extends CQN2SQL but does not
+// override from_args. The base-class from_args throws:
+//   "Parameterized views are not supported by CQN2PQLRenderer"
+
+const CQN2PQLRenderer = require('../lib/cqn2pql')
+const { expect } = require('@sap/cds').test
+
+describe('CQN2PQLRenderer - parameterized views (issue #1723)', () => {
+  // CQN that cqn4sql produces for: SELECT from ParamBooks(available: {val:100}) { ID, title }
+  const cqn = {
```
patricebender: prefer CQL

db-service/test/param-views-debug.test.js
patricebender, line 1: Lets call this test suite something more generic, if we need more tests maybe `cqn2pql.test.js`

hm23: done

db-service/test/param-views-debug.test.js
hm23: done

patricebender (approved):

